PR #115, opened 2016-03-01
Copy images

kookster:
We need to also copy image files over from existing feeds, not just the audio.

- [x] Create a model for episode images, and import/async update handling
- [x] Convert episode image_url attribute to episode_images in db
- [x] For podcast models, like audio, keep current good until new has processed
- [x] Refactor podcast images to single table
- [x] Trigger image processing in controllers and jobs


app/models/concerns/image_file.rb
```diff
@@ -4,16 +4,57 @@ module ImageFile
   extend ActiveSupport::Concern
 
   included do
+    has_one :task, as: :owner
+
     before_validation :detect_image_attributes
 
-    validates :url, presence: true
+    before_validation :guid
```
gcampo88: a little redundant since also called in #initialize_attributes, no?  -- unless you're thinking it's possible we'll have an image without a guid on a non-create ?

app/models/episode.rb
```diff
@@ -59,6 +65,10 @@ def enclosure
     enclosures.complete.first
   end
 
+  def image
+    images.complete.first
```
gcampo88: How come we need this? If it's for a default image, could we name it that?

app/models/concerns/image_file.rb
kookster: I think it's redundant, I'll fix.

app/models/episode.rb
kookster: this is not the default image, it is the current image for the episode.
There will be multiple images in `images` when we have an updated image, and there is a current image that is there, and then a new image is being processed and will replace it.

kookster: note: this is how enclosures and contents for audio both work too, somewhat poorly explained in the readme changes.

db/schema.rb
```diff
@@ -11,11 +11,32 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20170123205547) do
+ActiveRecord::Schema.define(version: 20170223174042) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
 
+  create_table "episode_images", force: :cascade do |t|
```
cavis: So in this new world, we've got `itunes_images`, `episode_images` and `podcast_images`.  The latter 2 make sense... but what is `itunes_images` used for, going forward?  (I thought maybe the "type" on the new image tables would be used to differentiate between the itunes/non-itunes channel image).

app/models/episode.rb
```diff
@@ -154,10 +164,15 @@ def file_size
   def copy_media(force = false)
     enclosures.each { |e| e.copy_media(force) }
     all_contents.each { |c| c.copy_media(force) }
+    images.each{ |i| i.copy_media(force) }
   end
 
   def base_published_url
-    "#{podcast.base_published_url}/#{guid}"
+    "https://#{feeder_cdn_host}/#{path}"
```
kookster: getting to it differently - I now need the podcast + episode path, and before the code was parsing the final url to get it, now I added a `path` method to get it from the episode, and used it here as well.

app/models/episode_story_handler.rb
```diff
@@ -73,13 +74,21 @@ def update_audio
     episode.all_contents.where(['position > ?', max_pos]).delete_all if max_pos > 0
   end
 
+  def update_image
+    if image_url = cms_url(story.objects['prx:image'].links['original'].href) rescue nil
```
kookster: yeah, could change some day

app/models/episode_story_handler.rb
```diff
@@ -88,7 +97,7 @@ def update_content(c, audio)
     c
   end
 
-  def audio_url(url)
+  def cms_url(url)
```
kookster: belt and suspenders...but yes, they SHOULD be fine.

db/schema.rb
kookster: Ok, almost. Now we have podcast_images and episode_images.
However, there are 2 kinds of podcast image: the itunes_image, and the feed_image, which are subclasses of podcast_image

app/controllers/api/auth/episodes_controller.rb
```diff
@@ -7,13 +7,26 @@ class Api::Auth::EpisodesController < Api::BaseController
   filter_resources_by :podcast_id
   find_method :find_by_guid
 
+  after_action :process_media, only: [:create, :update]
```
kookster: I remember there being a good reason for this, but I don't remember what. I'll try changing it and see what breaks.

kookster: Here it is, line 15 below in `show`, the super call. That doesn't work if it falls through to the episodes_controller, b/c it does different checks on the request - I think I can do something about that though.

cavis: I'm fine with it being explicit for now, if that's easiest.

cavis:
Works for me.

app/controllers/api/auth/episodes_controller.rb
```diff
@@ -1,32 +1,21 @@
...
-    relation.with_deleted
-  end
-
-  def process_media
-    resource.copy_media if resource
-  end
-
-  def publish
-    resource.podcast.publish! if resource && resource.podcast
+  def visibile?
```
gcampo88: visibile -> visible x 6

cavis:
Yeah, second on the `visibile` thing.  Otherwise looks good to me.

PR #122, opened 2016-05-04
Fix crier update episode

kookster:
dumb mistake, the update of an episode is not getting persisted via announce message processing.


cavis:
Yep yep.  👍

PR #126, opened 2016-07-21
Fix explicit handling

kookster:


cavis:
@kookster so what's the thinking here?  Podcasts Connect seems to indicate that those values are equivalent to them... `yes == explicit == true` and `no == clean == false`.

kookster:
@cavis Yeah, sorta, but there are preferred values, and ones that are sorta deprecated/old but still work. At one point, the docs said only "yes" and "clean", but they backed off that and allowed the other values. The blubrry powerpress plugin, for example, only supports "yes" and "clean" as values for itunes:explcit.

cavis:
👍 works for me.

PR #130, opened 2016-09-01
Fix resource attribute case

kookster:
old crier was sending underscore, cms has always been camel.
update feeder to handle either input from hal api responses, and convert to underscore keys for use in ruby.

### DIFF
--- a/lib/prx_access.rb
+++ b/lib/prx_access.rb
@@ -1,4 +1,11 @@
 module PRXAccess
+
+  class PRXHyperResource < HyperResource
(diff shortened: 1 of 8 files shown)


lib/prx_access.rb
```diff
@@ -1,22 +1,29 @@
 module PRXAccess
```
kookster: perhapst his could be the start of a gem to make these common methods available, and make it easier to call prx apis via ruby.

cavis:
Works for me! 👍

PR #131, opened 2016-09-07
Upgrade docker base, ruby, and gems

kookster:


Dockerfile
```diff
@@ -17,13 +23,21 @@ WORKDIR $APP_HOME
...
+    zlib-dev \
+    libxml2-dev \
+    libxslt-dev \
+    libffi-dev \
+    libgcrypt-dev && \
+    cd $APP_HOME && \
     bundle config --global build.nokogiri  "--use-system-libraries" && \
     bundle config --global build.nokogumbo "--use-system-libraries" && \
     bundle config --global build.ffi  "--use-system-libraries" && \
+    bundle config --global build.ruby-audio --with-cflags=-Wno-error=format-security && \
```
cavis: Does feeder actually use ruby-audio?

Dockerfile
```diff
@@ -1,10 +1,16 @@
...
-    ruby-bigdecimal ruby-io-console libstdc++ tzdata postgresql-client nodejs \
-    linux-headers libc-dev zlib libxml2 libxslt libffi less
+RUN apk update && apk --update add \
+    ca-certificates \
+    tzdata \
+    linux-headers \
+    libxml2 \
+    libxslt \
+    postgresql-client \
+    nodejs \
```
cavis: Is node actually required for this image?

kookster: it's for the asset pipeline, and we do actually compile assets for this one.
I could do some more work and dismantle the use of the pipeline in this app, so it would not complain it needs js support, but it adds little to the size, and means I can add pages as needed easily.

I had thought for awhile we might have some admin pages for this app, but I am increasingly thinking we will not, and interact with the API for whatever might need to be done, like using a slack action/bot to ping a feed or see the last time an episode dropped, so I am increasingly ok with dropping the asset pipeline and nodejs - what do you think @cavis?

Dockerfile
kookster: it doesn't, good catch. I left that in thinking it might make sense to standardize dockerfile across apps, but nvm bad idea.

Dockerfile
cavis: Would certainly speed stuff up to ditch the asset pipeline.  But since this PR already does 2 things ... I'd vote for opening up a chore-ticket to come back to later.

kookster: ok, will open a ticket. Here is the error w/o node.js on asset precompile:

```
Step 17 : RUN bundle exec rake assets:precompile
 ---> Running in 23240795a6a3
rake aborted!
Bundler::GemRequireError: There was an error while trying to load the gem 'uglifier'.
Gem Load Error is: Could not find a JavaScript runtime. See https://github.com/rails/execjs for a list of available runtimes.
Backtrace for gem load error is:
/usr/local/bundle/gems/execjs-2.7.0/lib/execjs/runtimes.rb:58:in `autodetect'
```

kookster: added #132

Dockerfile
kookster: fixed @cavis

cavis:
:+1:

PR #134, opened 2016-10-07
default content type to audio/mpeg, prevent bad type from fixer

kookster:


app/models/media_resource.rb
```diff
@@ -46,13 +46,23 @@ def update_from_fixer(fixer_task)
   end
 
   def update_attributes_with_fixer_info(info)
-    self.mime_type = info['content_type']
-    self.file_size = info['size'].to_i
+    # this is not working
+    # self.mime_type = info['content_type']
```
cavis: This isn't working in the "is null" sense, or in the "doesn't have a sane default" sense?

app/models/media_resource.rb
```diff
@@ -46,13 +46,23 @@ def update_from_fixer(fixer_task)
...
     self.sample_rate = info['sample_rate'].to_i
     self.channels = info['channels'].to_i
     self.duration = info['length'].to_f
     self.bit_rate = info['bit_rate'].to_i
     self
   end
+
+  def update_mime_type_with_fixer_info(info)
+    if info['content_type'].nil? || info['content_type'] == 'application/octect-stream'
+      self.mime_type ||= 'audio/mpeg'
```
cavis: You're thinking this is a safe-enough assumption, just because we deal with so little video?

kookster: yes, that is my thought. If it is ever wrong, we can console update the enclosure to have a correct value

cavis:
👍

PR #147, opened 2016-11-20
Improve API for updating podcasts and episodes

kookster:
- [x] Add in cors
- [x] Authentication
- [x] Pundit policies
- [x] Representer for podcast
- [x] Representer for episode

### DIFF
--- a/app/controllers/api/episodes_controller.rb
+++ b/app/controllers/api/episodes_controller.rb
@@ -3,4 +3,8 @@ class Api::EpisodesController < Api::BaseController
   represent_with Api::EpisodeRepresenter
   filter_resources_by :podcast_id
   find_method :find_by_guid
+
+  def scoped(relation)
+    relation.published.released
+  end
 end
--- a/app/models/itunes_image.rb
+++ b/app/models/itunes_image.rb
@@ -1,5 +1,6 @@
 class ITunesImage < ActiveRecord::Base
   belongs_to :podcast
+  attr_accessor :link, :description
 
   validates :height, :width, numericality: {
     less_than_or_equal_to: 3000,
--- a/app/models/podcast.rb
+++ b/app/models/podcast.rb
@@ -19,6 +19,10 @@ class Podcast < ActiveRecord::Base
     DateUpdater.last_build_date(self)
   end
 
+  def account_id
+    URI.parse(prx_account_uri || '').path.split('/').last.to_i
+  end
+
   def path
     self[:path] || id
   end
@@ -126,7 +130,7 @@ class Podcast < ActiveRecord::Base
 
   def feed_episodes
     feed = []
-    feed_max = max_episodes.to_i
+    feed_max = display_episodes_count.to_i
     episodes.published.released.each do |ep|
       feed << ep if ep.include_in_feed?
       break if (feed_max > 0) && (feed.size >= feed_max)
--- a/app/representers/api/podcast_representer.rb
+++ b/app/representers/api/podcast_representer.rb
@@ -2,6 +2,10 @@
 
 class Api::PodcastRepresenter < Api::BaseRepresenter
   property :prx_uri
+  property :prx_account_uri
+  property :created_at
+  property :updated_at
+
   property :path
   property :url
   property :published_url, writeable: false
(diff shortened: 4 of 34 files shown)


kookster:
Assigning a few folks, thought @gcampo88 and I could review together also to see how some of this API stuff works

app/controllers/api/episodes_controller.rb
```diff
@@ -3,4 +3,8 @@ class Api::EpisodesController < Api::BaseController
   represent_with Api::EpisodeRepresenter
   filter_resources_by :podcast_id
   find_method :find_by_guid
+
+  def scoped(relation)
+    relation.published.released
```
cavis: Is there a reason to show only published/released episodes here?  There may be some uses for the API results not being an exact representation of what's in the feed.

app/models/podcast.rb
```diff
@@ -19,6 +19,10 @@ class Podcast < ActiveRecord::Base
     DateUpdater.last_build_date(self)
   end
 
+  def account_id
+    URI.parse(prx_account_uri || '').path.split('/').last.to_i
```
cavis: I don't have a better suggestion (and have been doing this myself)... but these always catch my eye.

app/representers/api/podcast_representer.rb
```diff
@@ -1,4 +1,67 @@
 # encoding: utf-8
 
 class Api::PodcastRepresenter < Api::BaseRepresenter
+  property :id
```
cavis: 👍!!!  Every couple weeks, I end up trying to use the feeder api to find a podcast.  And it always ends in tears.

app/views/podcasts/show.rss.builder
```diff
@@ -74,11 +74,7 @@ xml.rss 'xmlns:atom' => 'http://www.w3.org/2005/Atom',
     xml.sy :updateFrequency, @podcast.update_frequency if @podcast.update_frequency
     xml.sy :updateBase, @podcast.update_base if @podcast.update_base
 
-    episode_count = @podcast.display_episodes_count.to_i
```
cavis: Do we no longer use this item limit on the feed?

config/environments/development.rb
```diff
@@ -24,4 +24,9 @@
 
   # Raises error for missing translations
   # config.action_view.raise_on_missing_translations = true
+  # Better logging for docker-compose
+  config.logger = Logger.new(STDOUT)
+
+  # PrxAuth::Rails.middleware = false
+  # config.middleware.insert_before 'ActionDispatch::ParamsParser', 'Rack::PrxAuth', cert_location: 'http://id.prx.dev/api/v1/certs', issuer: 'id.prx.dev'
```
cavis: Yeah, I usually comment this out in cms-dev as well, and just use id.prx.org.  Maybe we should make that our default there too.

app/models/itunes_image.rb
```diff
@@ -1,5 +1,6 @@
 class ITunesImage < ActiveRecord::Base
   belongs_to :podcast
+  attr_accessor :link, :description
```
cavis: I'm not sure I follow what these are for.

app/controllers/api/episodes_controller.rb
kookster: This allows this API endpoint to be public and to mirror the feed in what episodes it returns.
I have another PR I started where non-public episodes can be returned via an authenticated request, like cms

app/representers/api/podcast_representer.rb
kookster: my hope is to make this api useful for admins as well

app/views/podcasts/show.rss.builder
kookster: there were 2 attributes enforcing this, one in the view, one in the model - now just using one in the model

config/environments/development.rb
kookster: I have it in there for when I am (rarely) working on id

app/models/itunes_image.rb
kookster: so the single image representer can handle both, in another PR I actually unify this under one table with all the same columns, this is a stop gap

app/controllers/api/episodes_controller.rb
cavis: Gotcha.  I was thinking that there would never be anything in the Feeder database that wasn't already "public" in CMS.  But with the ability to unpublish or stage-for-release something, I guess that's not the case.

kookster: yeah, and I am thinking ahead to knowing the episode or podcast url before it is public, and my current thought is even to create a stub podcast or episode entry sooner rather than later in the create/save process.

Even without that feeder, has a notion of episodes w/o audio fully processed that are not in the feed yet.

cavis:
This is a :ship: :it: for me.

kookster:
@gcampo88 and I reviewed at length

PR #148, opened 2016-11-20
Authenticated urls

kookster:
- [x] Add auth controller for getting to auth only endpoints
- [x] Add auth episode controller, showing private episodes
- [x] Add in a published_at field for podcasts so they can be private as well
- [x] Add authorization model
- [x] Re-parent models to use hal api rails RepresentedModel concern

### DIFF
--- a/app/controllers/api/authorizations_controller.rb
+++ b/app/controllers/api/authorizations_controller.rb
@@ -1,0 +1,13 @@
+# encoding: utf-8
+
+class Api::AuthorizationsController < Api::BaseController
+  include ApiAuthenticated
+  api_versions :v1
+  represent_with Api::AuthorizationRepresenter
+
+  private
+
+  def resource
+    prx_auth_token
+  end
+end
--- a/app/controllers/api/authorizations_controller.rb
+++ b/app/controllers/api/authorizations_controller.rb
@@ -8,6 +8,6 @@ class Api::AuthorizationsController < Api::BaseController
   private
 
   def resource
-    prx_auth_token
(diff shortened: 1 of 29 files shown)


app/models/authorization.rb
```diff
@@ -0,0 +1,28 @@
...
+
+  def to_model
+    self
+  end
+
+  def persisted?
+    false
+  end
+
+  def cache_key
```
cavis: Does this need a cache key?  I thought `/authorization` resources weren't cached.

app/controllers/api/authorizations_controller.rb
```diff
@@ -0,0 +1,13 @@
...
+
+class Api::AuthorizationsController < Api::BaseController
+  include ApiAuthenticated
+  api_versions :v1
+  represent_with Api::AuthorizationRepresenter
+
+  private
+
+  def resource
+    Authorization.new(prx_auth_token)
```
cavis: Ah, right ... because there is no `users` table in feeder.

app/models/authorization.rb
kookster: I don't think it is or should be, but not knowing exactly how this model may be used in the future, I wanted to provide a sensible implementation for it, since it is a standard method for a hal represented model

app/controllers/api/authorizations_controller.rb
kookster: right! If we really need it, should look it up via API, but the token provides some of the same info, so I went with that.

cavis:
👍

PR #149, opened 2016-12-01
Only show prx cms links for feeds and episodes connected to the cms

kookster:
fixes a bug where the representer throws an error when the prx related uris are null.


kookster:
@cavis small bug fix, ready for review

PR #152, opened 2016-12-14
Return 410

kookster:
- [x] return 410 instead of 404 for deleted resources
- [x] touch parent resource for cache invalidation
- [x] add `published_at` to podcast representer
- [x] update `PRXAccess` because why not (should be a gem soon I think)


kookster:
ok - @cavis  - this should do caching, good call.

PR #157, opened 2016-12-19
Simpler dates

kookster:
Podcasts:
- [x] use updated_at instead of last_build_date
- [x] use max episode published_at instead of pub_date

Episodes:
- [x] use updated_at instead of updated
- [x] use published_at instead of published
- [x] use published_at in the future instead of released_at


app/models/podcast.rb
```diff
@@ -14,10 +14,22 @@ class Podcast < BaseModel
...
   acts_as_paranoid
 
-  scope :published, -> { where('published_at IS NOT NULL') }
+  scope :published, -> { where('published_at IS NOT NULL AND published_at <= now()') }
 
-  after_save do
-    DateUpdater.last_build_date(self)
+  def pub_date
+    if max_episode_published_at
+      [published_at, max_episode_published_at].max
```
cavis: Is there a reason to not just change the `Podcast.published_at` to be the max `Episode.published_at`, when the updated episode comes in?

db/migrate/20161218183833_simpler_dates.rb
```diff
@@ -0,0 +1,25 @@
+class SimplerDates < ActiveRecord::Migration
+  def up
+    execute 'update podcasts set published_at = pub_date where pub_date is not null and pub_date > published_at'
+    execute 'update podcasts set updated_at = last_build_date where last_build_date is not null and last_build_date > updated_at'
+    remove_column :podcasts, :last_build_date
+    remove_column :podcasts, :pub_date
+
+    execute 'update episodes set published_at = published where published is not null and published > published_at'
+    execute 'update episodes set updated_at = updated where updated is not null and updated > updated_at'
+    remove_column :episodes, :released_at
```
cavis: This is just totally redundant with `published` then?  (I notice you're setting `published_at` without looking at this column)

app/representers/api/podcast_representer.rb
```diff
@@ -9,6 +9,9 @@ class Api::PodcastRepresenter < Api::BaseRepresenter
   property :updated_at
   property :published_at
 
+  property :pub_date, writeable: false
+  property :last_build_date, writeable: false
```
cavis: Is this necessary, since `updated_at` is already represented?  (And this doesn't look at the episodes the way `pub_date` does)

app/models/podcast.rb
kookster: I suppose we could.

I'm trying to think if we lose anything that way - I guess if you wanted to launch a podcast in the future, you would have to do so by having the first episode with a future published_at

I had set it this way thinking you might have a different published_at for the podcast itself for some reason, but yeah, pretty much dependent on the episodes.

db/migrate/20161218183833_simpler_dates.rb
kookster: I see no way they are different concepts, but I am looking up published_at and only overwriting if published is greater then published_at

app/representers/api/podcast_representer.rb
kookster: probably not necessary, I guess I was thinking these are useful as they correspond to the values in the rss feed exactly

app/models/podcast.rb
kookster: I'll try changing this.

kookster:
@cavis better I think

cavis:
Looks good to me.

PR #158, opened 2016-12-19
Call publish on actions that change podcasts and episodes

### LINKED ISSUE #155: `publish!` not called when podcast changed via api
and if publish isn't called `pub_date` isn't updated, and the latest isn't published to s3

kookster:
fixes #155 

Basically, when feeder was created, all changes came in async SQS messages SQS via `announce`.
Now that we are also affecting records via the API, need to also republish the rss xml feed when there is a change to related records via the API.

### DIFF
--- a/app/controllers/api/episodes_controller.rb
+++ b/app/controllers/api/episodes_controller.rb
@@ -4,6 +4,8 @@ class Api::EpisodesController < Api::BaseController
   filter_resources_by :podcast_id
   find_method :find_by_guid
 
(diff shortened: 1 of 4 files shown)


kookster:
Another for the super team of @gcampo88  & @cavis

cavis:
Despite the un-team-work-ness of it, going to merge this without gigi's review.

gcampo88:
redundant 👍 !

PR #163, opened 2017-01-10
Remove unused files

kookster:
- [x] `container` files replaced by cloud formation
- [x] unicorn replaced by puma
- [x] Procfile for heroku replaced by using ECS


gcampo88:
code looks good but this makes the [README scripts section](https://github.com/PRX/feeder.prx.org#scripts) pretty obsolete!

PR #164, opened 2017-01-10
Enclosure prefix

kookster:
- [x] Add enclosure prefix for user to easily set for podtrac/blubrry
- [x] Default the enclosure template to use dovetail url

(merge #163 first please)


app/models/podcast.rb
```diff
@@ -14,8 +14,18 @@ class Podcast < BaseModel
...
+  before_validation :set_defaults
+
   scope :published, -> { where('published_at IS NOT NULL AND published_at <= now()') }
 
+  def set_defaults
+    self.enclosure_template ||= enclosure_template_default
+  end
+
+  def enclosure_template_default
+    "https://#{ENV['DOVETAIL_HOST']}/{slug}/{guid}/{original_filename}"
```
cavis: Is the assumption that all of feeder will go through Dovetail then?  Rather than allowing someone to use the raw `f.prxu.org` enclosures?  (Or maybe I'm misreading this entirely).

cavis: Yeah, as I re-read this, the `enclosure_prefix` stuff you added makes sense.  But how `enclosure_template` fits in there is still a bit murky.  I may just need to revisit in the morning.

kookster: yes, I am assuming all our feeds will go through dovetail by default. This could be revisited later.

kookster: The template defines the overall order and shape of the url - does it have dovetail in there, is the original file name used, is the guid? It is optional - as you mention, we have that enclosure default that will return an f.prxu.org address.

The prefix is also optional, and will add a redirect prefix to whatever comes out for the enclosure url after the template (if there is one) has been applied to the enclosure url. It is only able to effect adding a prefix.

There is some overlap here, as a user could add a prefix in their template (which is what we have done in the past), but by adding this prefix, we have a way to create a user editable field that can only cause so much damage, where editing the template would be a bear.

Issue #168, opened 2017-01-11
App should return error for authorization error, not a 204


kookster:
feeder is returning a 204 even when an update has failed because of an authorization error

```
feeder_1 | Started PUT "/api/v1/podcasts/2" for 172.17.0.2 at 2017-01-11 17:52:51 +0000
feeder_1 | Value for params[:itunesCategories] was set to nil, because it was one of [], [null] or [null, null, ...]. Go to http://guides.rubyonrails.org/security.html#unsafe-query-generation for more information.
feeder_1 | Processing by Api::PodcastsController#update as HAL
feeder_1 |   Parameters: {"id"=>"2", "createdAt"=>"2017-01-11T17:44:47.416Z", "updatedAt"=>"2017-01-11T17:44:47.416Z", "publishedAt"=>"2017-01-10T17:44:42.334Z", "path"=>2, "publishedUrl"=>"http://development-f.prxu.org/2/feed-rss.xml", "title"=>"test 2", "owner"=>{}, "author"=>{}, "managingEditor"=>{}, "itunesCategories"=>nil, "api_version"=>"v1", "base"=>{"id"=>"2", "createdAt"=>"2017-01-11T17:44:47.416Z", "updatedAt"=>"2017-01-11T17:44:47.416Z", "publishedAt"=>"2017-01-10T17:44:42.334Z", "path"=>2, "publishedUrl"=>"http://development-f.prxu.org/2/feed-rss.xml", "title"=>"test 2", "owner"=>{}, "author"=>{}, "managingEditor"=>{}, "itunesCategories"=>nil}}
feeder_1 | Can't verify CSRF token authenticity
feeder_1 |   Podcast Load (0.4ms)  SELECT  "podcasts".* FROM "podcasts" WHERE "podcasts"."deleted_at" IS NULL AND (published_at IS NOT NULL AND published_at <= now()) AND "podcasts"."id" = $1 LIMIT 1  [["id", 2]]
feeder_1 | Unpermitted parameters: id, createdAt, updatedAt, publishedAt, path, publishedUrl, title, owner, author, managingEditor, itunesCategories, base
feeder_1 |   ITunesCategory Load (0.3ms)  SELECT "itunes_categories".* FROM "itunes_categories" WHERE "itunes_categories"."podcast_id" = $1  [["podcast_id", 2]]
feeder_1 | 
feeder_1 | NotAuthorized (You are not authorized to perform this action):
feeder_1 |   
feeder_1 | 
[...]

PR #169, opened 2017-01-11
Fix authorization error API response

kookster:
#168


kookster:
idk if this is a known thing, I found a somewhat related issue that was only for content type mismatch, where the `rescue_from` used a `render` rather than a `respond_with`.

I had seen that was how CMS worked for enforcing auth checks, with not using a responder, and had thought I was really clever and tried to use the responder in feeder - thus adding this bug.

Issue #170, opened 2017-01-11
On episode get, redirect to auth resource when user could access it instead of 404


kookster:

kookster:
#171

PR #171, opened 2017-01-11
Redirect authorized access to unpublished instead of returning 404

kookster:
#170 

- [x] Send redirect to authorized version of resource instead of 404 when unpublished, but auth token allows update access
- [x] Make podcasts work the same way as episodes for handling publishing and authorized access

### DIFF
--- a/app/controllers/api/auth/podcasts_controller.rb
+++ b/app/controllers/api/auth/podcasts_controller.rb
@@ -1,11 +1,24 @@
 # encoding: utf-8
 
-class Api::Auth::PodcastsController < Api::PodcastsController
+class Api::Auth::PodcastsController < Api::BaseController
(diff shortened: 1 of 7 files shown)


app/controllers/api/auth/podcasts_controller.rb
```diff
@@ -1,11 +1,24 @@
 # encoding: utf-8
 
-class Api::Auth::PodcastsController < Api::PodcastsController
+class Api::Auth::PodcastsController < Api::BaseController
```
cavis: Why change the inheritance?

kookster: so when I call super in show I am not calling the un-authed version that will redirect back to myself

PR #173, opened 2017-01-12
Fix API updates, return errors and resources

kookster:
- [x] Return error message json and 500 status for errors, instead of 204
- [x] http responses for PUT now include the message body
- [x] test for updating read only properties - these are ignored, no error
- [x] fix `id` and timestamp values to be read only that really should be read only
- [x] add `enclosure` link to episode - has same data as enclosure link in the rss feed


kookster:
@cavis @sandikbarr much the same as the update to the cms - I will go back and refactor this responder into a shared lib, but didn't want to hold things up to do that.

Issue #165, opened 2017-01-10
Subscribe to series changes


kookster:
subscribe to announcements of series changes, update podcast with this info

PR #185, opened 2017-01-19
Handle series announcements

kookster:
#165 fixed
- [x] Add handlers for the podcasts to update via crier feed updates, or cms series updates
- [x] Add `source_updated_at` for podcasts and episodes to make sure only latest changes used


kookster:
@cavis should be good to review!

kookster:
@cavis actually, there may be a problem with something in this.
We don't want to create podcasts for all new series in v4, which I believe this will do as it is.

I think we probably want to disable the podcast create for a series, and instead expect the create will come from publish calling the feeder api.

cavis:
Agreed.  Could the same argument be made about Feeder listening to story-create?

kookster:
I'm lying, I did think of this:
https://github.com/PRX/feeder.prx.org/pull/185/files#diff-c106db5cf2b1106e4a517c8ef3119351R21

kookster:
On story events, does not call create podcast, and will fail if there is no podcast already to create a new story.

kookster:
@cavis I think this is better now.

app/jobs/series_update_job.rb
```diff
@@ -0,0 +1,49 @@
...
+
+  def perform(*args)
+    ActiveRecord::Base.connection_pool.with_connection do
+      super
+    end
+  end
+
+  def receive_series_update(data)
+    load_resources(data)
+    if podcast
```
cavis: And I'm assuming that this will actually pick up create messages ... which _should_ be largely a no-op, since CMS just POSTed very similar data.

app/jobs/story_update_job.rb
```diff
@@ -20,7 +20,7 @@ def receive_story_update(data)
     load_resources(data)
     episode ? update_episode : create_episode
```
cavis: So this is getting create/update messages for ALL stories, not just ones we know are in a podcast-distributed series, right?  Couldn't this result in `episode` and `podcast` being undefined on the next line, if the story has no series?

Also seems like the error thrown from `Podcast.find_by!(prx_uri: series_uri)` would just bubble up through here.

... correct me if i'm way off here.

app/models/podcast_series_handler.rb
```diff
@@ -0,0 +1,91 @@
...
+      podcast.send("build_#{type}_image", url: url)
+    end
+  end
+
+  def get_series(account = nil)
+    return nil unless podcast.prx_uri
+    api(account: account).tap { |a| a.href = podcast.prx_uri }.get
+  end
+
+  def cms_url(url)
```
cavis: Is this still necessary, since CMS only returns absolute asset links?

app/jobs/feed_entry_update_job.rb
```diff
@@ -60,9 +83,13 @@ def receive_feed_entry_delete(data)
     podcast.try(:publish!)
   end
 
+  def update_sent
+    @_update_sent ||= Time.parse(message[:sent_at])
```
kookster: You are right, I am being paranoid, it should be set, but just in case the `announce` gem changes or has some bug I don't expect, I'm making sure things will continue as gracefully as possible.

app/jobs/series_update_job.rb
kookster: exactly on both counts.

app/jobs/story_update_job.rb
kookster: episodes require a podcast_id to be created, so we won't ever make an episode for a story when that story isn't both in a series, and that series already has a podcast previously created for it.

So if there is no series for the story at all, it won't get past `create_episode` where is checks for that.
If there is a series, but no matching podcast, it will fail on `Podcast.find_by!`.

If I handle that differently, and just return a nil episode,  `episode.copy_media` will fail with a no method error on nil object, I should probably make that a `try` to prevent that error bubbling up.

app/models/podcast_series_handler.rb
kookster: not necessary - belt and suspenders

app/jobs/story_update_job.rb
kookster: I'm going to make a change for this, then you can have another look?

app/models/podcast_series_handler.rb
cavis: 👍

app/jobs/story_update_job.rb
cavis: Sounds good.  I guess i just want this method to silently do nothing if (1) the story has no series, or (2) the story-series isn't in feeder.  Which feel like non-errors.  And currently it'll throw in either case, which will probably reach newrelic.

kookster: I agree, I didn't think those scenarios out enough in this, will shush it.

cavis: And hey, just saw a `undefined method 'copy_media' for nil:NilClass` from feeder-staging.  Progress!

kookster:
Right back at ya @cavis

cavis:
Yesssssss!

Issue #182, opened 2017-01-17
Item <itunes:author> not following <author>


cavis:
When no author is set for an episode, the item-itunes-author is set to the channel-itunes-author.  (Is that really what we want?)

And when you explicitly set the author for that episode, you do get an item-author tag, but the item-itunes-author does not change.  And it should.

PR #187, opened 2017-01-20
Correctly set episode itunes|author 

gcampo88:
#182

### DIFF
--- a/app/views/podcasts/show.rss.builder
+++ b/app/views/podcasts/show.rss.builder
@@ -94,9 +94,8 @@ xml.rss 'xmlns:atom' => 'http://www.w3.org/2005/Atom',
 
           Array(ep.categories).each { |c| xml.category { xml.cdata!(c || '') } }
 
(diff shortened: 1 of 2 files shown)


app/views/podcasts/show.rss.builder
```diff
@@ -94,7 +94,10 @@ xml.rss 'xmlns:atom' => 'http://www.w3.org/2005/Atom',
 
           Array(ep.categories).each { |c| xml.category { xml.cdata!(c || '') } }
 
-          xml.itunes :author, @podcast.author_name unless @podcast.author_name.blank?
+          if ep.author_name || @podcast.author_name
```
cavis: Should these still be checking for `blank?`?  Or would a blank string not make it this far?

kookster: I'm not sure, I usually check for blanks because I'm paranoid.

gcampo88: good call! updated

Issue #189, opened 2017-01-23
Provide enclosure url before all audio finished processing


kookster:
For folks who want to know the enclosure url for an episode in the response from creating the episode.
Means they don't have to poll for the url after it completes processing, nor do we need a webhook to let them know when the enclosure url is available.

Also include the status of the enclosure on the link so the API user can know the status, both for the enclosure, and the individual audio files.

PR #190, opened 2017-01-23
Enable direct upload of episodes and audio

kookster:
This is a collection of changes to make direct API access to add episodes to feeder, without messages from crier or a related story in cms.

- [x] Allow direct adding of media files via the api
- [x] Fix auth to enforce resource creates to have correct account id
- [x] Show enclosure url when audio is still processing #189

Misc. cleanup:
- [x] Get rid of that old api no one uses anymore


app/models/episode_entry_handler.rb
```diff
@@ -90,15 +90,12 @@ def update_contents
       return
     end
 
-    final_contents = []
     new_contents.each do |c|
-      existing_content = find_existing_content(c)
+      existing_content = episode.find_existing_content(c[:position], c[:url])
```
cavis: Is this not duplicating the work of `Episode.update_contents`?  Or is this slightly different?

app/models/content.rb
```diff
@@ -18,7 +18,7 @@ def update_attributes_with_content(content)
       self.try("#{at}=", content[at])
     end
 
-    self.original_url = content['url']
+    self.href = content['url']
```
cavis: So if i'm following this correctly ... this is basically doing the same thing, except that it's going to null-out the status every time the original_url changes.  Which will cause the representer to return the original url instead of the completed/processed url.

Right?

app/models/episode.rb
```diff
@@ -207,6 +218,48 @@ def media_files
     end
   end
 
+  def media_files=(files)
```
cavis: Where is this ever used?  Just via the API (episode representer)?

kookster: yes, just for the api.

app/models/content.rb
kookster: When an audio file is changed (different original_url) via crier or cms announcement, the existing file is not updated, instead a new one is inserted, so it will continue to show the previously existing file for the same position until the new file has finished processing and deletes it. You can see this in the handlers and how they process changes to files.

So really, this should never result in an update.

That said, if logic elsewhere was changed, and a content was updated with a different url, then I do think this is the correct behavior. Without this, the original url would be changed, but because the file was already marked completed and a task to copy the (previous) original_url still exists, the new original_url audio won't actually get processed or updated at all - it will still act like it has processed b/c the prior url was processed and the task and status have not been reset.

app/models/episode.rb
kookster: yes, just via the API - I thought about creating an api handler (like the ones for crier and cms) since this is similar, but ended up deciding the API updates should stay here as they are more central or essential to the app (this is the app's api in its own format, not data coming from another system in that systems format).

app/models/episode_entry_handler.rb
kookster: slightly different - that is processing updates from the API, this is updates from crier, subtly different data being processed.

kookster: there was enough commonality to refactor `find_existing_content` to a single model method.

cavis:
That'll do, @kookster.  That'll do.

Issue #102, opened 2015-12-13
Add human readable episode identifier for adzerk keyword


kookster:
We don't know if the keyword can contain spaces, special characters, etc.
We also don't know how long it can be.
We do want to add an episode specific keyword, which could be a sanitized version of the title (persisted even if the title changes) or sanitized original GUID.

kookster:
http://dev.adzerk.com/docs/keyword-targeting
http://dev.adzerk.com/docs/keyword-targeting#section-reporting-on-keywords

gcampo88:
per @kookster : 
We do reporting in adzerk for what ads get served, and we're really good 
at reporting on what ads were served in what timeframe, on what feed. What we 
don't do well is report on which episode those ads were served on.
For example, wanting to know how many and what ads were served on the latest episode. 
To do that reporting better, we ought to have an episode identifier as 
one of the values you can filter a report on. So, this ticket is to come up with a good, 
human readable identifier which will uniquely identify an episode for that reporting

Considerations: 

- one is making it unique to the episode
- two is human readable 
- and I guess three is making sure it is a value adzerk can handle
- 

The flow here is that the feeder api will provide it on the episode representer.
So dovetail will call feeder api for an episode by guid, in the resource that comes 
back will be this identifier, and dovetail will pass that along as a 'keyword' to adzerk.
b/c that's one of the customizable things we can pass along to adzerk
http://dev.adzerk.com/docs/keyword-targeting
http://dev.adzerk.com/docs/keyword-targeting#section-reporting-on-keywords
Info + rules about keywords + formatting in links above.

Note, the max for all keywords on a request is 512, but that is for all keywords, not 
just this id, so we probably want to keep it shorter than that, both to make it easier to read, 
and leave space for more. We're mostly planning to use this keyword for reporting, not so much for 
targeting. So the human readable part is key, so it is easy to use in the reports

Maybe some combo of part of the title + part of the guid + part of the published_at. 
To make sure these values are unique, and also don't change, may need to "freeze" 
the value in its own column, even if it is derived from several.

PR #191, opened 2017-01-24
Add external keyword to episodes for tracking purposes

gcampo88:
#102
Adds a unique + unchanging identifying keyword to each published episode (combo of title, published_at, and guid at time of first publishing). 
Adzerk keywords are case-insensitive, can contain spaces, but can't contain commas or colons; and are capped at 512 characters for all keywords for a given request. I've limited this adzerk keyword (which we'll use for identifying + [reporting](http://dev.adzerk.com/docs/keyword-targeting#section-reporting-on-keywords) purposes) to 30 characters, to leave plenty of space for us to potentially use other keywords for [targeting](http://dev.adzerk.com/docs/keyword-targeting) purposes down the line.

### DIFF
--- a/app/models/episode.rb
+++ b/app/models/episode.rb
@@ -216,7 +216,7 @@ class Episode < BaseModel
     identifiers = []
     [:title, :published_at, :guid].each do |attr|
       # Adzerk does not allow commas or colons in keywords
-      identifiers << self.send(attr).slice(0, 10).gsub(/[:,]/,'')
+      identifiers << self.send(attr).to_s.slice(0, 10).gsub(/[:,]/,'')
     end
     self.adzerk_keyword = identifiers.join('_')
   end
--- a/app/models/episode.rb
+++ b/app/models/episode.rb
@@ -30,7 +30,7 @@ class Episode < BaseModel
 
   validates :podcast_id, :guid, presence: true
 
-  before_validation :initialize_guid, :set_adzerk_keyword
+  before_validation :initialize_guid, :set_external_keyword
 
   after_save :publish_updated, if: -> (e) { e.published_at_changed? }
 
@@ -211,13 +211,14 @@ class Episode < BaseModel
     media_files
   end
 
-  def set_adzerk_keyword
-    return unless published? && adzerk_keyword.nil?
+  def set_external_keyword
+    return unless !published_at.nil? && keyword_xid.nil?
     identifiers = []
-    [:title, :published_at, :guid].each do |attr|
-      # Adzerk does not allow commas or colons in keywords
-      identifiers << self.send(attr).to_s.slice(0, 10).gsub(/[:,]/,'')
+    [:published_at, :guid].each do |attr|
+      # Adzerk does not allow commas or colons in keywords; omitting dashes for space
+      identifiers << self.send(attr).to_s.slice(0, 10).gsub(/[:,-]/,'')
     end
-    self.adzerk_keyword = identifiers.join('_')
+    identifiers << title.slice(0, 20)
+    self.keyword_xid = identifiers.join('_')
   end
 end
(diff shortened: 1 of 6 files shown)


app/models/episode.rb
```diff
@@ -210,4 +210,14 @@ def media_files
...
   end
+
+  def set_adzerk_keyword
+    return unless published? && adzerk_keyword.nil?
+    identifiers = []
+    [:title, :published_at, :guid].each do |attr|
+      # Adzerk does not allow commas or colons in keywords
+      identifiers << self.send(attr).to_s.slice(0, 10).gsub(/[:,]/,'')
+    end
+    self.adzerk_keyword = identifiers.join('_')
```
kookster: I think I would be tempted to put the date first, and I wonder if we need the dashes in it? (save 2 chars!)

Likewise with the guid: "6d6f619d-5052-4330-80b8-34f414d076f9" becomes "6d6f619d-5", could become "6d6f619d50" without the dash.

And maybe put those additional chars towards the title which is pretty short at just 10 chars, especially if there is an episode # up front?

`20170124_100 - This is a_6d6f619d50`

is that better to anyone else? Maybe we should check with Kali?

app/models/episode.rb
```diff
@@ -30,7 +30,7 @@ class Episode < BaseModel
 
   validates :podcast_id, :guid, presence: true
 
-  before_validation :initialize_guid
+  before_validation :initialize_guid, :set_adzerk_keyword
```
kookster: Maybe also have a uniqueness validation on this to mirror the postgres constraint?

app/representers/api/episode_representer.rb
```diff
@@ -12,6 +12,9 @@ class Api::EpisodeRepresenter < Api::BaseRepresenter
   property :prx_uri
   property :published_at
 
+  # combo of title, published_at, and guid at time of first pub
+  property :adzerk_keyword
```
kookster: Is this supposed to be writeable? Probably not, since it is generated, perhaps add a `writeable: :false` option on this?

db/migrate/20170123205547_add_adzerk_keyword.rb
```diff
@@ -0,0 +1,5 @@
+class AddAdzerkKeyword < ActiveRecord::Migration
+  def change
+    add_column :episodes, :adzerk_keyword, :string
```
cavis: `external_id`?  `keyword_xid`?  `name_xid`?  `external_key`?

sandikbarr: I like `keywork_xid`, good idea

app/models/episode.rb
gcampo88: :+1:

app/models/episode.rb
```diff
@@ -210,4 +210,14 @@ def media_files
   def audio_files
     media_files
   end
+
+  def set_adzerk_keyword
+    return unless published? && adzerk_keyword.nil?
```
gcampo88: sure, that sounds good. I'm guessing it's unlikely we'll have episodes that have a set `published_at` but no `title` yet.

app/models/episode.rb
gcampo88: Skipping the dashes, putting date first, and giving extra space to title sounds good -- I'll check with Kali tomorrow to see how that combo sounds too.

db/migrate/20170123205547_add_adzerk_keyword.rb
gcampo88: `keyword_xid` it shall be!

app/representers/api/episode_representer.rb
gcampo88: 👍

gcampo88:
@kookster think I covered all your points! Will ask Kali tomorrow about how this plan -- `up-to-10-char-date _ up-to-10-char-GUID _ up-to-20-char-title` works from her perspective.

kookster:
thank you!

Issue #188, opened 2017-01-20
Set podcast published_at / pubDate to max visible published episode published_at


kookster:
- [ ] Use the max published_at for published and visible episodes (not future `published_at`)
- [ ] Do not nil out the podcast.published_at, only change the value if the episode max is different than current value, but not nil

PR #192, opened 2017-01-25
Don't unpublish a podcast based on episode published_at

gcampo88:
#188

### DIFF
--- a/app/models/episode.rb
+++ b/app/models/episode.rb
@@ -42,7 +42,7 @@ class Episode < BaseModel
   end
 
   def publish_updated
-    podcast.publish_updated if podcast && published?
+    podcast.publish_updated if podcast
   end
 
   def published?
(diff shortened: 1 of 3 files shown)


app/models/episode.rb
```diff
@@ -42,7 +42,7 @@ def self.by_prx_story(story)
   end
 
   def publish_updated
-    podcast.publish_updated if podcast
+    podcast.publish_updated if podcast && published?
```
cavis: I sort of think you should always update the podcast here.  What if the update to this episode was that it was unpublished?

kookster: I agree with @ryan, also the episode shouldn't be aware of what the podcast will do with this message, it should just poke it

kookster:
definitely a nit, but I think that would be ever so slightly better.

gcampo88:
ok, back to you @kookster !

PR #194, opened 2017-01-28
Import script fixes

kookster:
Found some issues while working on using the API directly to publish episodes
- [x] Need to start media files processing after an API episode save, like with message handlers
- [x] Audio files need a default status on create to have an episode status, otherwise all nil
- [x] Add a filter param on api podcasts controller to get the podcasts for a certain prx account

### DIFF
--- a/test/controllers/api/podcasts_controller_test.rb
+++ b/test/controllers/api/podcasts_controller_test.rb
@@ -106,4 +106,19 @@ describe Api::PodcastsController do
     get(:index, { api_version: 'v1', format: 'json' } )
     assert_response :success
   end
+
+  it 'should list podcasts for an account' do
+    podcast.id.wont_be_nil
+    get(:index, { api_version: 'v1', format: 'json', prx_account_uri: "/api/v1/accounts/#{account_id}" } )
+    assert_response :success
+    podcasts = JSON.parse(response.body)['_embedded']['prx:items']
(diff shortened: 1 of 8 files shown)


test/controllers/api/podcasts_controller_test.rb
```diff
@@ -106,4 +106,19 @@ class << @controller; attr_accessor :prx_auth_token; end
...
+  it 'should list podcasts for an account' do
+    podcast.id.wont_be_nil
+    get(:index, { api_version: 'v1', format: 'json', prx_account_uri: "/api/v1/accounts/#{account_id}" } )
+    assert_response :success
+    podcasts = JSON.parse(response.body)['_embedded']['prx:items']
+    podcasts.count.must_equal 1
+  end
+
+  it 'should not list podcasts for a different account' do
+    get(:index, { api_version: 'v1', format: 'json', prx_account_uri: "/api/v1/accounts/#{account_id}99" } )
```
gcampo88: Might want to add a `podcast.id.wont_be_nil` or some other call to `podcast` to make this test a little better! (without that, looks like the podcasts at `"/api/v1/accounts/#{account_id}"` have a count of 0 too)

kookster: good point.

kookster:
@gcampo88 I think that should do it

PR #198, opened 2017-02-22
fix how author appears in the feed

### LINKED ISSUE #197: <author> tag missing email address
@debenedictis commented on [Thu Jan 12 2017](https://github.com/PRX/dovetail.prx.org/issues/168)

According to the [RSS 2.0 SPECIFICATION](https://validator.w3.org/feed/docs/rss2.html#hrelementsOfLtitemgt) the `<author>` tag is the "Email address of the author of the item."  In [the millennial source feed](http://www.feedvalidator.org/check.cgi?url=http%3A%2F%2Fmillennial.prx.org%2Ffeed%2Fpodcast%2F) There is no `<author>` tag, but there are these two tags:
- `<itunes:author>Megan Tan &amp; Radiotopia</itunes:author>`
- `<itunes:email>prxwpadmin@prx.org</itunes:email>`

It seems that DT is populating the `<author>` tag from `<itunes:author>`.  For adherence to the RSS 2.0 it might be better to use `<itunes:email>` instead, or, to populate `<author>` by combining those two tags, e.g.:
-  `<author>prxwpadmin@prx.org (Megan Tan &amp; Radiotopia)</author>`

Please let me know what you think of this idea.

---

@farski commented on [Tue Jan 17 2017](https://github.com/PRX/dovetail.prx.org/issues/168#issuecomment-273265367)

Minor note: `&amp;` should not appear in that tag, since that tag doesn't support HTML content so that would be treated as the string literal ampersand-a-em-pea

---

@kookster commented on [Thu Feb 02 2017](https://github.com/PRX/dovetail.prx.org/issues/168#issuecomment-277046406)

The name & email that appear in the channel level itunes:author are not used at the episode level, which is why that email address is not used in the items. The author tags in the items are considered overrides of the channel, so it would not be right to mix the name from one with the email from the other.

In the episode/item, if there is an `<itunes:author>` tag specified, we also put this in the standard rss `<author>` tag. We do this whether that contains an email address or not, displaying the author name in parenthesis, but with no email address.

I can change it to only show the <author> tag if an email is specified, not just the name.

@farski the `&amp;` is in the feed from WP, feeder gets it from there, so that is where it would need to change.

kookster:
fixes #197

### DIFF
--- a/app/helpers/podcasts_helper.rb
+++ b/app/helpers/podcasts_helper.rb
@@ -1,0 +1,12 @@
+module PodcastsHelper
+  def episode_contact(type, ep)
+    name = ep.try("#{type}_name")
+    email = ep.try("#{type}_email")
+    result = nil
+    unless email.blank?
+      result = email
+      result = "#{result} (#{name})" unless name.blank?
+    end
+    result
+  end
+end
--- a/app/views/podcasts/show.rss.builder
+++ b/app/views/podcasts/show.rss.builder
@@ -88,9 +88,8 @@ xml.rss 'xmlns:atom' => 'http://www.w3.org/2005/Atom',
         xml.itunes :duration, ep.duration.to_i.to_time_summary if ep.media?
 
         if @podcast.display_full_episodes_count.to_i <= 0 || index < @podcast.display_full_episodes_count.to_i
-          unless ep.author_email.blank? && ep.author_name.blank?
-            xml.author "#{ep.author_email} (#{ep.author_name})"
-          end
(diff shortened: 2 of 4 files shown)


gcampo88:
Author contact stuff looks good! Did the updates to PRXAccess sneak in from your other PR?  Might vote to remove those changes from here and leave them in #115 if that's where they belong.

kookster:
ok @gcampo88 took your suggestion, and moved the prx access changes out

PR #200, opened 2017-03-01
Reject invalid story updates and publishes

### LINKED ISSUE #199: Don't let invalid episodes publish
- [ ] don't let an invalid episode publish
- [ ] don't accept invalid updates to an existing episode 
   - silently return early if `story.status !== COMPLETE`
- [ ] still need to be able to unpublish an invalid episode though, which comes through as an async update

gcampo88:
fixes #199

### DIFF
--- a/app/jobs/story_update_job.rb
+++ b/app/jobs/story_update_job.rb
@@ -18,7 +18,7 @@ class StoryUpdateJob < ActiveJob::Base
 
   def receive_story_update(data)
     # don't allow invalid episodes to update or publish
-    return if ['update', 'publish'].include?(action) && data[:status] == 'invalid'
+    return if ['update', 'publish'].include?(action) && data[:status] != 'complete'
 
     load_resources(data)
     episode ? update_episode : create_episode
--- a/app/jobs/story_update_job.rb
+++ b/app/jobs/story_update_job.rb
@@ -19,7 +19,7 @@ class StoryUpdateJob < ActiveJob::Base
   def receive_story_update(data)
     # don't allow invalid episodes to update or publish
     return if ['update', 'publish'].include?(action) && data[:status] == 'invalid'
-    
+
     load_resources(data)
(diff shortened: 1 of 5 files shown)


app/models/episode_story_handler.rb
```diff
@@ -54,7 +54,7 @@ def update_attributes
     episode.summary = sa[:description]
     episode.content = sa[:description]
     episode.categories = sa[:tags]
-    episode.published_at = Time.parse(sa[:published_at]) if sa[:published_at]
+    episode.published_at = sa[:published_at] ? Time.parse(sa[:published_at]) : nil
```
gcampo88: added this so that there would be a way to set `episode.published_at` to `nil` -- otherwise it kept retaining its previous value and staying published. Is the unpublish action / the setting of `published_at` to `nil` actually happening elsewhere in the code? I assume so but can't find it...

app/jobs/story_update_job.rb
```diff
@@ -17,6 +17,9 @@ def perform(*args)
   end
 
   def receive_story_update(data)
+    # don't allow invalid episodes to update or publish
+    return if ['update', 'publish'].include?(action) && data[:status] == 'invalid'
```
cavis: Hmm... what status does an `uploaded` / `failed` / `notfound` audio_file end up setting on the story?  We should catch those cases as well.

gcampo88: 👍

app/models/episode_story_handler.rb
cavis: In CMS, "unpublish" is just a simple attribute update.  So yes, I think this is how we need to handle that here.

cavis:
Not entirely sure if we should be also ignoring `status != COMPLETE` for create messages.  But I don't think it'll hurt anything, so let's give this a shot!

Issue #202, opened 2017-03-03
Migration fails b/c user can't create uuid extension


kookster:
```
/app $ rake db:migrate RAILS_ENV=staging
Migrating to CreateEpisodeImages (20170223114125)
== 20170223114125 CreateEpisodeImages: migrating ==============================
-- create_table(:episode_images)
   -> 0.0970s
-- enable_extension("uuid-ossp")
rake aborted!
StandardError: An error has occurred, this and all later migrations canceled:

PG::InsufficientPrivilege: ERROR:  permission denied to create extension "uuid-ossp"
HINT:  Must be superuser to create this extension.
: CREATE EXTENSION IF NOT EXISTS "uuid-ossp"
/usr/local/bundle/gems/activerecord-4.2.7.1/lib/active_record/connection_adapters/postgresql_adapter.rb:592:in `async_exec'
/usr/local/bundle/gems/activerecord-4.2.7.1/lib/active_record/connection_adapters/postgresql_adapter.rb:592:in `block in exec_no_cache'
/usr/local/bundle/gems/activerecord-4.2.7.1/lib/active_record/connection_adapters/abstract_adapter.rb:484:in `block in log'
/usr/local/bundle/gems/activesupport-4.2.7.1/lib/active_support/notifications/instrumenter.rb:20:in `instrument'
/usr/local/bundle/gems/activerecord-4.2.7.1/lib/active_record/connection_adapters/abstract_adapter.rb:478:in `log'
/usr/local/bundle/gems/activerecord-4.2.7.1/lib/active_record/connection_adapters/postgresql_adapter.rb:592:in `exec_no_cache'
/usr/local/bundle/gems/activerecord-4.2.7.1/lib/active_record/connection_adapters/postgresql_adapter.rb:584:in `execute_and_clear'
/usr/local/bundle/gems/activerecord-4.2.7.1/lib/active_record/connection_adapters/postgresql/database_statements.rb:160:in `exec_query'
/usr/local/bundle/gems/activerecord-4.2.7.1/lib/active_record/connection_adapters/postgresql_adapter.rb:338:in `enable_extension'
/usr/local/bundle/gems/activerecord-4.2.7.1/lib/active_record/migration.rb:665:in `block in method_missing'
/usr/local/bundle/gems/activerecord-4.2.7.1/lib/active_record/migration.rb:634:in `block in say_with_time'
[...]

PR #203, opened 2017-03-03
Do not add the uuid extension

kookster:
In RDS, it must be added by a superuser before this migration is run.
#202


gcampo88:
👍

PR #206, opened 2017-03-04
Get image info for large images

kookster:
- [x] add http timeout
- [x] web-like useragent
- [x] raise errors when FastImage fails
- [x] log them to NewRelic and log

#207


app/models/concerns/image_file.rb
```diff
@@ -64,7 +64,18 @@ def reset_image_attributes
...
-    info = FastImage.new(original_url)
+    info = nil
+    begin
+      fastimage_options = {
+        timeout: 10,
+        raise_on_failure: true,
+        http_header: { 'User-Agent' => 'Mozilla/5.0 (Macintosh; Intel Mac OS X) PRX Feeder/1.0' }
+      }
+      info = FastImage.new(original_url, fastimage_options)
+    rescue FastImageException => err
```
gcampo88: Do you need to define this anywhere? And do we want a test for it? (Just tried testing it out and got *** NameError Exception: uninitialized constant ImageFile::FastImageException ***)

kookster:
@gcampo88 yeah, good call, sloppy not to ave a test for that, fixed it and added one.

PR #208, opened 2017-03-06
Fix error in JSON parsing

gcampo88:
We were trying to access check `body[:status]` instead of `body['status']`, and so all async episode updates were rejected. Instead, since we're turning the data into a Story anyway, just check the story's `status`.


kookster:
good direction, now that it comes after load resources, good to just check the story.

Issue #219, opened 2017-03-07
Description should be wrapped in CDATA


kookster:
It can have html in it, wrap it in cdata - the podcast & episode descriptions

Issue #216, opened 2017-03-06
Loosen find_existing_image check


cavis:
Currently, uploading the same image-name over and over doesn't trigger feeder to re-process that image.  May want to loosen the restrictions in `Episode.find_existing_image` and `Podcast.find_existing_image` to allow reprocessing if any part of the url has changed.  (Not just the filename).

PR #222, opened 2017-03-07
Feed import fixes

kookster:
- [x] #219 description wrapped in CDATA
- [x] #216 loosen check for image dupes - match entire url not just file name


kookster:
ok - NOW this is ready @gcampo88 @cavis

app/views/podcasts/show.rss.builder
```diff
@@ -14,7 +14,7 @@ xml.rss 'xmlns:atom' => 'http://www.w3.org/2005/Atom',
     xml.language @podcast.language || 'en-us'
     xml.copyright @podcast.copyright unless @podcast.copyright.blank?
     xml.webMaster @podcast.web_master unless @podcast.web_master.blank?
-    xml.description @podcast.description unless @podcast.description.blank?
+    xml.description { xml.cdata!(@podcast.description || '') }
```
gcampo88: do we need `unless @podcast.description.blank?` ? So we don't have an empty <description>?

PR #223, opened 2017-03-08
Use storage audio urls

kookster:
- [x] From story messages, set audio_file.original url from `prx:storage` href (e.g. s3://bucket/path/file.mp3)
- [x] Always use the original url on the audio file to send to fixer, instead of calling cms api
- [x] Stop worrying about relative cms api urls, this is long fixed

### DIFF
--- a/app/models/episode_story_handler.rb
+++ b/app/models/episode_story_handler.rb
@@ -63,7 +63,7 @@ class EpisodeStoryHandler
     audio.each do |a|
       next unless a.status == 'complete'
 
-      existing_content = episode.find_existing_content(a.position, a.links['enclosure'].href)
+      existing_content = episode.find_existing_content(a.position, a.links['prx:storage'].href)
       if existing_content
         update_content(existing_content, a)
       else
@@ -88,8 +88,8 @@ class EpisodeStoryHandler
 
   def update_content(c, audio)
     c.position = audio.attributes['position']
-    c.href = cms_url(audio.links['enclosure'].href)
-    c.mime_type = audio.links['enclosure'].type || audio.attributes['content_type']
+    c.href = audio.links['prx:storage'].href
+    c.mime_type = audio.links['prx:storage'].type || audio.attributes['content_type']
     c.file_size = audio.attributes['size']
     c.duration = audio.attributes['duration']
     c.bit_rate = audio.attributes['bit_rate']
--- a/app/models/episode_story_handler.rb
+++ b/app/models/episode_story_handler.rb
@@ -75,7 +75,7 @@ class EpisodeStoryHandler
   end
 
   def update_image
-    if image_url = cms_url(story.objects['prx:image'].links['original'].href) rescue nil
+    if image_url = story.objects['prx:image'].links['original'].href rescue nil
       episode.images.build(original_url: image_url) if !episode.find_existing_image(image_url)
     else
       episode.images.destroy_all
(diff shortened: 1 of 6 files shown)


app/models/tasks/copy_media_task.rb
```diff
@@ -32,59 +32,25 @@ def task_status_changed(fixer_task, new_status)
   def task_options
     {
       job_type: 'audio',
-      destination: destination_url(media_resource),
-    }.merge(episode_options || story_options).with_indifferent_access
+      source: source_url(media_resource),
+      destination: destination_url(media_resource)
```
kookster: this class is just a whole lot simpler when it assumes it can always use the `original_url`, nice to remove some code!

test/fixtures/prx_story_all.json
```diff
@@ -1,4 +1,75 @@
 {
+  "appVersion": "v4",
```
kookster: I rearranged this to make it easier for me to look through, but the real change is the addition of the 4 `prx:storage` links, one to each audio file

test/fixtures/prx_story_all.json
```diff
@@ -101,6 +172,10 @@
               "profile": {
                 "href": "http://meta.prx.org/model/audio-file"
               },
+              "prx:storage": {
```
kookster: This is what we're using now instead of hitting the api

app/models/episode_story_handler.rb
```diff
@@ -63,7 +63,7 @@ def update_audio
     audio.each do |a|
       next unless a.status == 'complete'
 
-      existing_content = episode.find_existing_content(a.position, a.links['enclosure'].href)
+      existing_content = episode.find_existing_content(a.position, a.links['prx:storage'].href)
```
cavis: This will cause all existing stuff (in staging) to refresh content, yes?

test/fixtures/prx_story_all.json
```diff
@@ -101,6 +172,10 @@
...
+                "href": "s3://mediajoint.production.prx.org/public/audio_files/1200648/lcs_spring16_act1.mp3",
+                "type": "audio/mpeg"
```
cavis: I'm remembering that evaporate uploads (at least from Chrome) will`audio/mp3` in the temp S3 bucket.  And the cms-audio-lambda will either copy that, or detect `ffprobe.format = mp3` ... and set it to `audio/mp3` anyways.

Is that incorrect?  Should we be using the probably-more-compatible `audio/mpeg` there too?

app/models/episode_story_handler.rb
kookster: I don't think so - only if the filename has changed:
```
  def find_existing_content(pos, url)
    return nil if url.blank?
    content_file = URI.parse(url || '').path.split('/').last
    all_contents.
      where(position: pos).
      where('original_url like ?', "%/#{content_file}").
      order(created_at: :desc).
      first
  end
```

test/fixtures/prx_story_all.json
kookster: that's a chrome bug, afaict, https://bugs.chromium.org/p/chromium/issues/detail?id=227004
`audio/mpeg` is what I have seen in all the standards docs I checked

If the evap upload has 'audio/mp3', can we fix that in the lambda to be `audio/mpeg`?

kookster: https://github.com/PRX/cms-audio-lambda/issues/4

Issue #225, opened 2017-03-08
Strip all but link tags from episode summary


kookster:
apple allows only anchor html tags, and mailto: links don't work

kookster:
fixed by #227

PR #227, opened 2017-03-09
Sanitize text

kookster:
- [x] #225 fixes summary having only links, and sanitizes all the other text fields too
- [x] #226 adds a `summaryPreview` to the api responses when there is no summary
- [x] Fixes podcasts columns that should have been text to match episodes (summary, title, subtitle)
- [x] Do not override the summary with the story.description

### DIFF
--- a/app/views/podcasts/show.rss.builder
+++ b/app/views/podcasts/show.rss.builder
@@ -59,7 +59,7 @@ xml.rss 'xmlns:atom' => 'http://www.w3.org/2005/Atom',
     end
 
     xml.itunes :subtitle, @podcast.subtitle unless @podcast.subtitle.blank?
-    xml.itunes(:summary) { xml.cdata!(@podcast.summary) } unless @podcast.summary.blank?
+    xml.itunes(:summary) { xml.cdata!(itunes_summary(@podcast)) }
     xml.itunes :keywords, @podcast.keywords.join(',') unless @podcast.keywords.blank?
 
     xml.media :copyright, @podcast.copyright unless @podcast.copyright.blank?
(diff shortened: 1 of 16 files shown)


app/views/podcasts/show.rss.builder
```diff
@@ -59,7 +59,7 @@ xml.rss 'xmlns:atom' => 'http://www.w3.org/2005/Atom',
     end
 
     xml.itunes :subtitle, @podcast.subtitle unless @podcast.subtitle.blank?
-    xml.itunes(:summary) { xml.cdata!(@podcast.summary) } unless @podcast.summary.blank?
+    xml.itunes(:summary) { xml.cdata!(itunes_summary(@podcast)) }
```
gcampo88: Do we still want/need the check for blank?

app/views/podcasts/show.rss.builder
```diff
@@ -96,7 +96,7 @@ xml.rss 'xmlns:atom' => 'http://www.w3.org/2005/Atom',
           has_au_name = first_nonblank('author_name', [ep, @podcast])
           xml.itunes :author, has_au_name.author_name if has_au_name.author_name
 
-          xml.itunes(:summary) { xml.cdata!(ep.summary) } unless ep.summary.blank?
+          xml.itunes(:summary) { xml.cdata!(itunes_summary(ep)) }
```
gcampo88: ditto - Do we still want/need the check for blank?

app/views/podcasts/show.rss.builder
kookster: good question. hmm. I was thinking that between desc and summary we would always have something to put in here, since for a podcast description is a required attribute, but you know, people don't follow the rules, so I'm going to put a check back in.

app/views/podcasts/show.rss.builder
kookster: good call - title or description is required, so desc could be null.

kookster:
@gcampo88 ready for re-review!

PR #229, opened 2017-03-13
Fix episode link from libsyn to go to the enclosure

kookster:
When a feed uses libsyn, like twww, then the episode links are to the mp3 at libsyn.com
In cms, when these are imported, they get blanked out.
- [x] Change feeder story handler to leave these blank.
- [x] Change feeder entry handler to also blank these out from crier.
- [x] Change feeder rss view to fallback on enclosure url for the link when none is present

### DIFF
--- a/app/models/episode_entry_handler.rb
+++ b/app/models/episode_entry_handler.rb
@@ -62,7 +62,7 @@ class EpisodeEntryHandler
   def update_link
     self.episode.url = overrides[:feedburner_orig_link] || overrides[:url] || overrides[:link]
     # libsyn sets the link to a libsyn url, instead set to media file or page
-    self.episode.url = episode.media_url if episode.url && episode.url.match(/libsyn\.com/)
+    self.episode.url = nil if episode.url && episode.url.match(/libsyn\.com/)
   end
 
   def update_enclosure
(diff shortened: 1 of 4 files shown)


app/models/episode_entry_handler.rb
```diff
@@ -62,7 +62,7 @@ def update_dates
   def update_link
     self.episode.url = overrides[:feedburner_orig_link] || overrides[:url] || overrides[:link]
     # libsyn sets the link to a libsyn url, instead set to media file or page
-    self.episode.url = episode.media_url if episode.url && episode.url.match(/libsyn\.com/)
+    self.episode.url = nil if episode.url && episode.url.match(/libsyn\.com/)
```
gcampo88: update comment on line 64?

kookster:
@gcampo88 comment fixed

gcampo88:
@kookster do we care about the ones on lines 47 & 60? They make reference to depending on `media_url`, which is no longer the case, right?
...
update_contents
```Ruby
    update_image
    # must come after update_enclosure & update_contents, depends on media_url
    update_link
  end
```
and 
```Ruby
end

  # must be called after update_enclosure and update_contents
  # as it depends on media_url
  def update_link
    self.episode.url =
```

Issue #230, opened 2017-03-14
Remove stale comments


gcampo88:
app/models/episode_entry_handler.rb has some obsolete comments about a method relying on media_url

PR #231, opened 2017-03-16
Fix podcast images insert, comments

kookster:
- [x] This was not written correctly, had to fix the data after deploy. Here is the updated sql.
- [x] #230 Fix incorrect comments left in `episode_entry_handler.rb`


kookster:
@gcampo88 should be fixed now

PR #237, opened 2017-05-20
Add aws-secrets

kookster:
* install secrets in dockerfile
* load secrets in bin/application


kookster:
I have added this to feeder staging and prod, worker and web:
```
                {
                    "name": "LOCAL_ENV",
                    "value": "true"
                },
```

cavis:
Then away we go!

PR #241, opened 2017-05-26
Enable CI builds

kookster:
- [x] Add a simplified docker compose config for ci
- [x] Add `.prxci` file with instructions to build and test the feeder image
- [x] Wait for the db to start before running tests


bin/application
```diff
@@ -28,47 +22,60 @@ UpdateEnvironment () {
   fi
 }
 
+WaitForPostgres () {
```
cavis: Neat!

PR #244, opened 2017-05-31
Check if file name or the path preceding it have changed

### LINKED ISSUE #243: Consider file audio file updated if file plus immediate path are different
right now if the file name is the same, considered the same.

this is b/c we had incidents where the domain or prefix for urls changed, and we didn't want to consider those changes and regenerate all the audio.

also take into consideration the path segment before the filename.
this is much more unique, but will not change when a prefix is applied.
it will catch when an audio file is replaced via the cms.

kookster:
fixes #243

### DIFF
--- a/app/models/episode.rb
+++ b/app/models/episode.rb
@@ -263,10 +263,11 @@ class Episode < BaseModel
 
   def find_existing_content(pos, url)
     return nil if url.blank?
-    content_file = URI.parse(url || '').path.split('/').last
+    content_file = URI.parse(url || '').path.split('/')[-2, 2].join('/')
+    content_file = "/#{content_file}" unless content_file[0] == '/'
     all_contents.
       where(position: pos).
-      where('original_url like ?', "%/#{content_file}").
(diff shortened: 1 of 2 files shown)


app/models/episode.rb
```diff
@@ -263,10 +263,11 @@ def update_contents(files)
 
   def find_existing_content(pos, url)
     return nil if url.blank?
-    content_file = URI.parse(url || '').path.split('/').last
+    content_file = URI.parse(url || '').path.split('/')[-2, 2].join('/')
```
cavis: So you're assuming this isn't going to be a root-file then?  (Like `http://prx.org/file.mp3`)

cavis: Ah, but then you'd end up doing `["", "/file.mp3"].join("/")`.  So that's fine!

cavis: Also noting that the `url || ''` isn't actually necessary.  And if url ever was blank, that would throw an error.

PR #250, opened 2017-07-21
Update feed publish date on release of episode

kookster:


kookster:
aww yeah - 1 line code change, 29 lines of testing

PR #253, opened 2017-08-31
Whitelist specific chars for keyword_xid and keywords

gcampo88:
Addresses DT-Adzerk issues with keywords containing single quotes; whitelists only alphanumeric characters and spaces, dashes, and underscores.


app/models/episode.rb
```diff
@@ -309,30 +309,28 @@ def set_external_keyword
...
-                                         .strip
-                                         .slice(0, 20)
+    identifiers << sanitize_keyword(title || 'undefined', 20)
     self.keyword_xid = identifiers.join('_')
   end
 
+  def sanitize_keyword(kw, length)
+    kw.to_s
+      .downcase
+      .gsub(/[^ a-z0-9_-]/,' ')
```
cavis: In dovetail, @kookster is just blanking these out `''` rather than spaces `' '`.  Do we want to do the same here?

Issue #255, opened 2017-09-05
Support new iTunes tags


gcampo88:
Apple Podcasts now supports a number of new tags with iOS 11. We need to add these to Feeder and to [Publish](https://github.com/PRX/publish.prx.org/issues/428). 

**Channel** (podcast-level):
- `<itunes:type>`:  The type of podcast you are submitting. Where type can be one of the following: 
• **episodic** (default). Specify episodic for stand-alone episodes or when you want your episodes presented and recommended last-to-first. This option also supports seasons, with the newest episode at the top. For new subscribers, Apple Podcasts adds the latest, most recent episode in their Library. 
• **serial**. Specify serial when you want your episodes presented and recommended first-to-last. This option supports narratives, storytelling, thematic, and multiple seasons. For new subscribers, Apple Podcasts adds the first episode to their Library, or the entire current season if using seasons.

**Item** (episode-level):
- `<itunes:season>` A season number. Where season is a non-zero integer (1, 2, 3, etc.) representing your season number. To allow the season feature for shows containing a single season, if only one season exists in the RSS feed, Apple Podcasts does not display a season number. When you add a second season to the RSS feed, Apple Podcasts displays the season numbers. 
- `<itunes:episode>` An episode number. Where episode is a non-zero integer (1, 2, 3, etc.) representing your episode number. Use this tag to specify the recommended order for episodes within a season. 
- `<itunes:title>` An episode title. Where title is a string containing a clear, concise name for your episode. Apple recommends you not specify your podcast title, episode number, or season number in this tag. 
- `<itunes:episodeType>` The type of episode you are submitting. Where episodeType can be one of the following: 
 • **full** (default). Specify full when you are submitting the complete content of a show. 
 • **trailer**. Specify trailer when you are submitting a short, promotional piece of content that represents a preview of a show.  
 • **bonus**. Specify bonus when you are submitting extra content for a show (for example, behind the scenes information or interviews with the cast).

Source: http://podcasts.apple.com/resources/spec/ApplePodcastsSpecUpdatesiOS11.pdf

db + models:
- [x] add a boolean serial v. episodic to podcasts in db (default: episodic); set all current feeder podcasts to episodic
- [x] add season, number, and episode_type type (default: full) to episodes in db
- [x] validate `episode_type` in episode model
- [X] add a full_title method that concats season, number, and title, for usage in RSS in place of current title method. so <title> will be "Season X Episode Y {Title}" and <itunes:title> will be "Title".
[...]

PR #257, opened 2017-09-27
Support new Apple Podcasts tags for iOS11

gcampo88:
#255 

along with this, we'll change CMS to keep track of a story's season as well (https://github.com/PRX/cms.prx.org/issues/319), possibly send the season + episode info to feeder when it creates an EpisodeDistribution for a story (tbd, question in #319), and also update Publish to allow users to set all these fields. (https://github.com/PRX/publish.prx.org/issues/428). This is a blocker for #428.

### DIFF
--- a/app/models/episode.rb
+++ b/app/models/episode.rb
@@ -46,6 +46,9 @@ class Episode < BaseModel
 
   scope :published, -> { where('published_at IS NOT NULL AND published_at <= now()') }
 
+  alias_attribute :number, :episode_number
+
+
   def self.release_episodes!(options = {})
     podcasts = []
     episodes_to_release.each do |e|
--- a/app/models/episode.rb
+++ b/app/models/episode.rb
@@ -39,6 +39,8 @@ class Episode < BaseModel
 
   validates :podcast_id, :guid, presence: true
   validates :itunes_type, inclusion: { in: %w(full trailer bonus) }
+  validates :episode_number, numericality: { only_integer: true }, allow_nil: true
+  validates :season_number, numericality: { only_integer: true }, allow_nil: true
 
   before_validation :initialize_guid, :set_external_keyword, :sanitize_text
 
@@ -47,7 +49,7 @@ class Episode < BaseModel
   scope :published, -> { where('published_at IS NOT NULL AND published_at <= now()') }
 
   alias_attribute :number, :episode_number
-
+  alias_attribute :season, :season_number
 
   def self.release_episodes!(options = {})
     podcasts = []
--- a/app/models/episode.rb
+++ b/app/models/episode.rb
@@ -342,4 +342,11 @@ class Episode < BaseModel
   def feeder_cdn_host
     ENV['FEEDER_CDN_HOST']
   end
+
+  def full_title
+    ft = ''
+    ft += "Season #{season} " if season
+    ft += "Episode #{number}: " if number
+    ft + title
+  end
 end
--- a/app/models/episode.rb
+++ b/app/models/episode.rb
@@ -342,11 +342,4 @@ class Episode < BaseModel
   def feeder_cdn_host
     ENV['FEEDER_CDN_HOST']
   end
-
-  def full_title
-    ft = ''
-    ft += "Season #{season} " if season
-    ft += "Episode #{number}: " if number
-    ft + title
-  end
 end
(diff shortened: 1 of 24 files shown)


kookster:
are you also going to handle sync events from cms stories that include season and episode identifiers?

gcampo88:
will if you say so! Was wondering about that -- assume they should always be in sync with each other?

kookster:
yes, I can make sure cms sends that along in the announcement

gcampo88:
@kookster does this mean you're eager to do the review? 😁

kookster:
yes, I'm on it.

app/models/episode_story_handler.rb
```diff
@@ -60,6 +60,11 @@ def update_attributes
     episode.content = sa[:description]
     episode.categories = sa[:tags]
     episode.published_at = sa[:published_at] ? Time.parse(sa[:published_at]) : nil
+
+    season_num = sa[:season_identifier].to_i
+    episode.season_number = season_num unless season_num.blank? || season_num.zero?
```
kookster: might someone have a season 0? or an episode 0 for a preview episode? I think you need to check for blank  value before you call to_i

app/models/episode.rb
```diff
@@ -336,4 +342,11 @@ def sanitize_text
   def feeder_cdn_host
     ENV['FEEDER_CDN_HOST']
   end
+
+  def full_title
```
kookster: I think folks may have different ideas about how to include these values for their show.

so one way to go is add a truly new field, just for an itunes title (which feels duplicative, so I was originally against it)

or we may need to offer some kind of template so shows can have their own format for how to add the season and episode numbers to their titles.

e.g.
99% invisible: "247- Usonia the Beautiful"
template: "{episode_number}- {title}"

twww: "0.04: Big Block of Cheese III (with 99% Invisible)"
template: "{episode_number}.{season}: {title}"
(that is season 4, episode 0 by the way)

gcampo88:
ok @kookster, think this is ready for you again

db/migrate/20170926210603_add_episode_tags.rb
```diff
@@ -0,0 +1,16 @@
+class AddEpisodeTags < ActiveRecord::Migration
+  def up
+    add_column :episodes, :season_number, :integer
+    add_column :episodes, :episode_number, :integer
+    add_column :episodes, :itunes_type, :string, default: 'full'
+    execute "UPDATE episodes SET itunes_type = 'full'"
+    execute "UPDATE episodes SET itunes_type = 'bonus' WHERE title LIKE '%bonus%'"
+    execute "UPDATE episodes SET itunes_type = 'trailer' WHERE title LIKE '%trailer%'"
```
kookster: I'm not sure this is a good idea, I think maybe they should have to mark these themselves.

test/integration/podcast_request_test.rb
```diff
@@ -91,6 +91,28 @@
...
+                          clean_title: 'Stripped-down title')
+    end
+    @podcast.update_attributes(serial_order: false)
+    get "/podcasts/#{@podcast.id}"
+    @feed = Nokogiri::XML(response.body).css('channel')
+
+    @feed.at_css('itunes|type').text.must_equal @podcast.itunes_type
+    @feed.css('item').reverse.each_with_index do |node, ind|
+      node.css('title').text.must_match /Season \d+ Episode \d+/
+      node.css('itunes|title').text.wont_match /Season/
```
kookster: doesn't seem like you need this test if you are checking exact match on the next line

test/models/episode_story_handler_test.rb
```diff
@@ -32,5 +32,52 @@
...
+    episode.episode_number.must_equal 4
+    episode.clean_title.must_equal 'Stripped-down title'
   end
+
+  describe 'with episode identifiers' do
+
+    let(:zero_identifiers_story) do
+      msg = json_file(:prx_story_zero_identifiers)
+      body = JSON.parse(msg)
+      href = body['_links']['self']['href']
```
kookster: I just started using this `dig` method, and it is life changing
https://ruby-doc.org/core-2.3.0_preview1/Hash.html#method-i-dig

gcampo88: ooh boy. new favorite

gcampo88:
ok @kookster -- i removed the lines setting bonus or teaser episode type, but left the line setting all types to 'full' (since that is default anyway). LMK if you'd like me to nix that line too! also, used `dig` on that line and then did a find + replace for that exact line everywhere. :taco: @dig

Issue #273, opened 2018-01-17
Don't republish podcast feed during imports


gcampo88:
On a recent mega-import, Fixer became overwhelmed from attempting to build + publish the RSS for the importing feed 1.4k times. Don't publish the podcast til the import is done. (Not sure if this belongs here or in CMS). 
OR, debounce the publish-podcast-requests.

cavis:
Yeah, to keep down the number of fixer jobs, we'd have to keep the feed unpublished in Feeder until the import is done, then set it to public.  Maybe some sort of flag on the `podcasts` table, that Feeder checks internally before firing off any sort of background jobs?

kookster:
add a frozen attribute, don't publish new feed rss version when this is set.

Maybe also send back the current s3 file contents when this is set so the feed doesn't change, even if there is an accidental job trying to copy it

gcampo88:
the Feeder part of this is done, just have to add to CMS now => https://github.com/PRX/cms.prx.org/issues/399

PR #278, opened 2018-02-14
Lock feeds

kookster:
#273


db/migrate/20180214150849_add_locked_to_podcasts.rb
```diff
@@ -0,0 +1,15 @@
+class AddLockedToPodcasts < ActiveRecord::Migration
```
kookster: Had to add this to undo the prior migration as it was already deployed to staging by the PR test in CICD.

PR #280, opened 2018-03-03
Fix pinging feedburner

kookster:
- [x] Add new pinger class that actually works and doesn't throw errors
- [x] Add tests, how novel
- [x] Swap out the old updater for the new one


kookster:
feedburner has retired support for pinging.

PR #284, opened 2018-03-15
Call API on announce message

kookster:
instead of relying on message contents
also getting rid of the webmock conditional - should always be mocking


test/jobs/story_update_job_test.rb
```diff
@@ -4,8 +4,12 @@
 
   let(:podcast) { create(:podcast, prx_uri: '/api/v1/series/20829') }
 
-  let(:body) { json_file(:prx_story_updates) }
-  let(:msg) { { subject: 'story', action: 'update', body: body, sent_at: 1.second.ago } }
+  let(:prx_story_update) { json_file(:prx_story_updates) }
+  let(:prx_story_id) { 149726 }
+  let(:real_story_update) { json_file(:this_is_love) }
+  let(:real_story_id) { 235196 }
```
kookster: Are these kinds of labels what you had in mind @pkarman ?

pkarman: yes perfect @kookster

PR #286, opened 2018-03-27
User public uri for prx story, not with authorization in it

kookster:
- [x] Make sure there is not an /authorization/ in the prx uri

### DIFF
--- a/app/models/episode_story_handler.rb
+++ b/app/models/episode_story_handler.rb
@@ -33,19 +33,11 @@ class EpisodeStoryHandler
 
   def update_from_story(story)
     self.story = story
-    episode.prx_uri = clean_prx_uri(story.links['self'].href)
+    episode.prx_uri = Episode.story_uri(story)
 
     update_attributes
     update_audio
     update_image
-  end
-
-  def clean_prx_uri(url)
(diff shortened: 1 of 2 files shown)


app/models/episode_story_handler.rb
```diff
@@ -33,13 +33,21 @@ def update_from_story!(story)
...
 
     update_attributes
     update_audio
     update_image
   end
 
+  def clean_prx_uri(url)
+    result = url
+    if result && !result.match(/authorization/)
+      result = result.gsub('/', '/authorization/')
```
pkarman: this line replaces every `/` in the `url` with `/authorization/` - that isn't what I would expect a method called `clean_prx_uri` to do. Likewise, the name of this PR is "not with authorization in it" and this method is *adding* the `authorization`.

So I'm confused.

kookster: me too, not sure how my test passed locally. updating.

kookster: looks like `guard` was misbehaving - ok, fixing.

pkarman:
looks great @kookster - much clearer.

PR #292, opened 2018-04-12
Query for podcasts/episodes updated since a date

cavis:
Query param to filter/sort since a datetime.

### DIFF
--- a/app/controllers/api/podcasts_controller.rb
+++ b/app/controllers/api/podcasts_controller.rb
@@ -1,6 +1,8 @@
 # encoding: utf-8
 
 class Api::PodcastsController < Api::BaseController
+  include ApiUpdatedSince
+
   api_versions :v1
   represent_with Api::PodcastRepresenter
   filter_resources_by :prx_account_uri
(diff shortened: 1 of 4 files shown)


app/controllers/concerns/api_updated_since.rb
```diff
@@ -0,0 +1,31 @@
...
+
+  included do
+    class_eval do
+      allow_params :index, [:since]
+    end
+  end
+
+  def resources_base
+    if updated_since?
+      super.where('updated_at >= ?', updated_since).order('updated_at asc')
```
kookster: Hmm, wonder what this would do if someone tried to combine it with any other filters or sorts in the query params

app/controllers/api/podcasts_controller.rb
```diff
@@ -1,6 +1,8 @@
 # encoding: utf-8
 
 class Api::PodcastsController < Api::BaseController
+  include ApiUpdatedSince
```
kookster: I like concerns 👍

app/controllers/concerns/api_updated_since.rb
cavis: Fortunately ... feeder does neither `?filter` nor `?sort` right now!

PR #294, opened 2018-04-14
WIP - No more alpha fixer

kookster:
- [x] Use an activejob class in the worker process to copy the rss feed to s3
- [ ] Use a lambda to copy and analyze audio files
- [ ] Use a lambda to copy and analyze image files

Also:
- [x] Get rid of highwinds


pkarman:
@kookster what's the status of this?

kookster:
replaced with porter

PR #296, opened 2018-04-30
No fixer for feed copying

kookster:


app/jobs/publish_feed_job.rb
```diff
@@ -0,0 +1,61 @@
...
+    @rss = generate_rss_xml
+    save_podcast_file(@rss)
+  end
+
+  def setup_data(podcast)
+    @podcast = podcast
+    @episodes = @podcast.feed_episodes
+  end
+
+  def generate_rss_xml
```
cavis: Ah, okay, so you're manually generating the xml in the worker, rather than calling out to the web to render the view.

kookster: you got it.

cavis:
Update: the feeder aws user does have putObject permissions.

PR #304, opened 2018-07-06
Upgrade hal_api-rails

cavis:
Upgrade hal api gem, to get NewRelic error handling fixes.  Plus other minor-version fixes in the v0.3.x upgrade ... which were a bit more applicable to PRX/cms.prx.org#360 than here.

### DIFF
--- a/app/controllers/api/episodes_controller.rb
+++ b/app/controllers/api/episodes_controller.rb
@@ -56,8 +56,8 @@ class Api::EpisodesController < Api::BaseController
     visible
   end
 
-  def included(relation)
-    relation.with_deleted
+  def find_base
+    super.with_deleted
   end
 
   def process_media
--- a/app/representers/api/paged_collection_representer.rb
+++ b/app/representers/api/paged_collection_representer.rb
@@ -1,66 +1,6 @@
 # encoding: utf-8
+require 'hal_api/representer/collection_paging'
 
 class Api::PagedCollectionRepresenter < Api::BaseRepresenter
-  property :count
-  property :total
-
-  embeds :items,
-    decorator: lambda { |*| item_decorator },
-    class: lambda { |*| item_class },
-    zoom: :always
-
-  link :prev do
-    href_url_helper(params.merge(page: represented.prev_page)) unless represented.first_page?
-  end
-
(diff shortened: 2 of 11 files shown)


app/controllers/api/episodes_controller.rb
```diff
@@ -56,8 +56,8 @@ def visible?
     visible
   end
 
-  def included(relation)
-    relation.with_deleted
+  def find_base
```
cavis: This gets a bit into hal_api-rails internals ...

But apparently in the 0.3.x upgrade, `included` now gets called on index resources as well as show.  So to target the `with_deleted` only at show resources (so we get 410s instead of 404s), i moved to the `find_base` override.

This was the best place I could find to differentiate between the index-vs-show query.  The only other option I could see would be setting `@episode` myself ... but that feels less clean.  But maybe I missed something.

config/initializers/say_when.rb
```diff
@@ -16,14 +16,16 @@
...
-  SayWhen.schedule(
-    group: 'application',
-    name: 'release_episodes',
-    trigger_strategy: 'cron',
-    trigger_options: { expression: '0 0/5 * * * ?', time_zone: 'UTC' },
-    job_class: 'Episode',
-    job_method: 'release_episodes!'
-  )
+  unless Rails.env.test?
+    SayWhen.schedule(
```
cavis: This was being noisy in CI runs.

app/controllers/api/episodes_controller.rb
kookster: yeah, this makes sense, it does make me wonder if we need some better way for hal_api-rails to work for this.

app/representers/api/paged_collection_representer.rb
```diff
@@ -1,66 +1,6 @@
...
-      self.send(represented_url, options)
-    else
-      represented_url.to_s
-    end
-  end
-
-  def represented_url
-    @represented_url ||= represented.try(:url)
-  end
+  include HalApi::Representer::CollectionPaging
```
kookster: yeah, good call, we should be able to use the hal_api-rails code for this

PR #308, opened 2019-01-10
Unset season or episode if the value from cms is nil or less than 1

kookster:
Fixes a bug where a season value was removed from a story, but it wasn't blanked out on the feeder version of the episode.


app/models/episode_story_handler.rb
```diff
@@ -63,8 +63,8 @@ def update_attributes
     episode.published_at = sa[:published_at] ? Time.parse(sa[:published_at]) : nil
 
     %w(season episode).each do |time|
-      id = sa["#{time}_identifier"]
-      episode["#{time}_number"] = id.to_i if id && (!id.to_i.zero? || [0, '0'].include?(id))
+      id = sa["#{time}_identifier"].to_i
```
kookster: This is a safe operation, as if there is no key on the `sa[]` or if has the key, but a nil, blank, 0, or non-numeric value, the `.to_i` will turn any of those cases into the int `0`.

Both season` and `episode` should be integers and > 0, so if the value is 0, we set it to nil, and if it is greater than 0, then we use the numeric value.

test/models/episode_story_handler_test.rb
```diff
@@ -69,15 +69,23 @@
...
      podcast = create(:podcast, prx_uri: '/api/v1/series/32164')
      episode = EpisodeStoryHandler.create_from_story!(zero_identifiers_story)
-     episode.season_number.must_equal 0
-     episode.episode_number.must_equal 0
+     episode.season_number.must_be_nil
+     episode.episode_number.must_be_nil
    end
-  end
 
+   it 'blanks out identifiers on update' do
```
kookster: This is the case that was failing - there was a value, then an update came though that should have blanked it out.

Issue #311, opened 2019-04-16
More friendly API for updating media


cavis:
Currently, to change the episode `media` via the API, you have to change something other than the filename in the `{href: "the-url"}`.

Options:

1. Make any change _including_ filename to the href trigger an update.  Make sure this doesn't cause a bunch of churn from Crier/CMS announces.
2. Make any media coming in from the API force an update.  Reprocess the source files even if the href didn't change.

PR #312, opened 2019-05-31
API media updates

cavis:
For #311.

When setting `{ media: [] }` through an API create/update, always append new Contents instead of looking for existing.  This ensures that any media change coming through the API always reprocesses the files.

This _will_ result in some churn in the resulting mp3 urls, and thus the dovetail arrangement digests/urls.  So definitely don't want anyone automating a bunch of non-change PUTs to the episode media.  But I don't _think_ anyone is doing that right now anyways.  Should only PUT `media:[]` if it's actually changed.

### DIFF
--- a/app/models/episode.rb
+++ b/app/models/episode.rb
@@ -258,33 +258,20 @@ class Episode < BaseModel
     end
   end
 
+  # API updates for media= ... just append new files and reprocess
   def media_files=(files)
-    update_contents(files)
-  end
-
-  def update_contents(files)
     ignore = [:id, :type, :episode_id, :guid, :position, :status, :created_at, :updated_at]
-    ignore_nil = [:mime_type, :file_size, :duration]
     files.each_with_index do |f, index|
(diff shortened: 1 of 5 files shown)


app/models/episode.rb
```diff
@@ -258,33 +258,20 @@ def media_files
     end
   end
 
+  # API updates for media= ... just append new files and reprocess
   def media_files=(files)
-    update_contents(files)
-  end
-
-  def update_contents(files)
```
cavis: This method wasn't actually called anywhere apart from `media_files =`, through the HalApi/Roar representer stuff.  So I rewrote it to just _always_ append new Contents, even if the url hasn't changed.  (It'll do the usual thing where it waits for the fixer callback to swap the files, though).

The `EpisodeEntryHandler` and `EpisodeStoryHandler` have their own "check if the files changed" logic built into them, but they _do_ rely on the `find_existing_content` method, which I've left intact.

PR #314, opened 2019-07-24
updates iTunes Categories

sandikbarr:
Updates categories per https://help.apple.com/itc/podcasts_connect/#/itc9267a2f12

### DIFF
--- a/db/migrate/20190724191053_update_itunes_categories.rb
+++ b/db/migrate/20190724191053_update_itunes_categories.rb
@@ -1,5 +1,5 @@
 class UpdateITunesCategories < ActiveRecord::Migration
-  def change
+  def up
     ITunesCategories.find_each do |category|
       # 'Arts' sub 'Literature' -> 'Books'
       if category.name == 'Arts' && category.subcategories.include? = 'Literature'
@@ -179,5 +179,9 @@ class UpdateITunesCategories < ActiveRecord::Migration
         category.save!
       end
     end
+  end
+
+  def down
+    puts "No going back"
   end
 end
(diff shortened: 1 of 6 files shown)


db/migrate/20190724191053_update_itunes_categories.rb
```diff
@@ -0,0 +1,187 @@
...
+        category.subcategories.delete_at(category.subcategories.index('Podcasting'))
+      end
+
+      # 'Technology' sub 'Software How-To' discontinued
+      if category.name == 'Technology' && category.subcategories.include?('Software How-To')
+        category.subcategories.delete_at(category.subcategories.index('Software How-To'))
+      end
+
+      if category.changed?
+        puts "updated category[#{category.id}] for podcast[#{category.podcast}] => #{category.changes}"
```
sandikbarr: Log outputs nothing for `category.podcast`. Not very useful for communicating to support which podcasts are being updated.

cavis: Hmm... it's possible we have categories hanging around for deleted podcast_ids.  Or your dev environment is missing the `podcasts` table?

Either way - logging `#{category.podcast_id}` will just log the foreign key out of the database, instead of joining the podcasts table at all.  And is probably enough info for our case?

sandikbarr: Yes, I actually found later that I am missing Podcast data. 💯 good guess. I already got Paloma this information and she is reviewing it, so we can probably look at this logging as needed.

PR #316, opened 2019-10-01
Add state machine prototype job for image copies

farski:


app/models/tasks/copy_image_task.rb
```diff
@@ -3,9 +3,37 @@ def start!
...
+
+    send_rexif_job
+
     save!
   end
 
+  def send_rexif_job
+    sns = Aws::SNS::Client.new
+    sns.publish({
+      topic_arn: ENV['REXIF_JOB_EXECUTION_SNS_TOPIC'],
```
cavis: Is it worth guarding this entire `sns.publish()` call with a check for this ENV?

app/models/tasks/copy_image_task.rb
```diff
@@ -3,9 +3,36 @@ def start!
...
     job = fixer_copy_file(options)
     self.job_id = job[:job][:id]
+
+    send_rexif_job
+
     save!
   end
 
+  def send_rexif_job
+    if ENV['REXIF_JOB_EXECUTION_SNS_TOPIC']
```
kookster: appreciate the "feature flag" use of the env var

app/models/tasks/copy_image_task.rb
```diff
@@ -3,9 +3,36 @@ def start!
...
+      sns.publish({
+        topic_arn: ENV['REXIF_JOB_EXECUTION_SNS_TOPIC'],
+        message: JSON.dump({
+          Job: {
+            Id: self.job_id,
+            Source: {
+              URI: image_resource.original_url
+            },
+            Copy: {
+              Destinations: [
```
kookster: this is probably a good change from the current fixer format of always using a URI - is the `Source` also able to use this same type of format, specifying s3, a bucket, and key?

app/models/tasks/copy_media_task.rb
```diff
@@ -4,9 +4,36 @@ def start!
     self.options = task_options
     job = fixer_copy_file(options)
     self.job_id = job[:job][:id]
+
+    # send_rexif_job
```
kookster: not doing the media files too? or planning to expand this PR?

farski: Just starting small. Assuming nothing blows up with the image tasks this should get enabled in short order.

app/models/tasks/copy_image_task.rb
farski: @kookster I've made this change, but following the code from the task subclasses through a polymorphic association I'm not quite sure what format the `source: source_url(media_resource)` and `source: image_resource.original_url` are in. Are they always `s3://urls`?

PR #317, opened 2019-10-07
Add multi mode handling of image task sources

farski:


app/models/tasks/copy_image_task.rb
```diff
@@ -12,9 +12,21 @@ def start!
   def send_rexif_job
     return if !image_resource.original_url
 
-    parts = image_resource.original_url.gsub(/^s3:\/\//, '').split('/', 2)
-    source_bucket_name = parts[0]
-    source_object_key = parts[1]
+    if image_resource.original_url.match(/^s3:/)
```
cavis: Would be nicer to just move this to a separate method, and return nil in the `else` (and maybe `logger.warn()` it?)

Then

```ruby
if rexif_source && ENV['REXIF_JOB_EXECUTION_SNS_TOPIC']
```

app/models/tasks/copy_image_task.rb
```diff
@@ -12,9 +12,21 @@ def start!
...
+      parts = image_resource.original_url.gsub(/^s3:\/\//, '').split('/', 2)
+      source = {
+        Mode: 'AWS/S3',
+        BucketName: parts[0],
+        ObjectKey: parts[1]
+      }
+    elsif image_resource.original_url.match(/^https:/)
```
cavis: Should also allow http here ... `/^https?:/`

PR #332, opened 2019-12-17
Add `explicit_content` attribute

kookster:
Needed for https://github.com/PRX/dovetail.prx.org/issues/344

This PR adds an `explicit_content` read-only attribute to episode which is
`true` if the podcast is explicit and the episode doesn't specify, or the episode is explicit
`false` if the podcast is clean and the episode doesn't specify, or the episode is clean
otherwise `nil`

This can then be used in DT to target clean only, or avoid explicit episodes


brandonhundt:
Added to the Advanced Targeting epic for Grove for tracking.

PR #343, opened 2020-03-12
More episode meta

cavis:
Castle needs unpublished episodes from Feeder.  And a few extra fields just added to CMS:

- [x] Accept SNS announced updates for unpublished stories, whether or not they're `status = 'complete'`
    Previously, we weren't getting any updates for stories until the user uploaded audio and either published or scheduled them.  This allows all updates in _until the episode is published_.  Then only allow `status = complete` updates.
- [x] Record the audio-version for the episode (uri, label, and segment_count)
    NOTE: Dovetail (both old and new) have been expecting this `episode.audioVersion` key in the episode JSON.  Which it will attempt to match against a programs.json placements key.  We've long been planning to use this to manually indicate which placements-template to use for an episode.
    If there is no placements with that key, dovetail falls back to matching on num-original-segments.
- [x] Fixup the authorized podcasts/episodes endpoints, to have all url template parameters
- [x] Fixup auth podcasts/episodes to correctly apply the `?since=` scope
- [x] Fixup podcast/episode list endpoints to not have N+1 queries


cavis:
Not entirely sure how I'll backfill Feeder with these new fields, for existing episodes.  Would be nice to do it without touching them all.  But this should be able to merge prior to that plan!

app/models/media_resource.rb
```diff
@@ -43,8 +43,13 @@ def copy_media(force = false)
...
+    media_url_for_base(episode.base_published_url) if episode
+  end
+
+  def media_url_for_base(base_published_url)
     ext = File.extname(original_url || '')
     ext = '.mp3' if ext.blank?
-    "#{episode.base_published_url}/#{guid}#{ext}" if episode
+    "#{base_published_url}/#{guid}#{ext}"
   end
+
```
chrisrhoden: I'm not sure I see the practical change here, but it looks like this is covering a case where episodeA.media = mediaA and mediaA.episode != episodeA, am I getting that right?

cavis: This was tracking down an N+1 query.  When rendering the `/api/v1/episodes` list, it was: 

1. SQL-ing the single page of episodes from the episode table
2. Calling this for each episode
3. The `if episode` and `base_published_url` were actually reloading every episode in the list, individually by `WHERE ID = ?`.

cavis: (I'm curious if you have other ideas on how to prevent it reloading the episodes-by-id ... this was just the quick and dirty way for me ... "inject the episode")

PR #344, opened 2020-03-17
Feat/tune database pool

svevang:
Sets up a method to independently tune the feeder worker DB pool.

We can adjust some env vars to match Puma defaults:
```
# default number of threads in puma
DATABASE_POOL_SIZE=16
# bump the worker count back to where it was before debugging started
WORKER_DATABASE_POOL_SIZE=10
WORKER_COUNT=10
```


kookster:
fixed the `Gemfile.lock` merge with master, checking the build...

config/puma.rb
```diff
@@ -1,2 +1,4 @@
 port ENV['PORT'] || 3000
 environment ENV['RAILS_ENV'] || 'development'
+workers 1
+threads 0, 16
```
cavis: So it looks like the [0, 16] `threads` is the default.  But `workers` defaults to 0... so how is workers = 1 any different?  It switches to clustered mode, but with a single forked thread ... doesn't that amount to the same thing?

kookster: no, it's 2 processes, a master and a worker, instead of everything in 1 process:
https://github.com/puma/puma/issues/1364

kookster: which is to say, it's different from the default single mode, but I have no idea how differently this will perform - seems like the same draw on db connections, but with a different master thread, maybe easier to interrupt and manage? idk

PR #347, opened 2020-03-26
Porter api changes

cavis:
Update to newer Porter [better_iterator_failures](https://github.com/PRX/Porter) API _(merged to master now)_.  Basically, JobResults never fail.  Only the embedded tasks fail.

- [x] Remove alpha-fixer code completely.  Wooo!
    _Okay, except some of the filenames and ENVs are still a bit Fixer-y, because I'm too lazy to rename_
- [x] Update to new Porter API.  Plus better tests!  Factories!
- [x] Fix an annoyance that `content.task` was always returning the first CopyMediaTask instead of the latest one.


app/models/task.rb
```diff
@@ -29,10 +27,7 @@ def result
 
   def self.callback(msg)
     Task.transaction do
-      if job_id = fixer_callback_job_id(msg)
-        status = fixer_callback_status(msg)
-        time = fixer_callback_time(msg)
-      elsif job_id = porter_callback_job_id(msg)
+      if job_id = porter_callback_job_id(msg)
         status = porter_callback_status(msg)
```
cavis: So a few lines below here, we only update the task if the status is non-nil.  And only JobReceived and JobResult get a status - so we're ignoring the TaskResults / TaskErrors.

PR #351, opened 2020-04-14
Fill in and fix some authorized links

cavis:
Missing some links:

- [x] Adds `prx:authorization` link to the API root.
- [x] Fix the authorization podcast/episode self urls to stay under `/api/v1/authorization`
- [x] Keep the authorization podcast `prx:episodes` link under `/api/v1/authorization`
- [x] The `/api/v1/authorization/podcasts/1234/episodes` routes were broken

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -13,7 +13,7 @@ Rails.application.routes.draw do
 
       resource :authorization, only: [:show] do
         resources :podcasts, except: [:new, :edit], module: :auth do
-          resources :episodes, except: [:new, :edit], module: :auth
+          resources :episodes, except: [:new, :edit]
         end
 
         resources :episodes, except: [:new, :edit], module: :auth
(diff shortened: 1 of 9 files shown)


config/routes.rb
```diff
@@ -13,7 +13,7 @@
 
       resource :authorization, only: [:show] do
         resources :podcasts, except: [:new, :edit], module: :auth do
-          resources :episodes, except: [:new, :edit], module: :auth
```
chrisrhoden, line 16: I confess I don't know enough about rails routes.rb to know why this change with no apparent controller change - is it just redundant because of the parent route?

cavis: It was doubling the `auth` module ... so looking for the `Api::Auth::Auth::EpisodesController`.  (This route has never worked).

chrisrhoden: Got it!

chrisrhoden: Got it!

PR #352, opened 2020-04-15
Upgrade PrxAuth::Rails gem

### LINKED ISSUE #350: Upgrade prx_auth-rails gem & update code for checking permissions for new API

chrisrhoden:
closes #350

### DIFF
--- a/app/models/authorization.rb
+++ b/app/models/authorization.rb
@@ -22,12 +22,12 @@ class Authorization
   end
 
   def cache_key
-    token_key = OpenSSL::Digest::MD5.hexdigest(token.attributes.flatten.join)
+    token_key = OpenSSL::Digest::MD5.hexdigest([token.scopes, token.resources].map(&:to_s).join(''))
     ActiveSupport::Cache.expand_cache_key(['PRX::Authorization', token_key])
   end
 
   def token_auth_account_ids
-    token.authorized_resources.try(:keys) || []
+    token.resources
   end
 
   def token_auth_account_uris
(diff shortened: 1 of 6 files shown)


config/initializers/prx_auth.rb
```diff
@@ -0,0 +1,5 @@
+require 'prx_auth/rails'
+
+PrxAuth::Rails.configure do |config|
+  config.install_middleware = ENV['ID_HOST'].blank?
```
cavis, line 4: No feeder namespace here?  Oh, I guess `PrxAuth::Rails::Configuration.initialize` will guess the namespace from your rails app.  Does that not work over in CMS or something?

Also I'm still a little confused why you only install the middleware if ID_HOST is blank - shouldn't it be the opposite?

app/models/authorization.rb
```diff
@@ -22,12 +22,12 @@ def persisted?
   end
 
   def cache_key
-    token_key = OpenSSL::Digest::MD5.hexdigest(token.attributes.flatten.join)
+    token_key = OpenSSL::Digest::MD5.hexdigest([token.scopes, token.resources].map(&:to_s).join(''))
```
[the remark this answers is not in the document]
chrisrhoden: https://github.com/PRX/rack-prx_auth/issues/21

config/initializers/prx_auth.rb
chrisrhoden: the cms application is called Prx, so I use the namespace option to make it :cms instead of :prx - :feeder is already inferred here, but I can make it explicit if you prefer.

I answered the second question in cms, but short version is that we disable middleware installation when we are installing our own version to manually set the ID host

cavis: Makes sense!  +1

PR #363, opened 2020-08-07
Scope-based policies

chrisrhoden:
This is the first PR that is ready for review to start using the new scope-based tokens for checking permissions.

There are a couple of potential issues I want to flag: first, I have left things requiring the `admin` meta-scope to allow deleting episodes. That's because the existing behavior only allowed admins to delete episodes, but we didn't have a real scope that seemed to fit that case. I think this is at best a stopgap and we should plan to retire that scope and determine what we want to do going forward (personally, I think you can do as much damage editing an episode as deleting it but maybe I don't know everything going on here...)

Second, this is the introduction of draft semantics and they may not work as you initially assume, so probably good to really make sure you're following how it works.

I guess last, I assume that the Authorization model is really only used for reads (that's how it looks to me, and I dug pretty deep) but if there's somewhere that it's being used to further constrain writes (e.g. building models based on the scopes here) then obviously we might run into problems when someone has write access without read access. That feels pretty contrived.

### DIFF
--- a/app/policies/episode_policy.rb
+++ b/app/policies/episode_policy.rb
@@ -1,10 +1,11 @@
 class EpisodePolicy < ApplicationPolicy
   def create?
-    token && token.authorized?(account_id)
+    update?
   end
 
   def update?
-    token && token.authorized?(account_id)
+    token&.authorized?(account_id, :episode) ||
+      ( token&.authorized?(account_id, :episode_draft) && resource.draft? && resource.was_draft? )
   end
 
   def destroy?
--- a/app/policies/podcast_policy.rb
+++ b/app/policies/podcast_policy.rb
@@ -1,14 +1,14 @@
 class PodcastPolicy < ApplicationPolicy
   def create?
-    token && token.authorized?(account_id)
+    token&.authorized?(account_id, :podcast_create)
   end
 
   def update?
-    token && token.authorized?(account_id)
+    token&.authorized?(account_id, :podcast_edit)
   end
 
   def destroy?
-    token && token.authorized?(account_id, :admin)
+    token&.authorized?(account_id, :podcast_delete)
   end
 
   def account_id
--- a/test/controllers/api/episodes_controller_test.rb
+++ b/test/controllers/api/episodes_controller_test.rb
@@ -81,7 +81,7 @@ describe Api::EpisodesController do
     let(:podcast) { create(:podcast, prx_account_uri: "/api/v1/accounts/#{account_id}") }
     let(:episode_redirect) { create(:episode, podcast: podcast, published_at: nil) }
     let(:episode_update) { create(:episode, podcast: podcast, published_at: nil) }
-    let(:token) { StubToken.new(account_id, ['member']) }
+    let(:token) { StubToken.new(account_id, ['member feeder:read-private feeder:podcast-edit feeder:podcast-create feeder:episode feeder:episode-draft']) }
 
     around do |test|
       class << @controller; attr_accessor :prx_auth_token; end
--- a/app/models/authorization.rb
+++ b/app/models/authorization.rb
@@ -27,7 +27,7 @@ class Authorization
   end
 
   def token_auth_account_ids
-    token.resources
+    token.resources(:read_private)
   end
 
   def token_auth_account_uris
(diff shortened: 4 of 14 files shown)


app/policies/episode_policy.rb
```diff
@@ -1,10 +1,11 @@
 class EpisodePolicy < ApplicationPolicy
   def create?
-    token && token.authorized?(account_id)
+    update?
   end
 
   def update?
-    token && token.authorized?(account_id)
+    token&.authorized?(account_id, :episode) ||
+      ( token&.authorized?(account_id, :episode_draft) && resource.draft? && resource.was_draft? )
```
cavis, line 8: Does this imply a user with `episode-draft` can publish an episode ... but then cannot edit the episode after it's published (even if published_at is in the future)?

I see the case for preventing unpublishing - but why not let them edit it when it's published or released (with a future published_at)?  Mostly trying to remember the use cases around this...

test/controllers/api/episodes_controller_test.rb
```diff
@@ -81,7 +81,7 @@
     let(:podcast) { create(:podcast, prx_account_uri: "/api/v1/accounts/#{account_id}") }
     let(:episode_redirect) { create(:episode, podcast: podcast, published_at: nil) }
     let(:episode_update) { create(:episode, podcast: podcast, published_at: nil) }
-    let(:token) { StubToken.new(account_id, ['member']) }
+    let(:token) { StubToken.new(account_id, ['member feeder:read-private feeder:podcast-edit feeder:podcast-create feeder:episode feeder:episode-draft']) }
```
cavis, line 84: Why leave `member` in here?

cavis: Actually, I do see `member` and `admin` scopes in my JWT right now.  Are those for backwards compatibility or something?

app/policies/podcast_policy.rb
```diff
@@ -1,14 +1,14 @@
 class PodcastPolicy < ApplicationPolicy
   def create?
-    token && token.authorized?(account_id)
+    token&.authorized?(account_id, :podcast_create)
```
cavis, line 3: Not exactly tied to this line of code ... but just to think through the way we'll use auth-Feeder on day 1 ...

1. CMS distributions doing synchronous creates of podcasts/episodes.  This will get the account tied to [the series](https://github.com/PRX/cms.prx.org/blob/master/app/models/distributions/podcast_distribution.rb#L51) or [the story's series](https://github.com/PRX/cms.prx.org/blob/master/app/models/story_distributions/episode_distribution.rb#L56).  And since those aren't requesting any scope yet, they should just continue to get all scopes, and work.
2. `publish.prx.org` users updating podcasts/episodes.  But the JWTs already contain all the feeder specific scopes, so things should just continue to work as-is.
3. Announcement/SNS based updates.  Which don't rely on account auth at all.

app/policies/episode_policy.rb
chrisrhoden: They can't publish an episode at all - they are not able to touch episodes that have a published_at date at all, whether in the future or not.

I think I was sort of thinking this would be useful as an approval-like flow, or maybe for apps that do imports but you actually have to hit OK on (though that might be overloading the published_at somewhat)

test/controllers/api/episodes_controller_test.rb
chrisrhoden: Those are in tokens right now for compatibility with the apps as currently deployed (until this patch is released, and its partners in the other repos) - at some point in the future, we can remove the meta-scopes.

app/policies/podcast_policy.rb
cavis: Oh, and there's a (4) ...

Castle requesting the wildcard token to scrape all Feeder podcasts/episodes.  /me looks at that code.

cavis: Looks like it's a [wildcard read-private token](https://github.com/PRX/castle.prx.org/blob/master/lib/feeder/api.ex#L12) ... and since that won't be namespaced (i think?) it should just work against these changes.  Right?

chrisrhoden: Yes, it should. Unnamespaced scopes are treated as `*:scope` which means the token would be equivalent to `*:'*:read-private'` which should work!

PR #368, opened 2020-08-18
Prevent podcast and episode theft

chrisrhoden:


app/policies/episode_policy.rb
```diff
@@ -4,15 +4,24 @@ def create?
...
 
+  private
+
   def account_id
     resource.podcast.account_id
   end
+
+  def account_id_was
+    if resource.podcast_id_changed? && resource.podcast_id_was.present?
+      Podcast.find(resource.podcast_id_was)
```
cavis, line 22: Ah, so also catching the case where they `set_podcast_uri`

app/policies/application_policy.rb
```diff
@@ -10,4 +10,23 @@ def update?
...
+      token&.authorized?(account_id, scope)
+    end
+  end
+
+  def account_id
+    resource.account_id
+  end
+
+  def account_id_was
+    nil
```
cavis, line 30: Why nil?  (vs `resource.account_id_was`)

chrisrhoden: Could go either way, I guess I was imagining that we shouldn't do this check by default but :shrug:

app/policies/episode_policy.rb
chrisrhoden: Yes - I actually think that account_id_was is unnecessary given the existing controllers but completeness doesn't hurt in this case.

PR #370, opened 2020-10-23
Update default feed webMaster value

### LINKED ISSUE #365: Update RSS webMaster value
Currently using the older `prxhelp@prx.org`, but we've mostly transitioned to `help@prx.org`

farski:
Closes #365

This value is not defined in aws-secrets for production, so the default change will apply.


cavis (approved):
👍

PR #373, opened 2020-12-14
Media restriction tag

### LINKED ISSUE #371: Media restriction tag
Spotify (and maybe others?) do honor the `media:restriction` tag.  Add some support for this in the feeder database/API.

https://www.rssboard.org/media-rss/#media-restriction

cavis:
Closes #371.

Add support for the (optional) `<media:restriction>` tag at the Podcast level.  Even though it's rendered in the RSS per-episode.

For podcasts with a non-nil restriction, you'll now have a nested `media:restriction` -

```xml
<item>
  <media:restriction type="country" relationship="allow">US CA GB</media:restriction>
</item>
```

Currently I'm restricting these to allowed-countries, as that's the only type Spotify will recognize.  Also validates the list of country codes.

And the JSON API will look like:

```js
# /api/v1/flights/1234
{
  "id": 1234,
  "restrictions": [
    {
      "type": "country",
      "relationship": "allow",
      "values": ["US", "CA", "GB"]
    }
  ]
}
```


app/views/podcasts/show.rss.builder
```diff
@@ -115,13 +115,20 @@ xml.rss 'xmlns:atom' => 'http://www.w3.org/2005/Atom',
...
-            fileSize: ep.file_size,
-            type: ep.content_type,
-            url: ep.media_url) if ep.media?
-
           xml.content(:encoded) { xml.cdata!(ep.content) } unless ep.content.blank?
         end
+
+        if ep.media?
+          if @podcast.restrictions.present?
+            xml.media(:content, fileSize: ep.file_size, type: ep.content_type, url: ep.media_url) do
```
cavis, line 123: Discovered that Spotify _does_ include non-full episodes in their app.  So they must fall back to the enclosure-url if the media:content is not present.

Had to move the `media:content` block(s) down here, so we can also include it for non-full episodes of a restricted podcast.

app/views/podcasts/show.rss.builder
```diff
@@ -115,13 +115,20 @@ xml.rss 'xmlns:atom' => 'http://www.w3.org/2005/Atom',
...
+              @podcast.restrictions.each do |r|
```
kookster, line 124: I think this should just be in the item, not in the `media:content` block?

the examples in the mrss spec and the spotify docs show `media:restriction` as a child tag of item, not contained in a `media:content` block:
https://podcasters.spotify.com/terms/Spotify_Podcast_Delivery_Specification_v1.6.pdf
https://www.rssboard.org/media-rss#media-restriction

app/models/media_restrictions_validator.rb
```diff
@@ -0,0 +1,49 @@
...
+    end
+  end
+
+  private
+
+  def valid_restriction?(value)
+    value.is_a?(Hash) && %i(type relationship values).all? { |k| value.key?(k) }
+  end
+
+  # NOTE: Spotify only accepts allow+country ... so validate that for now
```
kookster, line 37: agreed

app/models/podcast.rb
```diff
@@ -114,6 +116,10 @@ def feed_episodes
     feed
   end
 
+  def display_full_episode?(index)
```
kookster, line 119: maybe should be a helper method instead? seems like a view concern?

db/schema.rb
```diff
@@ -209,6 +210,7 @@
     t.boolean  "locked",                      default: false
     t.boolean  "itunes_block",                default: false
     t.string   "feed_rss_alias"
+    t.string   "restrictions"
```
kookster, line 213: Seems unlikely to cross 256 for this purpose...
maybe this should be a `text` column since it is json?

kookster (changes requested):
I could be wrong, but I think the tag should be a child of `item` not `media:content`, and I wonder if the column type or method location ought to be slightly different.

app/views/podcasts/show.rss.builder
cavis: Oh, interesting.  In the spec:

> The following elements are optional and may appear as sub-elements of <channel>, <item>, <media:content> and/or <media:group>.

Would be convenient if we could just set it at the `channel` level.  But sounds like Spotify wants it at the `item`, so I guess that's what I should do!

app/models/podcast.rb
cavis: +1

db/schema.rb
cavis: Good point.  If we ever supported `type = uri` ... could hit that 256 pretty quick.

app/views/podcasts/show.rss.builder
cavis: Nevermind!  Should have been at `<item>` level.

cavis:
Okay, ready for re-review @kookster.  Moving up to the item level simplified the diff a good amount.

The only other weirdness I see - the examples given by both Spotify and the RSS Advisory Board list lowercase country codes.

Though [the spec](https://www.rssboard.org/media-rss#media-restriction) isn't entirely clear (and that link for iso3166 is broken) ... it's a bit unusual to represent a country with lowercase.  More typically that would be a language in lowercase.  If they actually meant `ISO 3166-1 alpha-2` those should be capitalized.

kookster (approved):
lgtm

PR #374, opened 2020-12-14
Fix minitest 6 deprecations

cavis:
Pretty much all our tests were using the deprecated `obj.must_equal` syntax.  The main choices are to wrap everything in an underscore `_(obj).must_equal` or convert to a `assert_equal obj, "str"` syntax.  I opted for the latter, as we've gone down that route in other repos.  (And I'm just not a fan of the underscore syntax in ruby).

Didn't make any attempt to rubocop/lint things.  Just finding/replacing.


svevang (approved):
Looks good!

PR #377, opened 2021-02-04
Change explicits back to true/false

cavis:
Will take some CMS and Publish changes deployed concurrently.

According to the [Apple Docs](https://help.apple.com/itc/podcasts_connect/#/itcb54353390), I guess their spec for `<itunes:explicit` has changed recently-ish:

1. Must be `"true"` or `"false"` now.  No blanks/yes/no/clean/explicit.
2. Is now required at the channel level

I opted to keep the field as a string, but migrate to true/false strings.  There is some nuance here, which I'll try to note below.


app/models/podcast.rb
```diff
@@ -27,6 +27,18 @@ class Podcast < BaseModel
...
+  # these keep changing - so just translate to the current accepted values
+  EXPLICIT_ALIASES = {
+    '' => 'false',
+    'no' => 'false',
+    'clean' => 'false',
+    false => 'false',
+    'yes' => 'true',
+    'explicit' => 'true',
+    true => 'true'
+  }.freeze
```
cavis, line 41: To avoid test churn, and things like feed-imports that could have older data ... I'm just translating into the current true/false scheme.

But note that podcasts now have a _default_ of `"false"` - since it's required.  So if we're importing a feed with no channel-level explicit tag, we're going to mark it as clean in feeder.

app/views/podcasts/show.rss.builder
```diff
@@ -87,6 +87,7 @@ xml.rss 'xmlns:atom' => 'http://www.w3.org/2005/Atom',
 
         xml.itunes :title, ep.clean_title unless ep.clean_title.blank?
         xml.itunes :subtitle, ep.subtitle unless ep.subtitle.blank?
+        # NOTE: you'll only get a tag if this was explicitly set (pun intended)
         xml.itunes :explicit, ep.explicit unless ep.explicit.blank?
```
cavis, line 91: This does _not_ fallback to the podcast tag right now.  Which is a bit contrary to @kookster's comments - that he'd prefer this tag always be present.

I think right now it _looks_ like the field is always present in our feeds because Publish explicitly sets this on every Episode.  (Setting it to the same value as the podcast - but not leaving it nil).

db/migrate/20210204184430_change_explicit_allowed_values.rb
```diff
@@ -0,0 +1,53 @@
...
+    all_ids = not_explicit_ids + explicit_ids
+
+    # set episodes where different than podcast
+    n3 = Episode.with_deleted.where(podcast_id: not_explicit_ids, explicit: NOT_EXPLICITS).update_all(explicit: nil)
+    n4 = Episode.with_deleted.where(podcast_id: not_explicit_ids, explicit: EXPLICITS).update_all(explicit: 'true')
+    n5 = Episode.with_deleted.where(podcast_id: explicit_ids, explicit: NOT_EXPLICITS).update_all(explicit: 'false')
+    n6 = Episode.with_deleted.where(podcast_id: explicit_ids, explicit: EXPLICITS).update_all(explicit: nil)
+
+    # set any episodes orphaned from their podcasts (for some reason)
+    orphaned = 'podcast_id IS NULL OR podcast_id NOT IN (?)'
```
cavis, line 27: I didn't realize doing `WHERE explicit NOT IN (1, 2, 3)` does not select rows where explicit = null.  Thanks SQL.

db/migrate/20210204184430_change_explicit_allowed_values.rb
```diff
@@ -0,0 +1,53 @@
...
+    # stats
+    puts "Total of #{Podcast.with_deleted.count} podcasts, #{Episode.with_deleted.count} episodes"
+    puts "  #{n1} clean podcasts"
+    puts "    #{n3} clean (nil) episodes"
+    puts "    #{n4} explicit (true) episodes"
+    puts "  #{n2} explicit podcasts"
+    puts "    #{n5} clean (false) episodes"
+    puts "    #{n6} explicit (nil) episodes"
+    puts "  #{n7} orphaned clean (false) episodes"
+    puts "  #{n8} orphaned explicit (true) episodes"
```
cavis, line 44: Running on a months-stale database locally, I got:

```
Total of 224 podcasts, 20105 episodes
  182 clean podcasts
    16885 clean (nil) episodes
    648 explicit (true) episodes
  42 explicit podcasts
    1662 clean (false) episodes
    879 explicit (nil) episodes
  31 orphaned clean (false) episodes
  0 orphaned explicit (true) episodes
```

test/models/episode_test.rb
```diff
@@ -200,9 +200,6 @@
       podcast.explicit = val
       assert_equal episode.explicit_content, false
     end
-
-    podcast.explicit = nil
-    assert_nil episode.explicit_content
```
cavis, line 205: Some no-longer-applicable tests, now that `podcast.explicit` can't be null.

app/models/episode.rb
```diff
@@ -115,16 +116,12 @@ def guid
...
-    e = (explicit.blank? && podcast) ? podcast.explicit : explicit
-    e = e.to_s.downcase.strip
-    if ["clean", "no", "false", "f"].include?(e)
-      false
-    elsif ["explicit", "yes", "true", "t"].include?(e)
-      true
-    else
-      nil
-    end
+    (explicit || podcast&.explicit) == 'true'
```
cavis, line 124: Oh, so this is interesting.  Dovetail (nodejs) actually uses this field, per #332.  [Dovetail actually concats](https://github.com/PRX/dovetail.prx.org/blob/master/lib/util/dovetail.js#L49) "explicit" or "clean" onto the episode's keywords, so they can be targeted in Adzerk.

The only change here is that `episode.explicitContent` will _never_ be null.  Checked with ad-ops, and that should be okay.

svevang (approved):

app/models/episode.rb
kookster: makes sense to me - if anything, will make such targeting more useful.

app/models/podcast.rb
kookster: Yeah, this might need to be a) a checklist item for onboarding, or b) updated in the UI to prompt for a value rather than default on import?

PR #380, opened 2021-05-05
Re-raise job errors

cavis:
We were sending job exceptions to New Relic, but not re-raising them.

Also did a bit of reading on [ActiveJob Retries](https://github.com/ruby-shoryuken/shoryuken/issues/553).  I _think_ what happens now is:

- A job raises an exception
- This handler catches it, logs to new relic, and re-raises
- The SQS message is _not_ deleted, but ends up hidden for 1 hour (our visibility timeout)
- After 10 receives/retries (10 hours?), the SQS message goes into the DLQ

### DIFF
--- a/app/jobs/application_job.rb
+++ b/app/jobs/application_job.rb
@@ -3,7 +3,11 @@ require 'newrelic_rpm'
 class ApplicationJob < ActiveJob::Base
 
   rescue_from(StandardError) do |e|
-    NewRelic::Agent.notice_error(e)
+    begin
+      NewRelic::Agent.notice_error(e)
+    ensure
+      raise e
+    end
   end
 
   around_perform do |job, block|


app/jobs/application_job.rb
```diff
@@ -3,7 +3,11 @@
 class ApplicationJob < ActiveJob::Base
 
   rescue_from(StandardError) do |e|
-    NewRelic::Agent.notice_error(e)
+    begin
+      NewRelic::Agent.notice_error(e)
+    ensure
+      raise e
```
kookster, line 9: This seems to be the main thing, I haven't tested this, but my understanding and from looking at the docs is that raised errors should lead to retries b/c the message won't be deleted, and so should reappear when it is no longer hidden in SQS

kookster (approved):
lgtm!

PR #390, opened 2021-06-25
Include deleted for wildcard token

cavis:
Castle uses a wildcard token to scrape Feeder data from the API, using the `?since=2021-06-25` query param to find updated_at >= some time.

We need castle to also be able to get deleted podcasts/episodes from that same API.

### DIFF
--- a/app/controllers/api/auth/episodes_controller.rb
+++ b/app/controllers/api/auth/episodes_controller.rb
@@ -26,6 +26,6 @@ class Api::Auth::EpisodesController < Api::EpisodesController
   end
 
   def resources_base
-    @episodes ||= authorization.token_auth_episodes.merge(super)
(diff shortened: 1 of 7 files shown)


app/controllers/api/auth/episodes_controller.rb
```diff
@@ -26,6 +26,6 @@ def visible?
   end
 
   def resources_base
-    @episodes ||= authorization.token_auth_episodes.merge(super)
+    @episodes ||= super.merge(authorization.token_auth_episodes)
```
cavis, line 29: Had to reverse the order of this merge, so the `token_auth_episodes` (which contains the `with_deleted`) can override the `super` query, which does not include deleted.

svevang (approved):
Looks good!

PR #392, opened 2021-07-14
Sort episodes by published_at

cavis:
Hit a bug with a script scraping the Feeder API -

Expected the most recent feed-rss episode to be first at `/api/v1/episodes`.  But because the episode was actually created 3 years ago and we're sorting by `id: :desc` ... it was way down the list.

This sorts both the authorized and public episode endpoints by published_at (and released_at) DESC.  I don't think this will cause any issues, with an API user depending on that `id desc` order.  The main feed-scraper is Castle, and since it uses `?since` it's already getting an explicit `order('updated_at asc')`.

### DIFF
--- a/app/controllers/api/auth/episodes_controller.rb
+++ b/app/controllers/api/auth/episodes_controller.rb
@@ -25,6 +25,10 @@ class Api::Auth::EpisodesController < Api::EpisodesController
     visible
   end
 
+  def sorted(res)
+    res.order('COALESCE(published_at, released_at) DESC NULLS LAST, id DESC')
+  end
+
   def resources_base
     @episodes ||= super.merge(authorization.token_auth_episodes)
   end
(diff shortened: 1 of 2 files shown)


app/controllers/api/auth/episodes_controller.rb
```diff
@@ -25,6 +25,10 @@ def visible?
     visible
   end
 
+  def sorted(res)
+    res.order('COALESCE(published_at, released_at) DESC NULLS LAST, id DESC')
```
kookster, line 29: there was no explicit sort on this before? wow, well, this is great.

kookster: oh, I see, b/c it's the API, not the actual rss feed - makes sense. This looks good!

kookster (approved):
lgtm!

PR #393, opened 2021-09-23
Use new queue name environment variables if present

farski:


cavis (approved):
👍  seems like that'll work!

PR #397, opened 2021-11-08
Upgrade the announce gem, use verify instead of configure broker

kookster:
This change allows the sns topics and sqs queues to be controlled by Cfn templates.


config/announce.yml
```diff
@@ -1,4 +1,5 @@
 app_name: feeder
+queue_name_prefix: <%= ENV['ANNOUNCE_RESOURCE_PREFIX'] || ENV['RAILS_ENV'] %>
```
cavis, line 2: Hrmm...

Could you instead delete this line, and change `application.rb` to:

```ruby
config.active_job.queue_name_prefix = ENV['ANNOUNCE_RESOURCE_PREFIX'] || ENV['RAILS_ENV']
```

I think you said, if you don't provide this yml config, it defaults to the active job `queue_name_prefix`, right?

config/shoryuken.yml
```diff
@@ -10,5 +10,5 @@ concurrency:         <%= ENV['WORKER_COUNT'] || 25 %>
 # The delay in seconds to pause a queue when it's empty. Default 0
 delay:               <%= ENV['WORKER_PAUSE'] || 30 %>
 queues:
-  - [<%= ENV['DEFAULT_JOB_QUEUE_NAME'] || "#{env}_feeder_default" %>, 2]
-  - [<%= ENV['FIXER_CALLBACK_QUEUE_NAME'] || "#{env}_feeder_fixer_callback" %>, 1]
+  - ["<%= prefix %>_feeder_default", 2]
+  - ["<%= prefix %>_feeder_fixer_callback" %>, 1]
```
cavis, line 14: I think you still need the `application.rb` change.  Otherwise it'll be queuing jobs to the old queues, but the workers will be polling these new ones.

cavis (changes requested):
Could the prefix just live in one place?  (application.rb).

config/shoryuken.yml
```diff
@@ -1,4 +1,4 @@
-<% env = ENV['RAILS_ENV'] || 'development' %>
+<% prefix = ENV['ANNOUNCE_RESOURCE_PREFIX'] || ENV['RAILS_ENV'] || 'development' %>
```
cavis, line 1: Per CMS [ca69546](https://github.com/PRX/cms.prx.org/pull/598/commits/ca69546806e8b897c27aa24ad98ce61aecb0940f) ... this can just reference the Rails configuration.

And this PR also needs to update `shoryuken.rb` (I actually already removed the `DEFAULT_JOB_QUEUE_NAME` ENV from Infrastructure).

PR #399, opened 2022-01-18
RFC: multiple feeds schema

cavis:
🚧 WIP

### DIFF
--- a/db/migrate/20220118221746_create_feeds.rb
+++ b/db/migrate/20220118221746_create_feeds.rb
@@ -1,0 +1,42 @@
+class CreateFeeds < ActiveRecord::Migration
+  def change
+    remove_column :podcasts, :max_episodes, :integer
+
+    create_table :feeds do |t|
+      t.references :podcast, index: true, foreign_key: true
+      t.string :slug
+
+      # optional overrides
+      t.text :title
+      t.string :url
+      t.string :path
+      t.string :enclosure_prefix
+      t.string :enclosure_template
+      t.integer :display_episodes_count
+      t.integer :display_full_episodes_count
+
+      # optional sub-feed settings
+      t.boolean :private, default: true
+      t.integer :publish_offset_seconds
+      t.text :filter_zones
+      t.text :filter_tags
+      t.text :audio_format
+
+      t.timestamps null: false
+    end
+
+    add_index :feeds, [:podcast_id, :slug], unique: true
+
+    create_table :feed_tokens do |t|
+      t.references :feed, index: true, foreign_key: true
+
+      t.string :label
+      t.string :token, null: false
+      t.datetime :expires_at
+
+      t.timestamps null: false
+    end
+
+    add_index :feed_tokens, [:feed_id, :token], unique: true
+  end
+end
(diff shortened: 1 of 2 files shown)


db/migrate/20220118221746_create_feeds.rb
```diff
@@ -0,0 +1,43 @@
...
+      t.string :path
+      t.string :enclosure_prefix
+      t.string :enclosure_template
+      t.integer :display_episodes_count
+      t.integer :display_full_episodes_count
+
+      # optional sub-feed settings
+      t.boolean :private, default: true
+      t.integer :publish_offset_seconds
+      t.text :reject_zones
```
cavis, line 22: Sticking with ruby semantics here - a JSON encoded array of zone ids/types you want to _remove_ from these arrangements.  So by default (when null) - you get everything.  But assuming a bit of magical allowed terms:

```ruby
feed.reject_zones = ["ad"] # all type="ad"
feed.reject_zones = ["house"] # just "house"
feed.reject_zones = ["house", "paid"] # equivalent to "ad"
feed.reject_zones = ["pre_"] # zones starting with "pre_"
feed.reject_zones = ["sonic"]
```

This is a bit ambiguous at the moment, since Dovetail arrangements only have 4 zone "types" (`a` ad, `o` original, `i` sonic_id, and `b` billboard).  I'd like to at least add `h` house to that someday.

But the point is - this is a list of some kind of strings, which DTR knows how to interpret.

db/migrate/20220118221746_create_feeds.rb
```diff
@@ -0,0 +1,43 @@
...
+      t.text :filter_tags
```
cavis, line 23: Another JSON encoded list of tags, but these are the episodes that get _included_ in the RSS.  So these are interpreted 100% by Feeder itself.

I called them `tags` thinking we'd want to check both Episode categories and keywords for these.  But I could be convinced otherwise.

db/migrate/20220118221746_create_feeds.rb
```diff
@@ -0,0 +1,43 @@
...
+      t.text :audio_format
```
cavis, line 24: Some hand waving here - but I imagine this will be a serialized Dovetail arrangement "audio format". Which in arrangement json, looks something like:

```json
{
  "version": 2,
  "data": {
    "f": [
      "files"
    ],
    "t": "aooi",
    "a": {
      "f": "mp3",
      "b": 80,
      "c": 1,
      "s": 44100
    }
  }
}
```

db/migrate/20220118221746_create_feeds.rb
```diff
@@ -0,0 +1,43 @@
...
+    end
+
+    add_index :feeds, [:podcast_id, :slug], unique: true
+
+    create_table :feed_tokens do |t|
+      t.references :feed, index: true, foreign_key: true
+
+      t.string :label
+      t.string :token, null: false
+      t.datetime :expires_at
```
cavis, line 36: Maybe overkill - but could be useful to expire a feed token at a specific timestamp?

PR #400, opened 2022-01-19
Cleanup DT3 enclosure templates

cavis:
Cleanup inconsistencies in the `podcasts.enclosure_template` field, ahead of multi-feed work.

- Fix default enclosure templates - the `/_/` in the path is no longer necessary since DTR gets all the ALB traffic.
- We don't set `podcasts.path` anymore - but make sure we always use the podcast_id instead of the path in enclosure urls.  
  - DTR only understands podcast_ids, not aliases, so I've already hardcoded the id into a bunch of enclosure templates that were aliased.

After this deploys, will need to do some manual database cleanup to get as many shows as possible back to the default: 

```
https://dovetail.prxu.org/{slug}/{guid}/{original_filename}
```

This includes:

- [ ] 209 podcasts have the underscore still in there `https://dovetail.prxu.org/_/{slug}/{guid}/{original_filename}`.
- [ ] A ~dozen have their podcast_ids hardcoded to avoid the custom path/alias that DTR can't understand `https://dovetail.prxu.org/_/36/{guid}/{original_filename}`
- [ ] 5 demo/test podcasts have just plain weird enclosures set, and can revert to the default to avoid confusing myself (ids = 1, 10, 12, 13, 21)

I don't love the spotify redownloads and listener_episode churn when enclosures change.  But it might be worth doing that late at night (lowest traffic) to get everything back in line.  Heck, we could consider dropping the `enclosure_template` field entirely at that point.  Except for Memory Palace, which hardcodes all filenames to `/thememorypalace.mp3` for some reason.

### DIFF
--- a/app/models/episode.rb
+++ b/app/models/episode.rb
@@ -269,7 +269,7 @@ class Episode < BaseModel
   end
 
   def podcast_slug
-    podcast.path
+    podcast_id
   end
 
   # used in the API, both read and write
--- a/app/models/podcast.rb
+++ b/app/models/podcast.rb
@@ -61,7 +61,7 @@ class Podcast < BaseModel
   end
 
   def enclosure_template_default
-    "https://#{ENV['DOVETAIL_HOST']}/_/{slug}/{guid}/{original_filename}"
+    "https://#{ENV['DOVETAIL_HOST']}/{slug}/{guid}/{original_filename}"
   end
 
   def publish_updated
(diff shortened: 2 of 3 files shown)


cavis:
Update: didn't fixup the underscores yet.  But got `{slug}` back into all the templates, including memory palace.  So should be good for some multi-feed work on those.

Issue #398, opened 2021-12-16
Add feed model to support multiple feeds


kookster:
* Generate a token to use in the path, essentially the guid for this feed
  * Specify which types of placements are allowed: 
    * Ad free (no paid ads, include house and sonic ids)
    * Original only (no house ads or Sonic IDs either)
  * Include in the API for the podcast, listing the feeds and their urls
    * This has to be a private API call now, so folks can't just discover ad free feeds
  * Specify which episodes appear
    * Max episodes on the feed (does it include the full archive)
    * Delay episodes (put only the latest in some feeds, delay/window release others)
    * Category filter - make a feed for only a single category/keyword
      * Could use this for sub feed like AOI or Book Exploder?
  * Specify audio type in the enclosures
    * Thinking about Apple and providing higher quality audio

PR #401, opened 2022-01-19
RFC multiple feeds

cavis:
For #398.

Sets up the schema/models for multiple feeds.  No functional changes in how feed-rss.xml is written yet.

- [x] Setup new `feeds` and `feed_tokens` tables/models.  Pull some podcasts columns into the new table.
- [x] Delegate from Podcast to the "default" feed (the one with the null `slug`).  This keeps all our APIs backwards compatible for now - setting `podcast.url` actually sets `podcast.default_feed.url`.
- [x] Basic validations.


app/jobs/publish_feed_job.rb
```diff
@@ -12,7 +12,9 @@ def perform(podcast)
     setup_data(podcast)
     @rss = generate_rss_xml
     save_podcast_file(@rss)
-    copy_podcast_file_alias if podcast.feed_rss_alias.present?
+    if podcast.default_feed.file_name != Feed::DEFAULT_FILE_NAME
+      copy_podcast_file_alias
```
cavis, line 16: I nuked the `podcasts.feed_rss_alias` column, but copied it into the default feed's `file_name`.  These are used for feed CDNs that need/want a plain path to the RSS, not a `.xml` extension.

For temporary backwards compatibility - every podcast still writes to `feed-rss.xml`, but these few with aliases also copy to them.  I need to check if any shows need _both_ those files - but may be safe to just write to the one.

app/models/validators/audio_format_validator.rb
```diff
@@ -0,0 +1,42 @@
...
+  BIT_RATES = [96, 112, 128, 160, 192, 224, 256, 320].freeze
+  BIT_DEPTHS = [16, 24, 32].freeze
+  CHANNELS = [1, 2].freeze
+  SAMPLE_RATES = [8000, 11025, 12000, 16000, 22050, 24000, 44100, 48000]
+
+  def validate_each(record, attribute, value)
+    return if value.nil?
+
+    # dovetail style: format, bitrate/bitdepth, channels, sample
+    unless value.is_a?(Hash) && value.keys.sort == %w(b c f s)
```
cavis, line 12: Maybe it's weird to have the literal dovetail arrangement format in here?  But we can always convert it - validating this at least keeps things sane.

app/models/validators/placement_zones_validator.rb
```diff
@@ -0,0 +1,16 @@
+class PlacementZonesValidator < ActiveModel::EachValidator
+  ZONE_TYPES = %w(o a b i)
```
cavis, line 2: Not sure if this is a good or bad idea.  The field is called `reject_zones` ... so presumably we'd reject any zones with these types?  This doesn't give a way to differentiate between house and paid ad zones at the moment though.

Guess we can figure that out later.

app/models/podcast.rb
```diff
@@ -184,4 +189,26 @@ def sanitize_text
...
   end
+
+  # TODO: temporary delegations, until Publish + our Representers get updated
+  # the tests also seem to have issues with this - and the need for just-in-time
+  # initializing the default feed
+  def method_missing(method, *args, &block)
+    if FEED_GETTERS.include?(method)
+      default_feed.try(:public_send, method, *args, &block)
+    elsif FEED_SETTERS.include?(method)
+      self.default_feed ||= feeds.new(private: false)
```
cavis, line 200: The only thing I could figure out to make the tests and API happy -

1) When any of these delegated field's setter is called, just-in-time create the default feed.
2) If nothing calls these setters on create, _still_ create the default feed up in the `set_defaults` that happens on `before_validate`.

Issue #402, opened 2022-01-27
Generate multiple RSS feeds


cavis:
Given the new `feeds` table/model:

- [ ] For each feed, generate an RSS file
- [ ] Default feeds in S3 subfolder at `#{feed.file_name}`
- [ ] Non-default feeds in S3 subfolder at `#{feed.slug}/#{feed.file_name}`
- [ ] Non-default feeds get an enclosure that includes the feed-slug
- [ ] Use the various feed settings (title, episode/full-episode counts, publish offset, keyword filter)
- [ ] May need each feed to have a custom `enclosure_prefix` sooner than i thought (to remove feedburner).  Maybe?

Mstandastote:
HTML simple text format all data

kookster:
multiple feeds released

PR #404, opened 2022-02-07
Multiple feeds support

kookster:
Adds both RSS and API support for podcast multiple feeds
Filters episodes by date, count, and tags.
Also supports different feed audio formats.

Implements #402


app/controllers/podcasts_controller.rb
```diff
@@ -1,6 +1,7 @@
 class PodcastsController < ApplicationController
   def show
     @podcast = Podcast.find(podcast_params)
+    @feed = @podcast.default_feed
```
kookster, line 4: This public endpoint will only continue to support the default feed

app/jobs/publish_feed_job.rb
```diff
@@ -9,58 +9,24 @@ class PublishFeedJob < ApplicationJob
   attr_accessor :podcast, :episodes, :rss, :put_object, :copy_object
 
   def perform(podcast)
-    setup_data(podcast)
```
kookster, line 12: Much of this logic is moved to the feed builder

app/models/episode.rb
```diff
@@ -43,8 +43,9 @@ class Episode < BaseModel
 
   after_save :publish_updated, if: ->(e) { e.published_at_changed? }
 
-  scope :published,
-        -> { where('published_at IS NOT NULL AND published_at <= now()') }
+  scope :published, -> { where('published_at IS NOT NULL AND published_at <= now()') }
+
+  scope :published_by, -> (offset) { where('published_at IS NOT NULL AND published_at <= ?', Time.now + offset) }
```
kookster, line 48: I left the current `published` scope, and added this new one to allow passing in a (positive or negative) offset

app/models/episode.rb
```diff
@@ -138,55 +139,11 @@ def keywords
   end
 
   def media_url
-    media = first_media_resource
```
kookster, line 141: Moved much of this code to the enclosure url builder

app/models/feed.rb
```diff
@@ -3,6 +3,12 @@
 class Feed < BaseModel
   DEFAULT_FILE_NAME = 'feed-rss.xml'.freeze
 
+  AUDIO_MIME_TYPES = {
+    'mp3' => 'audio/mpeg',
+    'flac' => 'audio/flac',
+    'wav' => 'audio/wav'
+  }.freeze
```
kookster, line 10: This was quickly added for the audio formats we support, as it doesn't need to be a general solution for all mime types

app/models/feed.rb
```diff
@@ -47,10 +53,50 @@ def default_runtime_settings?
...
   def published_url
-    if default?
-      "#{podcast.base_published_url}/#{file_name}"
-    else
-      "#{podcast.base_published_url}/#{slug}/#{file_name}"
+    "#{podcast.base_published_url}/#{published_path}"
+  end
+
+  def published_path
+    default? ? file_name : "#{slug}/#{file_name}"
```
kookster, line 60: Factored this path value out to use in both the urls and the writing to s3

app/models/feed.rb
```diff
@@ -47,10 +53,50 @@ def default_runtime_settings?
...
+  end
+
+  def feed_episodes
+    include_in_feed = []
+    feed_max = display_episodes_count.to_i
+    filtered_episodes.each do |ep|
+      next unless include_episode_categories?(ep)
```
kookster, line 67: The categories are stored as a json string attribute, so this filtering is happening in the Ruby code rather than the database query

app/models/podcast.rb
```diff
@@ -55,20 +55,20 @@ def self.by_prx_series(series)
     Podcast.find_by(prx_uri: series_uri)
   end
 
+  def self.enclosure_template_default
+    "https://#{ENV['DOVETAIL_HOST']}{/podcast_id,feed_slug,guid,original_basename}{feed_extension}"
```
kookster, line 59: Note the addition of a `feed_extension`, to match when the audio for this feed is different

app/models/podcast.rb
```diff
@@ -129,13 +129,7 @@ def managing_editor
...
 
   def feed_episodes
-    feed = []
-    feed_max = display_episodes_count.to_i
-    episodes.published.each do |ep|
-      feed << ep if ep.include_in_feed?
-      break if (feed_max > 0) && (feed.size >= feed_max)
-    end
-    feed
+    default_feed.feed_episodes
```
kookster, line 132: Makes more sense to have this at the feed level - delegating to the feed for compatibility

app/views/podcasts/show.rss.builder
```diff
@@ -24,20 +24,20 @@ xml.rss 'xmlns:atom' => 'http://www.w3.org/2005/Atom',
...
+    xml.atom :link, href: (@feed.url || @feed.published_url), rel: 'self', type: 'application/rss+xml'
 
-    unless @podcast.new_feed_url.blank?
-      xml.itunes :'new-feed-url', @podcast.new_feed_url
+    unless @feed.new_feed_url.blank?
+      xml.itunes :'new-feed-url', @feed.new_feed_url
     end
 
-    xml.itunes :block, 'Yes' if @podcast.itunes_block
+    xml.itunes :block, 'Yes' if @podcast.itunes_block || @feed.try(:private)
```
kookster, line 40: If the feed is private, add `itunes:block`

app/models/podcast.rb
```diff
@@ -1,6 +1,6 @@
 class Podcast < BaseModel
-  FEED_GETTERS = %i(url new_feed_url display_episodes_count display_full_episodes_count)
-  FEED_SETTERS = %i(url= new_feed_url= display_episodes_count= display_full_episodes_count=)
+  FEED_GETTERS = %i(url new_feed_url display_episodes_count display_full_episodes_count enclosure_prefix)
+  FEED_SETTERS = %i(url= new_feed_url= display_episodes_count= display_full_episodes_count= enclosure_prefix)
```
cavis, line 3: Doesn't this need to end with a `=`?

app/views/podcasts/show.rss.builder
cavis: Kind of makes me wonder if we should drop the `itunes_block` column someday.  There does seem to be some cleanup to do, once Publish is using the Feeds-API for a bunch of these columns.

cavis (commented):
It begins with the questions!

General strategy looks good, but I had some specific questions.  Esp. around `enclosure_template` and `enclosure_prefix` fields, which now seem to exist on multiple models.

app/models/enclosure_url_builder.rb
```diff
@@ -0,0 +1,75 @@
...
+      original_filename: orig_fn,
+      original_extension: orig_ex,
+      original_basename: File.basename(orig_fn, orig_ex),
+      filename: fn,
+      extension: ex,
+      basename: File.basename(fn, ex),
+      podcast_id: podcast.try(:id),
+      guid: episode.guid,
+      slug: podcast.try(:id),
+      feed_slug: feed.try(:slug),
```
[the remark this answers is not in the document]
kookster: they are, but some terrible person might pass in a `nil` for them

app/models/enclosure_url_builder.rb
```diff
@@ -0,0 +1,75 @@
...
+      feed_extension: feed_extension(feed) || ex
+    }.merge(original).merge(base)
+  end
+
+  def feed_extension(feed)
+    format = (feed.try(:audio_format) || {})[:f]
```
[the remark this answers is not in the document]
kookster: it shouldn't happen, but it could happen that a podcast has no default feed, in which case a `nil` will be passed in for it, and this would all start throwing errors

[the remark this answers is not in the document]
kookster: I believe I saw that happening from some of the tests, and so coded these defensively

app/models/enclosure_url_builder.rb
```diff
@@ -0,0 +1,75 @@
...
+  def enclosure_prefix_url(u, prefix)
+    return u if prefix.blank?
+    pre = Addressable::URI.parse(prefix)
+    orig = Addressable::URI.parse(u)
+    orig.path = File.join(pre.path, orig.host, orig.path)
+    orig.scheme = pre.scheme
+    orig.host = pre.host
+    orig.to_s
+  end
+end
```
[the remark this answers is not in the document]
kookster: fixed

app/models/episode.rb
```diff
@@ -138,55 +139,11 @@ def keywords
...
-    orig.to_s
-  end
-
-  def enclosure_template_url(base_url, original_url = nil)
-    return base_url if enclosure_template.blank?
-
-    expansions = enclosure_template_expansions(base_url, original_url)
-    template = Addressable::Template.new(enclosure_template)
-    template.expand(expansions).to_str
+    first_media_resource.try(:href)
```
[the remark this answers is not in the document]
kookster: yes, now `media_url` returns the url for the media, and `enclosure_url` will return the enclosure - it's better

app/models/feed.rb
```diff
@@ -47,10 +53,46 @@ def default_runtime_settings?
...
+  def normalize_category(cat)
+    cat.to_s.downcase.gsub(/[^ a-z0-9_-]/, '').gsub(/\s+/, ' ').strip
+  end
+
+  def filtered_episodes
+    podcast.episodes.published_by(episode_offset_seconds.to_i)
+  end
+
+  def enclosure_template
+    Podcast.enclosure_template_default
```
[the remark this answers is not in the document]
kookster: yeah, I see your point, I guess the cleanest thing would be to add this to the feed model like the prefix

app/models/podcast.rb
kookster: oopS!

app/views/podcasts/show.rss.builder
kookster: seems like another "maybe this should be feed specific?" attribute

db/migrate/20220207234307_add_feed_enclosure_prefix.rb
```diff
@@ -0,0 +1,21 @@
+class AddFeedEnclosurePrefix < ActiveRecord::Migration
+  def up
+    add_column :feeds, :enclosure_prefix, :string
+
+    Podcast.with_deleted.each do |podcast|
+      next unless feed = podcast.default_feed
+      feed.enclosure_prefix = podcast.enclosure_prefix
```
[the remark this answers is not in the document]
kookster: I can delete it

db/schema.rb
```diff
@@ -11,11 +11,10 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20220118221746) do
+ActiveRecord::Schema.define(version: 20220207234307) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
-  enable_extension "uuid-ossp"
```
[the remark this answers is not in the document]
kookster: it got removed by running migrations, maybe a db version difference?

docker-compose-ci.yml
```diff
@@ -1,21 +1,33 @@
...
+      VIRTUAL_HOST: 'feeder.prx.docker'
+      LOCAL_ENV: 'true'
+      DB_ENV_POSTGRES_USER: 'feeder'
+      DB_ENV_POSTGRES_PASSWORD: 'password'
+      DB_ENV_POSTGRES_DATABASE: 'feeder'
+      DB_PORT_5432_TCP_ADDR: 'db'
+      DB_PORT_5432_TCP_PORT: 5432
+      AWS_REGION: us-east-1
+  db:
+    image: postgres:9.4-alpine
```
cavis, line 23: Why 9.4?  Looks like our prod Aurora postgres is 13.3.

app/models/feed.rb
```diff
@@ -47,10 +58,46 @@ def default_runtime_settings?
...
+  def normalize_category(cat)
+    cat.to_s.downcase.gsub(/[^ a-z0-9_-]/, '').gsub(/\s+/, ' ').strip
+  end
+
+  def filtered_episodes
+    podcast.episodes.published_by(episode_offset_seconds.to_i)
+  end
+
+  def enclosure_template
+    self[:enclosure_template] || Feed.enclosure_template_default
```
cavis, line 96: Okay, so if I'm reading this all right ...

In the migration, you copy the podcast enclosure-template over too the feed(s).  So these will all be non-nil for default_feeds when this deploys.

But any newly added Feeds (and we should fixup the TTD adfree demo feed) will get the new enclosure template, with the additional fanciness.

cavis (approved):
One question about Postgres versions, but I think this looks good!

Would be nice to deploy a _bit_ carefully.  Maybe do a diff on an RSS file before/after deploying?

cavis (approved):
2nd +1

PR #406, opened 2022-02-22
Handle video enclosures in feeds differently

kookster:
If the episode has video enclosures, use that type, not the feed.


cavis (approved):
👍

PR #408, opened 2022-03-16
More feeds APIs

cavis:
Couple more Podcast-Feed APIs needed to build out the Publish pages.

- Include a link on the feed, for the _private_ feed proxy `p.prxu.org`.  Will need to add those ENVs to Feeder.
- Simple attribute getter/setter for feed tokens.  (Will be easier than attempting a nested resource for them).

### DIFF
--- a/app/models/feed.rb
+++ b/app/models/feed.rb
@@ -58,7 +58,11 @@ class Feed < BaseModel
   end
 
   def published_url
-    "#{podcast.base_published_url}/#{published_path}"
+    if private?
+      "#{podcast.base_private_url}/#{published_path}{?auth}"
+    else
+      "#{podcast.base_published_url}/#{published_path}"
+    end
   end
 
   def published_path
--- a/app/models/feed.rb
+++ b/app/models/feed.rb
@@ -16,7 +16,7 @@ class Feed < BaseModel
   serialize :audio_format, HashSerializer
 
   belongs_to :podcast, -> { with_deleted }
-  has_many :feed_tokens, dependent: :destroy
+  has_many :feed_tokens, autosave: true, dependent: :destroy
   alias_attribute :tokens, :feed_tokens
 
   validates :slug, allow_nil: true, uniqueness: { scope: :podcast_id, allow_nil: false }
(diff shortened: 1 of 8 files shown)


app/models/feed.rb
```diff
@@ -41,6 +41,38 @@ def set_defaults
...
+    feed_tokens.map do |t|
+      {
+        label: t.label,
+        token: t.token,
+        expires: t.expires_at
+      }.delete_if { |k, v| v.nil? }
+    end
+  end
+
+  def represented_tokens=(updates)
```
cavis, line 54: As I discovered in Augury - using a nested-collection setter like this doesn't play well with Representable/Roar/HalApi stuff.

So the easiest path is to bypass that entirely, and just pretend this list-of-hashes is a simple property on the Feed.

app/models/feed.rb
```diff
@@ -58,7 +90,11 @@ def default_runtime_settings?
   end
 
   def published_url
-    "#{podcast.base_published_url}/#{published_path}"
+    if private?
+      "#{podcast.base_private_url}/#{published_path}{?auth}"
+    else
+      "#{podcast.base_published_url}/#{published_path}"
+    end
```
cavis, line 97: As far as I could tell, this `feed.published_url` wasn't used anywhere yet.  So should be an issue to vary it based on public vs private feeds.

app/representers/api/auth/feed_representer.rb
```diff
@@ -17,11 +17,21 @@ class Api::Auth::FeedRepresenter < Api::BaseRepresenter
...
   link :podcast do
     api_authorization_podcast_path(represented.podcast) if represented.id && represented.podcast
   end
+
+  link :rss do
+    {
+      href: represented.published_url,
+      templated: true
+    }
+  end
```
cavis, line 36: I'm a bit lukewarm on this one.

The Podcast representer has a non-writeable _property_ called `published_url`.  Which is the default feed's `published_url`.  Not sure why that isn't a `_link`?

But here, especially because the value could be templated (private feeds need to include the `?auth=1234` token) - I made it a link.  Bad idea?  Maybe.

I'm also lukewarm on what to call the link.  Went with `prx:rss` here.

kookster: The Podcast representer having a url as a read only property instead of a link is an error.

I would compare this with how the episode representer has a feed _link, and an enclosure _link, which was a better way to do it, and that's consistent with this addition of a private-feed _link

in other words, this looks right to me, and Podcast representer published_url looks wrong

kookster (approved):
This looks good - I do wonder if roar/representable could handle the "lonely" hash collection of the feed tokens, but it's probably not worth it to rewrite this to make it work

app/models/feed.rb
cavis: I think the issue I had in Augury was related to our HalApiRails overrides.  I had some [notes here](https://github.com/PRX/augury.prx.org/blob/master/app/representers/api/flight_representer.rb#L22) ... and figured I'd just avoid the issue entirely here.

Yeah, if we do figure that out, at least the tests will stay the same!

Issue #426, opened 2022-07-07
Ensure has_one images have been processed


cavis:
Right now our images are has_many relations, but we do some `replace_resources!` logic to wait for them to process before deleting the old and using the new.

So on a Podcast, we have something like

```ruby
has_one :itunes_image, autosave: true, dependent: :destroy
```

For safety, we probably should provide a sort order instead of relying on the default postgres sort order.  _And_ it should also filter the images based on status, so we don't accidentally use an unprocessed image in the RSS.

PR #427, opened 2022-07-12
Move podcast images to feed

### LINKED ISSUE #422: Override podcast artwork per-feed
Getting into "season specific" Feeds, we're going to need the ability to override the podcast artwork.

I'm not exactly sure how this could/should work in the short-term.  Maybe we use the "purpose" field from CMS to indicate which SeriesImage to use?  Or we just make a Feeder schema change to allow directly uploading images _just_ to Feeder (not CMS)?

cavis:
Fixes #422, and untangles some other deprecated/inconsistent code.  Will attempt to comment this pretty thoroughly, but the main changes are:

1. Drop the deprecated `feed_images` and `itunes_images` tables.  The data was all moved to the STI `podcast_images` table by a migration back on 2016-02-16.
2. Recreate _new_ `feed_images` and `itunes_images` _non-STI_ tables.  And instead of belonging to a `podcast_id`, these now belong to a `feed_id`.
3. Migrate podcast images to their default feeds
4. Update/proxy all the podcast image getters/setters so most things just keep working.  (When you access the podcast.feed_image ... you get the podcast.default_feed.feed_image).
5. Fixup inconsistencies in our API (`media` vs images properties).
6. Fixup edge cases in processing images, so we always keep/show the old image until the new completes.
7. API for getting/setting feed images.


app/models/concerns/feed_image_file.rb
```diff
@@ -0,0 +1,22 @@
...
+  def destination_path
+    "#{feed.podcast.path}/#{feed_image_path}"
+  end
+
+  def published_url
+    "#{feed.podcast.base_published_url}/#{feed_image_path}"
+  end
+
+  def feed_image_path
+    "images/#{guid}/#{file_name}"
```
cavis, line 20: NOTE: I opted to keep _all_ feed/itunes images in the current path scheme: `s3://prx-feed/<podcast>/images/<image_guid>/<filename>`.  So the feed slug never ends up in here, and our random image guids should never collide.  This keeps the `default_feed.feed_image` backwards compatible with the path of the previous `podcast.feed_image`.

app/models/concerns/image_file.rb
```diff
@@ -19,6 +19,21 @@ module ImageFile
     enum status: [ :started, :created, :processing, :complete, :error, :retrying, :cancelled ]
   end
 
+  class_methods do
+    def build(file)
```
cavis, line 23: Helper to take in either a string url, a hash, or a model - normalize that and return either a `WhatevImage` model or nil if no original_url was set.

Used by the Feed/Episode setter methods, to find-or-build images.

app/models/concerns/image_file.rb
```diff
@@ -48,6 +67,17 @@ def url
     complete? ? self[:url] : self[:original_url]
   end
 
+  def href
+    complete? ? url : original_url
+  end
+
+  def href=(h)
```
cavis, line 74: I started switching the 3 image APIs to match the Episode `media` API.  Where the main property an API consumer cares about is `href` - you PUT it to change an image url.  And once processing completes, it becomes the `f.prxu.org` CDN url of the copied image.

app/models/concerns/image_file.rb
```diff
@@ -61,6 +91,7 @@ def reset_image_attributes
     self.height = nil
     self.width = nil
     self.size = nil
+    self.status = :created
```
cavis, line 94: This seemed like an oversight - but we probably didn't notice, since you normally run `copy_media` right away after changing the original_url.

app/models/episode.rb
```diff
@@ -101,6 +101,17 @@ def image
...
   end
 
+  # API updates for image=
+  def image_file; images.first; end
+  def image_file=(file)
+    img = EpisodeImage.build(file)
+    if img && img.original_url != image_file.try(:original_url)
+      images << img
+    elsif !img
+      images.destroy_all
```
cavis, line 111: You'll see similar methods for itunes and feed images.  These do similar logic as episode `media_files`, and the same logic the CMS handlers did:

- Check if we already have an image with the same original url - if so, skip
- If not, create a new image but leave the old one there.  The new image only replaces the old after the Porter callbacks complete.
- If null, immediately delete all images.

app/models/feed.rb
```diff
@@ -19,6 +19,11 @@ class Feed < BaseModel
   has_many :feed_tokens, autosave: true, dependent: :destroy
   alias_attribute :tokens, :feed_tokens
 
+  has_one :feed_image, -> { complete.order('created_at DESC') }, autosave: true, dependent: :destroy
```
cavis, line 22: Fixes #426.  This `feed_image` will return the _most recent completed_ image.  So any new images will end up in the has_many, but not here until processing completes.

app/models/feed.rb
```diff
@@ -104,4 +109,31 @@ def mime_type
...
     AUDIO_MIME_TYPES[f]
   end
+
+  def copy_media(force = false)
+    feed_images.each { |i| i.copy_media(force) }
+    itunes_images.each { |i| i.copy_media(force) }
+  end
+
+  # API updates for feed_image=
+  def feed_image_file; feed_images.first; end
```
cavis, line 119: Note: this API-getter returns the most recent image _of any status_.  So it's different than the `has_one :feed_image`.  Reasoning being - if you're working in the API, you expect the `{"feedImage":{"href":"/new/href"}}` you just PUT to show up immediately.  But we still want the RSS to contain the old image until processing completes.

app/representers/api/image_representer.rb
```diff
@@ -2,13 +2,16 @@
 
 class Api::ImageRepresenter < Roar::Decorator
   include Roar::JSON
+  include HalApi::Representer::FormatKeys
 
-  property :url
+  property :href
```
cavis, line 7: I want to remove `url` from this API entirely in favor of `href` ... but will have to update the Castle code that scrapes Feeder first.

app/representers/api/image_representer.rb
```diff
@@ -2,13 +2,16 @@
...
+  property :url, writeable: false # TODO: deprecate in favor of href ... but Castle scrapes this field
   property :original_url, writeable: false
```
cavis, line 9: Note: since I added `HalApi::Representer::FormatKeys` - this property now ends up camelcased `originalUrl`.

app/representers/api/media_resource_representer.rb
```diff
@@ -2,6 +2,7 @@
 
 class Api::MediaResourceRepresenter < Roar::Decorator
   include Roar::JSON
+  include HalApi::Representer::FormatKeys
 
   property :href
   property :original_url, writeable: false
```
cavis, line 8: Also here - this is now `originalUrl`.

app/views/podcasts/show.rss.builder
```diff
@@ -23,13 +23,13 @@ xml.rss 'xmlns:atom' => 'http://www.w3.org/2005/Atom',
...
     xml.image do
-      xml.url @podcast.feed_image.url
+      xml.url @feed_image.url
       xml.title @feed.title || @podcast.title
-      xml.link @podcast.link
-      xml.width @podcast.feed_image.width
-      xml.height @podcast.feed_image.height
-      xml.description @podcast.feed_image.description unless @podcast.feed_image.description.blank?
-    end if @podcast.feed_image
+      xml.link @feed_image.link || @podcast.link
```
cavis, line 28: Seemed like an oversight - we should use the FeedImage link if present.

svevang (approved):
Looks good!

PR #429, opened 2022-07-26
Added webmonetization/payment pointer

a-maci29:


.gitignore
```diff
@@ -25,3 +25,5 @@ notes*
 
 /public/assets
 /build
+
+.DS_Store
```
cavis, line 29: Missing a newline at the end of the file?  (May need to add that config to your editor).

.tool-versions
```diff
@@ -0,0 +1 @@
+ruby 2.3.8
```
cavis, line 1: This project already has a `.ruby-version` (which was also upgraded to 2.5.9 since you opened this PR).

It might be easiest to just remove this file again, and follow the steps in the [asdf-ruby readme](https://github.com/asdf-vm/asdf-ruby#migrating-from-another-ruby-version-manager) to make it recognize `.ruby-version` files?

app/views/podcasts/show.rss.builder
```diff
@@ -92,6 +93,17 @@ xml.rss 'xmlns:atom' => 'http://www.w3.org/2005/Atom',
     xml.sy :updateFrequency, @podcast.update_frequency if @podcast.update_frequency
     xml.sy :updateBase, @podcast.update_base if @podcast.update_base
 
+    payment_pointer = @feed.payment_pointer || @podcast.payment_pointer
+    unless payment_pointer.blank?
```
cavis, line 97: Generally, we want to check the _affirmative_ condition rather than the negative one.  So it would be slightly easier to read to change this line to:

```ruby
if payment_pointer.present?
```

db/migrate/20220823044927_add_payment_pointer.rb
```diff
@@ -0,0 +1,6 @@
+class AddPaymentPointer < ActiveRecord::Migration
+  def change
+    add_column :podcasts, :payment_pointer, :string
```
cavis, line 3: This has changed a bit since this PR was opened, but I believe the new pattern in Feeder is:

We _only_ add the column to the `:feeds` table, unless there's a compelling reason to also add it to `:podcasts`.  And then in the show rss builder do something like:

```ruby
payment_pointer = @feed.payment_pointer || @podcast.default_feed&.payment_pointer
```

cavis (changes requested):
The pieces are coming together!  And I appreciate the tests.

Couple of minor nitpicks, but mostly: i think the pattern for adding fields has changed since this is written.  Should only add the field to the `:feeds` table.

.tool-versions
cavis, line 1: We probably don't want both `.tool-versions` and `.ruby-version` in a single project.  And in any case... the `master` branch was upgraded to 2.5.9 since this PR was opened.  (And we don't want to downgrade prod).

It might be easiest to just remove this file again, and follow the steps in the [asdf-ruby readme](https://github.com/asdf-vm/asdf-ruby#migrating-from-another-ruby-version-manager) to make it recognize `.ruby-version` files?

cavis (changes requested):
I think it's just the ruby-version now, otherwise looking :+15:.

.tool-versions
a-maci29: Committed with the `.tool-versions` removed.

kookster (approved):
lgtm!

PR #433, opened 2022-08-01
Set up the exclude / include state

svevang:
In preparation for exclusive content, we can mark a set of episode tags as being excluded on the feed.


app/models/feed.rb
```diff
@@ -81,17 +82,14 @@ def feed_episodes
     include_in_feed = []
     feed_max = display_episodes_count.to_i
     filtered_episodes.each do |ep|
-      next unless include_episode_categories?(ep)
```
cavis, line 84: +1 ... makes sense to move these checks into `filtered_episodes`.

And I was surprised to see that `Episode.include_in_feed?` is making some completely unrelated checks ... if the media is ready.  I'm guessing the name of that method predates the `Feed` model.

app/models/feed.rb
```diff
@@ -100,8 +98,29 @@ def normalize_category(cat)
...
+  def use_include_tags?
+    !include_tags.nil?
+  end
+
+  def use_exclude_tags?
+    !exclude_tags.nil?
+  end
+
   def filtered_episodes
-    podcast.episodes.published_by(episode_offset_seconds.to_i)
```
cavis, line 104: Did we lose the `published_by` check?

cavis (changes requested):
Some minor regressions on how we handle episode-offset and empty arrays.  Otherwise I think this looks good.  And good to have some tests on this (which we were previously missing).

Issue #436, opened 2022-08-16
Move teaser/description to Feed


cavis:
Teaser (`podcast.subtitle`) and Description (`podcast.description`) need to overridable at the Feed level.  For when you're doing season specific or other keyword feeds.

- [ ] Move fields to the default feed
- [ ] Aliases for podcast and non-default feeds
- [ ] Add API endpoints on Feed
- [ ] Update Publish

TODO: what about the optional `podcast.summary` field?  (It defaults to a sanitize-links-only version of the description).

Why do we even allow varying those anyways?

PR #440, opened 2022-08-18
Move subtitle/description/summary to feed

cavis:
Backend work for #436.

Moves 3 fields from `Podcast` to `Feed`.  But keeps them proxied to the `podcast.default_feed` so existing APIs continue working.

### DIFF
--- a/app/helpers/podcasts_helper.rb
+++ b/app/helpers/podcasts_helper.rb
@@ -19,10 +19,14 @@ module PodcastsHelper
   end
 
   def show_itunes_summary?(model)
-    model.summary || model.description
+    model.summary.present? || model.description.present?
   end
 
   def itunes_summary(model)
-    model.summary || sanitize_links_only(model.description)
+    if model.summary.present?
+      model.summary
+    else
+      sanitize_links_only(model.description || '')
+    end
   end
 end
--- a/app/models/podcast.rb
+++ b/app/models/podcast.rb
@@ -1,6 +1,6 @@
 class Podcast < BaseModel
-  FEED_GETTERS = %i(url new_feed_url display_episodes_count display_full_episodes_count enclosure_prefix enclosure_template feed_image itunes_image)
-  FEED_SETTERS = %i(url= new_feed_url= display_episodes_count= display_full_episodes_count= enclosure_prefix= enclosure_template= feed_image= itunes_image=)
+  FEED_GETTERS = %i(subtitle description summary url new_feed_url display_episodes_count display_full_episodes_count enclosure_prefix enclosure_template feed_image itunes_image)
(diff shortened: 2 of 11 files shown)


app/helpers/podcasts_helper.rb
```diff
@@ -19,10 +19,14 @@ def first_nonblank(type, items)
   end
 
   def show_itunes_summary?(model)
-    model.summary || model.description
+    model.summary.present? || model.description.present?
```
cavis, line 22: It seemed like an oversight not to check for empty strings `""` in all these summary/description checkers?  So added them both here (in the RSS generation) and in the API.

app/models/podcast.rb
```diff
@@ -1,6 +1,6 @@
 class Podcast < BaseModel
-  FEED_GETTERS = %i(url new_feed_url display_episodes_count display_full_episodes_count enclosure_prefix enclosure_template feed_image itunes_image)
-  FEED_SETTERS = %i(url= new_feed_url= display_episodes_count= display_full_episodes_count= enclosure_prefix= enclosure_template= feed_image= itunes_image=)
+  FEED_GETTERS = %i(subtitle description summary url new_feed_url display_episodes_count display_full_episodes_count enclosure_prefix enclosure_template feed_image itunes_image)
+  FEED_SETTERS = %i(subtitle= description= summary= url= new_feed_url= display_episodes_count= display_full_episodes_count= enclosure_prefix= enclosure_template= feed_image= itunes_image=)
```
cavis, line 3: Fields are proxied to the podcast's default feed - so Publish can continue working with the Podcast API after this deploys.

app/representers/api/auth/feed_representer.rb
```diff
@@ -7,6 +7,10 @@ class Api::Auth::FeedRepresenter < Api::BaseRepresenter
   property :file_name
   property :private
   property :title
+  property :subtitle
+  property :description
+  property :summary
+  property :summary_preview, exec_context: :decorator, if: ->(_o) { represented.summary.blank? }
```
cavis, line 13: This seems to be a Publish display thing?  If you have a blank/nil `summary`, it displays what the sanitized-links-only version of the `description` is.

app/helpers/podcasts_helper.rb
kookster: great!

app/models/podcast.rb
kookster: makes the podcast model pretty anemic, but maybe that's a good thing!

app/representers/api/auth/feed_representer.rb
kookster: yeah, as I recall, that's how the summary shows up in iTunes/Apple Podcasts, which supports links, and not much else.

kookster (approved):
lgtm!

PR #448, opened 2022-09-28
Revised Payment Pointer

a-maci29:
[] Removed existing `payment_pointer` from column
[] Added new fields to `podcasts` and `feeds` tables with new parameters for payment options


cavis (approved):
👍

PR #460, opened 2022-10-18
validates original guid uniqueness

### LINKED ISSUE #444: Validate unique original_guids
There's currently a unique postgres index `index_episodes_on_original_guid_and_podcast_id`.  With the conditions:

```
((deleted_at IS NULL) AND (original_guid IS NOT NULL))
```

But there's no activerecord validation, which causes 500s sometimes.  Add the model validation, using the same criteria.

a-maci29:
Resolves #444 

Validates original guids


cavis (approved):
👍 looks good to me!

PR #465, opened 2022-11-02
Feat/apple api minimal error handling

svevang:


app/models/apple/episode.rb
```diff
@@ -9,33 +9,23 @@ class Episode
...
-      episode_bridge_results
+      api.bridge_remote_and_retry!("getEpisodes", episodes.map(&:get_episode_bridge_params))
     end
 
     def self.create_episodes(api, episodes)
       return if episodes.empty?
 
-      resp =
-        api.bridge_remote("createEpisodes", episodes.map(&:create_episode_bridge_params))
+      episode_bridge_results = api.bridge_remote_and_retry!("createEpisodes",
```
cavis, line 18: Looks like these calls could now throw errors (if the remote call fails enough retries).

Will these eventually bubble up to a worker/job then?  Or does this create `apple_podcast_deliveries` rows for failures?

/me reads on!

cavis (approved):
Couple of questions and minor nitpicks, but nothing that needs to hold this up.

👍 !

app/models/apple/api.rb
```diff
@@ -6,22 +6,24 @@
...
+      new(provider_id: ENV["APPLE_PROVIDER_ID"],
+          key_id: ENV["APPLE_KEY_ID"],
+          key: apple_key_pem)
     end
 
-    def initialize(provider_id, key_id, key)
-      @provider_id = provider_id
-      @key_id = key_id
-      @key = key
+    def initialize(**attributes)
```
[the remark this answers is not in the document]
cavis: Neat!

app/models/apple/episode.rb
cavis: Makes good sense.  I'd rather see the errors, when initially doing this all manually.

PR #467, opened 2022-11-10
Feat/guard for existing media

### LINKED ISSUE #452: Guard for existing media via podcast connect.
Guard for existing media:

- [x] Set apple show id for existing shows
- [x] Pull down podcast containers, create local podcast containers, and (potentially) delivery file records.
- [x] Add guards to indicate that syncing is not necessary if the episodes have files attached.

svevang:
Closes #452

This PR sets up
- [x] A way to attach an existing show to a local Apple::Show state
- [x]  a new 'update local state' pattern, in the case where there are remote resources that are not recorded locally

The behavior added will sidestep uploads for episodes that have existing containers + deliveries


app/models/apple/podcast_container.rb
```diff
@@ -29,76 +27,103 @@ def self.update_podcast_container_file_metadata(api, episodes_to_sync)
...
         end
 
       SyncLog.create!(feeder_id: pc.id, feeder_type: :podcast_containers, external_id: external_id)
 
       pc
     end
 
-    def self.get_podcast_containers(api, episodes_to_sync)
-      api.bridge_remote_and_retry!("getPodcastContainers", get_podcast_containers_bridge_params(api, episodes_to_sync))
+    def self.get_podcast_containers_via_episodes(api, episodes)
```
svevang, line 77: Following the hypermedia links here (and in the case of delivery / delivery_files).

Get the remote container resource _nested under the episode_. Flatten and follow those links to the canonical resource, the base podcast container resource in this case. Then update the local state.

This adopts the pattern, that if we're storing remote state locally, always store the non-enumerated "base" resource instead of the nested resource.  e.g. store `podcastContainers/1234` instead of `podcastContainers?filter[vendorId]=<some episode id>`

app/models/apple/podcast_delivery.rb
```diff
@@ -2,74 +2,151 @@
...
       failed: "FAILED"
     }
 
-    def self.create_podcast_deliveries(api, episodes_to_sync)
-      episodes_needing_delivery = episodes_to_sync.reject { |ep| ep.podcast_container.podcast_delivery.present? }
+    def self.update_podcast_deliveries_state(api, episodes)
+      results = get_podcast_deliveries_via_containers(api, episodes.map(&:podcast_container))
+
+      join_on_apple_episode_id(episodes, results).map do |(episode, delivery_row)|
+        upsert_podcast_delivery(episode, delivery_row)
```
svevang, line 25: Same pattern as podcast_containers:

1) get resources via hypermedia graph
2) join results to the episodes here and upsert local state

app/models/apple/api.rb
```diff
@@ -159,18 +159,22 @@ def unwrap_bridge_response(resp)
       [oks + fixed_errs, remaining_errors]
     end
 
+    def api_bridge_url
+      URI.parse(ENV.fetch("APPLE_API_BRIDGE_URL"))
+    end
+
     def bridge_remote(bridge_resource, bridge_options)
       # TODO: pull this in via the ENV
```
cavis, line 167: TODO done?

app/models/apple/api.rb
```diff
@@ -204,7 +208,7 @@ def make_bridge_request(uri, body)
       # TODO: vary this with the bridge endpoint url
       use_ssl = false
 
-      Net::HTTP.start(uri.hostname, uri.port, use_ssl: use_ssl) do |http|
+      Net::HTTP.start(uri.hostname, uri.port, use_ssl: use_ssl, read_timeout: 10.minutes) do |http|
```
cavis, line 211: That's a long timeout!  Could this potentially get stuck for some reason, and tie up the feeder worker(s)?

app/models/apple/episode.rb
```diff
@@ -44,10 +44,11 @@ def self.insert_sync_logs(episodes, results)
...
 
-    def initialize(show, episode)
-      @show = show
-      @feeder_episode = episode
-      @api = Apple::Api.from_env
+    def initialize(**kwargs)
+      kwargs = kwargs.with_indifferent_access
+      @show = kwargs["show"]
+      @feeder_episode = kwargs["feeder_episode"]
+      @api = kwargs["api"] || Apple::Api.from_env
```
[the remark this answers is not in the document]
cavis: Good idea!

app/models/apple/publisher.rb
```diff
@@ -65,39 +55,66 @@ def publish!
...
       # Apple RSS scaping means we don't need containers for freemium episode images
       # But we do need asset containers for apple-only (non-rss) images
-      Apple::PodcastContainer.create_podcast_containers(api, episodes_to_sync, show)
-      Apple::PodcastContainer.update_podcast_container_file_metadata(api, episodes_to_sync)
-    end
 
-    def get_podcast_deliveries
-      # TODO: right now we only create one container,
-      Apple::PodcastDelivery.get_podcast_deliveries(api, episodes_to_sync)
+      Rails.logger.info("Starting podcast container sync")
```
cavis, line 88: Logging++.  Especially up here where things are being orchestrated.

cavis (approved):
👍 looks good to me!  Hypermedia!

app/models/apple/api.rb
svevang: Right now there is no worker/job integration generally, or publishing pipeline specifically.  But afaik the worker timeout would take precedence, right?

cavis: Ah true.  Forgot about that worker timeout.

Issue #457, opened 2022-10-17
Multi-tenant Apple API credentials


svevang:
Set up a new multi-tenant approach to storing and using Apple credentials. This is modeling the case where the user retains control over their Podcasts Connect account and data, and authorizes uploading and publishing use via a set of credentials.

Currently there is a ENV based approach to loading the primary PRX testing apple keys. Extend this to handle user keys:

- [ ] Process key import format into usable keys (helper methods)
- [x] New columns ~on the Podcast model~ on a new AppleCredential model tying 2 Feeds together to hold these credentials
- [ ] Fallback to ENV based keys in the case where PRX is managing the show
- [ ] Set up parameter store to hold staging&test versus prod keys (in the case of ENV keys)

svevang:
Account ids exist in feeder, could store the apple key file keyed on that account id, and then joining that to the podcasts/feeds.

svevang:
Currently, the `Apple::Api.from_env` method is what builds the api instances:
https://github.com/PRX/feeder.prx.org/blob/feat/apple-api-integration/app/models/apple/api.rb#L13

This is called from a few places. Maybe it could be `Apple::Api.from_podcast_or_env(podcast)`

PR #468, opened 2022-11-30
Apple credentials schema

cavis:
Implements schema changes for #457

### DIFF
--- a/app/models/apple_credential.rb
+++ b/app/models/apple_credential.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-class AppleCredential < ActiveRecord::Base
+class AppleCredential < BaseModel
   belongs_to :public_feed, class_name: "Feed"
   belongs_to :private_feed, class_name: "Feed"
 
@@ -8,9 +8,14 @@ class AppleCredential < ActiveRecord::Base
   validates_presence_of :private_feed
   validates_associated :public_feed
   validates_associated :private_feed
-  validates_presence_of :apple_key_id
-  validates_presence_of :apple_key_pem_b64
+  validates_presence_of :apple_provider_id, if: :any_apple_credentials_exist?
+  validates_presence_of :apple_key_id, if: :any_apple_credentials_exist?
+  validates_presence_of :apple_key_pem_b64, if: :any_apple_credentials_exist?
   validates :public_feed, uniqueness: { scope: :private_feed,
                                         message: "can only have one credential per public and private feed" }
   validates :public_feed, exclusion: { in: ->(apple_credential) { [apple_credential.private_feed] } }
+
+  def any_apple_credentials_exist?
+    apple_provider_id.present? || apple_key_id.present? || apple_key_pem_b64.present?
+  end
 end
--- a/app/models/apple_credential.rb
+++ b/app/models/apple_credential.rb
@@ -3,6 +3,8 @@ class AppleCredential < ActiveRecord::Base
 
   validates_associated :podcast, if: -> { prx_account_uri.blank? }
   validates_presence_of :prx_account_uri, if: -> { podcast.blank? }
+  validates_presence_of :apple_key_id
+  validates_presence_of :apple_key_pem_b64
   validate :podcast_and_prx_account_uri_not_both_set
 
   def podcast_and_prx_account_uri_not_both_set
--- a/app/models/apple_credential.rb
+++ b/app/models/apple_credential.rb
@@ -1,0 +1,14 @@
+class AppleCredential < ActiveRecord::Base
+  belongs_to :podcast, -> { with_deleted }
+
+  validates_associated :podcast, if: -> { prx_account_uri.blank? }
+  validates_presence_of :prx_account_uri, if: -> { podcast.blank? }
+  validate :podcast_and_prx_account_uri_not_both_set
+
+  def podcast_and_prx_account_uri_not_both_set
+    if podcast.present? && prx_account_uri.present?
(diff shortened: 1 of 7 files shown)


app/models/apple_credential.rb
```diff
@@ -0,0 +1,18 @@
+class AppleCredential < ActiveRecord::Base
+  belongs_to :podcast, -> { with_deleted }
+
+  validates_associated :podcast, if: -> { prx_account_uri.blank? }
+  validates_presence_of :prx_account_uri, if: -> { podcast.blank? }
```
kookster, line 5: yeah, this seems right - we only associate accounts at the show/podcast level otherwise, and this does seem like the same level, rather than setting up credentials per show

cavis:
So per [this slack thread](https://prx.slack.com/archives/G0FKLK2Q5/p1670012340441439), need some changes to this PR.

The ask is now that an Apple Credential ties 2 `Feed` models together, the Public one and the Apple Subscriber (probably paid-ad-free) one.

So:

- [ ] Rollback schema and change it - `AppleCredential` belongs_to a `public_feed` and belongs_to a `private_feed` (naming is hard)
- [ ] Validate both exist, for now
- [ ] Validate unique public/private feed combination
- [ ] No more account_uri or podcast_id foreign keys
- [ ] The tests!

db/migrate/20221116182526_create_apple_credentials.rb
```diff
@@ -0,0 +1,13 @@
+class CreateAppleCredentials < ActiveRecord::Migration
+  def change
+    create_table :apple_credentials do |t|
+      t.references :podcast, index: true, foreign_key: true
+      t.string :prx_account_uri
+
+      t.string :apple_key_id
+      t.text :apple_key_pem_b64
```
kookster, line 8: yeah, `text` should do it

app/models/apple_credential.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+class AppleCredential < ActiveRecord::Base
+  belongs_to :public_feed, -> { with_deleted }, class_name: "Feed"
+  belongs_to :private_feed, -> { with_deleted }, class_name: "Feed"
```
cavis, line 5: I'm realizing that `Feed`s don't have a `deleted_at` column.  So this `-> { with_deleted }` scope can be removed.  I'm actually surprised that doesn't throw an error?  `Feed.with_deleted.count` certainly does.

db/migrate/20221116182526_create_apple_credentials.rb
```diff
@@ -0,0 +1,20 @@
+class CreateAppleCredentials < ActiveRecord::Migration
+  def change
+
```
cavis, line 3: Some extra/blank lines in this file?  Generally, I'd only leave a space where you're trying to call out some logical difference between pieces of code.  (Like between the `create_table` block here, and the 2 `add_foreign_key`s below it).

app/models/apple_credential.rb
a-maci29: Can that syntax be deleted without any other changes? I can take it out and re-commit.

cavis: Yep - and hopefully the tests would break if the syntax was invalid.

db/migrate/20221116182526_create_apple_credentials.rb
a-maci29: We don't generally want to edit migrations directly, if I'm remembering right...? Or am I thinking of something else?

app/models/apple_credential.rb
a-maci29: Ran the file individually and got no errors

db/migrate/20221116182526_create_apple_credentials.rb
cavis: You're thinking of `schema.rb` - which gets automatically changed when you run `db:migrate`.

app/models/apple_credential.rb
```diff
@@ -0,0 +1,16 @@
...
+class AppleCredential < ActiveRecord::Base
+  belongs_to :public_feed, class_name: "Feed"
+  belongs_to :private_feed, class_name: "Feed"
+
+  validates_presence_of :public_feed
+  validates_presence_of :private_feed
+  validates_associated :public_feed
+  validates_associated :private_feed
+  validates_presence_of :apple_key_id
+  validates_presence_of :apple_key_pem_b64
```
cavis, line 12: Okay, so i think the remaining TODOs from the slack thread are:

- [ ] Also validate presence of the new `apple_provider_id` field (+test)
- [ ] Add a [conditional validation](https://guides.rubyonrails.org/active_record_validations.html#conditional-validation) method on the 3 `apple_***` fields here
    - Called `def has_any_key_field?` or something
    - Check if any of the 3 `apple_***` fields are `.present?`
    - The result should be - if any `apple_***` field is present, all 3 must be present.  If all 3 are blank, that's also valid.  (+test!)

db/migrate/20221116182526_create_apple_credentials.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+class CreateAppleCredentials < ActiveRecord::Migration
+  def change
+      create_table :apple_credentials do |t|
```
cavis, line 5: Indentation off?

a-maci29: Redo hopefully one last time, or let into next migration?

svevang (approved):

PR #469, opened 2022-12-04
Upgrade to Rails 7 and Ruby 2.7

kookster:
Fixes #386 

It has many related library changes, implemented, and currently linked in the `Gemfile` as branches:
- https://github.com/PRX/Infrastructure/pull/663
- https://github.com/PRX/roar-rails/pull/1
- https://github.com/PRX/hal_api-rails/pull/30
- https://github.com/PRX/prx_auth-rails/pull/17
- https://github.com/kookster/say_when/pull/17

Removes the `announcements` gem, switched ActiveJobs to Shoryuken workers

PR #471, opened 2022-12-05
Feat/dont use episodes for lookup

svevang:
Fixes up a couple problems and regressions introduced in #467 and in the hypermedia traversal, uses models that are closer to the targeted type (e.g. use PodcastDelivery for locally looking up  PodcastDeliveryFile instead of Apple::Episode).


app/models/apple/podcast_delivery_file.rb
```diff
@@ -11,6 +11,8 @@ class PodcastDeliveryFile < ActiveRecord::Base
 
     delegate :apple_episode_id, to: :podcast_delivery
 
+    PODCAST_DELIVERY_ID_ATTR = "podcast_delivery_id"
```
cavis, line 14: Off topic: just remembering that #469 brings in the Augury rubocop settings (which i think does single quotes?).  I suppose we may want to lint the whole repo sometime after that hits.  Ah well, a problem for future selves.

cavis: But hey, constants +1!

cavis (approved):
Build seems to be failing (unrelated test?) but the codes here looks good to me.

app/models/apple/podcast_delivery_file.rb
svevang: Yeah, there was a rubocop config in feeder that this is using.  When I merge master (and get the new config in place) I can re-lint these.

app/models/apple/api.rb
```diff
@@ -160,7 +160,7 @@ def unwrap_bridge_response(resp)
     end
 
     def api_bridge_url
-      URI.parse(ENV.fetch("APPLE_API_BRIDGE_URL"))
+      URI.join(ENV.fetch("APPLE_API_BRIDGE_URL"), "/bridge")
```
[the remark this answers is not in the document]
svevang: Good point, maybe the way forward there is to fix that on the lambda side.

PR #479, opened 2022-12-09
Feat/publisher addresses two feeds

### LINKED ISSUE #470: Apple publisher addresses two feeds
Refactor the apple publisher model to take in:

- [x] A primary/subscribe-able public feed
- [x] A private feed that is the source of the premium content.
- [x] Apple API credentials

- [x] Pull from the private feed + CDN (w/ a token if needed)

svevang:
Closes #470 

More missing pieces:

- Apple::Publisher takes in two feeds, public and private (per #468)
- Updates the audio container reference on the apple episode
- Marks the episodes as published. They default to `DRAFTING` after they have the subscriber audio uploaded.

### DIFF
--- a/app/models/apple/podcast_container.rb
+++ b/app/models/apple/podcast_container.rb
@@ -45,6 +45,8 @@ module Apple
       join_on_apple_episode_id(episodes_to_create, new_containers_response).each do |ep, row|
         upsert_podcast_container(ep, row)
       end
+
+      episodes_to_create
     end
 
     def self.upsert_podcast_container(episode, row)
--- a/app/models/apple/podcast_container.rb
+++ b/app/models/apple/podcast_container.rb
@@ -172,5 +172,9 @@ module Apple
     def podcast_deliveries_url
       apple_data.dig("relationships", "podcastDeliveries", "links", "related")
     end
+
+    def podcast_container_id
+      id
+    end
   end
 end
--- a/app/models/apple/podcast_container.rb
+++ b/app/models/apple/podcast_container.rb
@@ -7,6 +7,7 @@ module Apple
     serialize :api_response, JSON
 
     has_many :podcast_deliveries
+    has_many :podcast_delivery_files, through: :podcast_deliveries
     belongs_to :episode, class_name: "::Episode"
 
     def self.update_podcast_container_file_metadata(api, episodes)
--- a/app/models/apple/podcast_container.rb
+++ b/app/models/apple/podcast_container.rb
@@ -42,7 +42,7 @@ module Apple
         api.bridge_remote_and_retry!("createPodcastContainers",
                                      create_podcast_containers_bridge_params(api, episodes_to_create))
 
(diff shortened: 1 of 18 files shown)


app/models/apple/podcast_container.rb
```diff
@@ -42,9 +43,11 @@ def self.create_podcast_containers(api, episodes)
         api.bridge_remote_and_retry!("createPodcastContainers",
                                      create_podcast_containers_bridge_params(api, episodes_to_create))
 
-      join_on_apple_episode_id(episodes_to_create, new_containers_response).each do |ep, row|
+      join_on_apple_episode_id(episodes_to_create, new_containers_response).map do |ep, row|
```
cavis, line 46: Why the map?  It seems like you're returning `episodes_to_create` after this.

app/models/apple/show.rb
```diff
@@ -2,15 +2,17 @@
 
 module Apple
   class Show
-    attr_reader :feed, :api
+    attr_reader :public_feed,
```
cavis, line 5: Could also change this to have `:apple_credential` instead, and pull the public/private feed from that?

But this also works.  And is maybe cleaner to test.

app/models/episode.rb
```diff
@@ -84,7 +84,7 @@ def self.story_uri(story)
   end
 
   def apple?
-    categories_include?([APPLE_ADFREE_TAG, APPLE_ONLY_TAG].freeze)
+    categories_include?([APPLE_FREEMIUM_TAG, APPLE_ONLY_TAG].freeze)
```
cavis, line 87: Maybe I already asked this ... but I'm still not entirely clear why we need these hardcoded "apple" tags?

Doesn't the `apple_credential.public_feed.filtered_episodes` and `apple_credential.private_feed.filtered_episodes` already tell us what content to send to apple?

app/models/feed_token.rb
```diff
@@ -10,7 +10,15 @@ def set_defaults
...
+  def self.feed_published_url_with_token(some_feed)
+    raise "missing token for private feed" unless some_feed.tokens.any?
+
+    token = some_feed.tokens.first.token
+
+    # use the feed's published_url, but replace the path with the token using substitution
+    some_feed.
+      published_url.
+      sub("{?auth}", "?auth=#{token}").
+      sub("f.prxu.org", "p.prxu.org")
```
cavis, line 22: I don't think this is necessary?

The `feed.published_url` already checks if it's `private?` - and varies the host to `ENV['FEEDER_CDN_PRIVATE_HOST']` if needed.

cavis (changes requested):
One unnecessary hardcoding of the feed hosts (`f.prxu.org`, etc) in the feed token.  Otherwise looks good - just some questions.

Does feel like the `AppleCredential` model will change/replace some things in here.  But I'd be fine with merging this first, then fixing that up.

svevang:
@cavis Moved the feedback over here: https://github.com/PRX/feeder.prx.org/pull/488

PR #490, opened 2022-12-29
Parallel uploads for hosted bridge

svevang:
More stress testing and testing with larger (flac) files revealed a few issues.

- Some ECONNRESET errors coming from the ApiGateway code in the SAM local dev setup for heavily parallel sets of requests.  I'm ignoring that for now, and serializing the uploads in development.  This should be easier to deal with on a home broadband connection, and at least allow some reasonable testing.
- The Api Gateway timeout (30 seconds) was being hit from the lambda side. https://github.com/PRX/Infrastructure/pull/666 addresses this and switches the "trigger" to using the Lambda URL, which honors the overall lambda timeout. (Merge that one first)
- Blocking IO makes parallel requests here pretty easy -- I added the parallel gem


cavis (approved):
👍 Looks good to me!

PR #491, opened 2023-01-03
Fixup/exception handling

svevang:
Discovered that the Apple::ApiError exceptions were not being formatted correctly.  This PR:

- Fixes a problem where exceptions were not being raised.
- Formats the exception message and response body.

### DIFF
--- a/app/models/apple/publisher.rb
+++ b/app/models/apple/publisher.rb
@@ -57,9 +57,6 @@ module Apple
 
       # success
       SyncLog.create!(feeder_id: public_feed.id, feeder_type: :feeds, external_id: show.apple_id)
-    rescue Apple::ApiError => e
-      SyncLog.create!(feeder_id: public_feed.id, feeder_type: :feeds)
-      raise e
     end
 
     def wait_for_upload_processing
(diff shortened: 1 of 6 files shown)


app/models/apple/publisher.rb
```diff
@@ -57,9 +57,6 @@ def publish!
 
       # success
       SyncLog.create!(feeder_id: public_feed.id, feeder_type: :feeds, external_id: show.apple_id)
-    rescue Apple::ApiError => e
-      SyncLog.create!(feeder_id: public_feed.id, feeder_type: :feeds)
```
cavis, line 61: No more `SyncLog` records for these?

cavis (approved):
👍 looks good!  (One minor question).

app/models/apple/publisher.rb
svevang: Retrying here gets us another log entry -- which is effectively the same?   The sync logs are doing a bit less than I had hoped. But still provide the reference. 

A 1:1 apple state model here might make sense, but the show and episodes need it the least.

PR #492, opened 2023-01-03
Feat/apple api integration

svevang:
Sets up an initial Apple Podcasts Connect implementation in Feeder.

- Sets up a multi-tenant system to manage credentials with a fallback to a set of system-wide credentials (in the case we're managing the content under the aegis of our podcasts connect account.
- Uses a _public_ and _private_ feed to model the public versus Apple freemium / premium content: A subscriber will see the private feed's audio.
- Manages API access through a "bridge" lambda: Api requests (including file uploads) are pooled and dispatched from the feeder code (in this PR) over to the Typescript / NodeJS Api bridge lambda https://github.com/PRX/api-bridge-lambda. Rate limiting is handled over there. Results are bundled up from the bridge lambda and returned to the Api handling code. Errors that are closer to Api semantics are retried and errors are thrown from Feeder Ruby.
- Guards for existing media: In the case that files are already uploaded into Podcasts Connect, the default action is to probe for existing Apple media model instances that are descendant from the episode and create matching local state.
- Apple publishing is initiated when the podcast is published and only if the dependent feeds have an `apple_credentials` associated, marking the feed as "Apple publishable".

To run an example Apple publishing routine, you need to have Apple Api credentials:
- `provider_id`
- `key_id`
- `key` (private key used to sign the Apple JWT)

```rb
# Set up the bridge URL in your dotenv and set up the Apple creds in your dotenv
api = Apple::Api.from_env
# Alternatively, you can build an `Apple::Api` instance from parameters you have in hand:
#  api = Apple::Api.new(provider_id: apple_provider_id, key_id: apple_key_id, key: apple_key_pem)

apple_publisher = Apple::Publisher.new(api: api,
                                       public_feed: pub_feed,
                                       private_feed: priv_feed)
apple_publisher.publish!
```

More Apple namespaced Publisher and Api examples can be seen in those classes' tests, as well as the updated PublishFeedJob.


app/models/apple/publisher.rb
```diff
@@ -0,0 +1,144 @@
...
+                :private_feed,
+                :api,
+                :show
+
+    def self.from_apple_credential(apple_credential)
+      api = Apple::Api.from_apple_credentials(apple_credential)
+
+      new(api: api,
+          public_feed: apple_credential.public_feed,
+          private_feed: apple_credential.private_feed)
```
svevang, line 15: Here's where the apple creds split up the feeds.

app/models/apple/podcast_delivery.rb
```diff
@@ -0,0 +1,167 @@
...
+      podcast_containers = episodes.map do |ep|
+        if ep.podcast_container.present?
+          ep.podcast_container
+        else
+          Rails.logger.error("Missing podcast container for episode #{ep.feeder_id}")
+          nil
+        end
+      end.compact
+
+      results = get_podcast_deliveries_via_containers(api, podcast_containers)
```
svevang, line 31: Most of the models follow this pattern for updating existing state:

1) Use the parent model to query for children
2) Reformat the results by querying the main api endpoint for the child model.

app/jobs/publish_feed_job.rb
```diff
@@ -9,7 +9,25 @@ class PublishFeedJob < ApplicationJob
   attr_accessor :podcast, :episodes, :rss, :put_object, :copy_object
 
   def perform(podcast)
-    podcast.feeds.each { |feed| save_file(podcast, feed) }
+    podcast.feeds.each { |feed| publish_feed(podcast, feed) }
+  end
+
+  def publish_feed(podcast, feed)
+    publish_apple(feed)
```
svevang, line 16: The hook into the publishing pipeline.

This method call does nothing unless the apple credentials are set up for the feed in question. And this feed must be on the "public" feed on the credentials.

app/models/apple/episode.rb
```diff
@@ -0,0 +1,266 @@
...
+    def item_guid
+      feeder_episode.item_guid
+    end
+
+    def feeder_id
+      feeder_episode.id
+    end
+
+    def enclosure_url
+      feeder_episode.enclosure_url
```
svevang, line 75: Since we're hitting the Dovetail CDN, this should include `?noImp`

app/models/apple/episode.rb
```diff
@@ -0,0 +1,266 @@
...
+    end
+
+    def initialize(show:, feeder_episode:, api: nil)
+      @show = show
+      @feeder_episode = feeder_episode
+      @api = api || Apple::Api.from_env
+    end
+
+    def item_guid
+      feeder_episode.item_guid
```
svevang, line 67: The item guid here should work, but it's probably better to use an unchanging / unmapped regular guid here

app/models/apple_credential.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+class AppleCredential < ApplicationRecord
```
cavis, line 3: Hindsight:

Should this have been `app/models/apple/credential`?

test/representers/api/auth/feed_representer_test.rb
```diff
@@ -27,7 +27,7 @@
   end
 
   it 'has a feed rss link' do
-    _(json['_links']['prx:private-feed']['href']).must_equal feed.published_url
+    _(json['_links']['prx:private-feed']['href']).must_equal "#{feed.published_url}{?auth}"
```
cavis, line 30: Yeah, digging down a bit, it looks like i made `feeds.private` default to `true` in the db.  So this href is now something like `https://p.prxu.org/jjgo/myfeed5/feed-rss.xml{?auth}{?auth}` in this test.

cavis (changes requested):
Breezed over some of the `app/models/apple/***` stuff that past-Ryan had looked at.  All looks good to me except that Feed representer change, which I think you'd already realized.

app/models/episode.rb
```diff
@@ -72,6 +83,36 @@ def self.story_uri(story)
...
+    apple_delivery_file_errors.present?
+  end
+
+  def apple_delivery_file_errors
+    # TODO: for now these are all considered audio files
+
+    apple_delivery_files.map { |p| p.asset_processing_state['errors'] }.flatten
+  end
+
+  def categories_include?(match_tags)
```
cavis, line 106: Isn't this all duplicative of the Feed?

app/models/apple/episode.rb
```diff
@@ -0,0 +1,273 @@
...
+
+        SyncLog.
+          create(feeder_id: ep.feeder_episode.id, feeder_type: :episodes, external_id: apple_id)
+      end
+    end
+
+    def self.add_no_imp_param(url)
+      # use the URI class to add a 'noImp' query param
+      url = URI.parse(url)
+      decoded_query = URI.decode_www_form(url.query.to_s) << ['noImp', '1']
```
cavis, line 63: Ah, ye olde `?noImp`.

Alternatively, could we set a bot-ish UA when making requests to Dovetail?  That's probably a good idea anyways, so we can tell where traffic is coming from.  And there's prior art for that in the podcast world (bots/scrapers like the Spotify one).

Then we can mark the UA as a "bot" in prx-podagent.  And get `is_duplicate=true` records in BQ, so we can debug what's going out to Apple.

app/models/apple/api.rb
```diff
@@ -0,0 +1,290 @@
...
+    end
+
+    def list_shows
+      get('shows')
+    end
+
+    def set_headers(req)
+      req['Authorization'] = "Bearer #{jwt}"
+      req['Content-Type'] = 'application/json'
+      req['User-Agent'] = "PRX-Feeder-Apple/1.0 (Rails-#{Rails.env})"
```
cavis, line 73: Oh interesting - I guess I thought the api-bridge-lambda would GET the enclosure from DTR (the redirect), download from the dovetail CDN, then upload it to Apple.  But I guess it's Feeder (rails) doing the downloading?

In any case, it's a good idea to set a user-agent.

app/models/episode.rb
cavis: I think this and `normalize_category` are unused in this class now.

cavis (approved):
Couple unused methods (I think?) ... but otherwise that's everything.

Looks good to me!

PR #494, opened 2023-01-04
Find by item guid

### LINKED ISSUE #487: Add API find method by RSS/public GUID
Give the ability to look up an episode using the custom GUID, or possibly an external_id specified by a publisher.

cavis:
Closes #487.

- [x] Adds `Episode.find_by_item_guid` helper to look for _either_ `original_guid`s OR our generated guids when nil
- [x] Adds 2 routes:
    ```
    GET /api/v1/podcasts/:podcast_id/guids/:id
    GET /api/v1/authorization/podcasts/:podcast_id/guids/:id
    ```
- [x] Add links to podcast representers

### DIFF
--- a/app/models/episode.rb
+++ b/app/models/episode.rb
@@ -72,6 +72,18 @@ class Episode < ApplicationRecord
     (story.links['self'].href || '').gsub('/authorization/', '/')
   end
 
+  def self.generate_item_guid(podcast_id, episode_guid)
+    "prx_#{podcast_id}_#{episode_guid}"
+  end
+
+  def self.decode_item_guid(item_guid)
+    item_guid.sub(/^prx_[0-9]+_/, '') if item_guid&.starts_with?('prx_')
+  end
+
+  def self.find_by_item_guid(guid)
+    where(original_guid: guid).or(where(original_guid: nil, guid: decode_item_guid(guid))).first
+  end
+
   def publish_updated
     podcast.publish_updated if podcast
   end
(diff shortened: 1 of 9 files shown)


app/representers/api/podcast_representer.rb
```diff
@@ -68,6 +68,13 @@ class Api::PodcastRepresenter < Api::BaseRepresenter
     } if represented.id
   end
 
+  link :guid do
+    {
+      href: api_podcast_guid_path_template(podcast_id: represented.id.to_s, id: '{guid}'),
```
cavis, line 73: I opted to only add to podcast representers (not the API root).  And to call the param `guid` instead of `id`.  But I have no strong opinion.

Shows up something like:

image

app/models/episode.rb
```diff
@@ -72,6 +72,18 @@ def self.story_uri(story)
...
+  def self.generate_item_guid(podcast_id, episode_guid)
+    "prx_#{podcast_id}_#{episode_guid}"
+  end
+
+  def self.decode_item_guid(item_guid)
+    item_guid.sub(/^prx_[0-9]+_/, '') if item_guid&.starts_with?('prx_')
+  end
+
+  def self.find_by_item_guid(guid)
+    where(original_guid: guid).or(where(original_guid: nil, guid: decode_item_guid(guid))).first
```
cavis, line 84: There was probably some clever SQL to generate the `prx_123_abcd` style guid in postgres?  But I opted to just decode it in Ruby.

app/representers/api/podcast_representer.rb
cavis: Update: okay, I added to the API root and the Authorization root.  Because it seemed useful as I was writing some docs.

svevang (approved):

kookster (commented):
Had a minor syntax question about a `self.`, otherwise looks good

PR #495, opened 2023-01-06
Bump prx_auth-rails

cavis:
Fixes some issues with inserting the middleware more than once, (and one of the two times without the right ENV configs!)

Should deploy the gem first PRX/prx_auth-rails#18.


cavis:
Seems pretty non-controversial.  Merging!

PR #500, opened 2023-01-10
Revert "Revert "Find by item guid""

cavis:
🚧 WIP.

But this time... don't break caching.

### DIFF
--- a/app/controllers/api/episodes_controller.rb
+++ b/app/controllers/api/episodes_controller.rb
@@ -3,9 +3,9 @@ class Api::EpisodesController < Api::BaseController
 
   api_versions :v1
   represent_with Api::EpisodeRepresenter
-  find_method :find_by_guid
   filter_resources_by :podcast_id
 
+  before_action :set_find_method
   after_action :process_media, only: [:create, :update]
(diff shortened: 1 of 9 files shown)


app/controllers/api/episodes_controller.rb
```diff
@@ -3,11 +3,12 @@ class Api::EpisodesController < Api::BaseController
 
   api_versions :v1
   represent_with Api::EpisodeRepresenter
-  find_method :find_by_guid
   filter_resources_by :podcast_id
 
+  find_method :find_by_guid
   after_action :process_media, only: [:create, :update]
   after_action :publish, only: [:create, :update, :destroy]
+  allow_params :show, [:id, :podcast_id, :guid_resource, :api_version, :format, :zoom]
```
cavis, line 11: Are these not additive to the `allow_params` already in `Api::BaseController`?

I guess maybe it just [replaces the params](https://github.com/PRX/hal_api-rails/blob/044e132a89f95a72d2f55a79f335547830e46d4b/lib/hal_api/controller/actions.rb#L106) for the action every time you call it.

PR #502, opened 2023-01-11
Redirect with episode guid, not id

### LINKED ISSUE #501: Fix unpublished episode redirect
If you try to GET an unpublished episode at `/api/v1/episodes/:guid` (or the podcast filtered version) ... and you _include_ your valid auth token, Feeder tries to [redirect you](https://github.com/PRX/feeder.prx.org/blob/master/app/controllers/api/episodes_controller.rb#L67) to the `/api/v1/authorization/episodes/:id` version.

But - it seems to be including the integer `episode.id` at the moment, instead of the guid one.  So you get a 404.  Fix it fix it fix it!

cavis:
Fixes #501.


kookster (approved):
nice catch, lgtm

Issue #510, opened 2023-01-13
Schema changes (parity w/CMS)


cavis:
Figure out what additional fields/tables we need to be compatible with CMS.  Don't want to leave anything behind that we might need!

CMS Tables to check:

```
audio_file_templates
audio_files
audio_version_templates
audio_versions
distributions
episode_imports
piece_images
pieces
podcast_imports
series
series_images
story_distributions
taggings
tags
```

cavis:
My notes:

## Series Images

Right now, our PodcastSeriesHandler _only_ copies over the image file (href).

`FeedImage` and `ItunesImage` have a title/description that are all null in prod. But no other fields.

CMS Images have a caption and credit, but no alt. There are 108 CMS v4 series with non-blank image caption or credit.

We should probably add these fields and start syncing them.

We'll also lose the original/source uploaded file (it's a CMS `/pub/` signed asset url). But I think that's okay.

## Podcast Imports

We don't have this at all in Feeder. But do we actually care about the history?

I vote we rebuild this in Feeder using new tables, and don't worry about syncing the ~168 existing CMS ones.

## Audio Version Templates

We do lose the validations on these. And I guess their concept entirely. Feeder knows how many segments an episode should have (`episodes.segment_count`, and the label (`episodes.audio_version`). But nothing about what is valid for a Podcast.

So we'd lose length, segment-count, and content-type validations. I vote that's okay, and that we ditch the concept of setting up "templates" for a podcast. (Except in Augury).

Feeder actually determines media content type by analyzing the files. And that's what DTR uses to do its "redirect to videos" logic anyways.

If we _really_ need that segment count validation, we could add an array-of-ints field to Podcasts.

## Pieces

For some reason, CMS pieces have account_ids? But Feeder episodes have to belong to a Podcast, which has the account. So I don't think we're losing anything.

We also track the `creator_id` on pieces, which Feeder doesn't do. I'm okay with that? Though we're making schema changes anyways, so maybe we should just bring in the user_ids that created the series/piece.

Production Notes is also missing. There are only 547 pieces with that set, but it was in the Publish UI, so we should probably bring it over.

Now the hard one: status. Feeder doesn't have this gatekeeping concept, but I'm suspicious we may need it? Otherwise could DTR get non-OK episode JSON? Also, status drives the "is this ready to publish?" logic in Publish.

## Piece Images

As with the series, we're missing caption/credit. And CMS had no alt.

## Audio Files

[...]

PR #512, opened 2023-01-17
CMS parity schema

cavis:
Recommended changes from #510.


app/models/episode_story_handler.rb
```diff
@@ -93,12 +94,15 @@ def update_audio
...
 
   def update_image
-    image_url = begin
-      story.objects['prx:image'].links['original'].href
-    rescue StandardError
-      nil
+    cms_image = story.objects['prx:image'] rescue nil
+    cms_href = cms_image.links['original'].href rescue nil
+
+    episode.image_file = cms_href
```
cavis, line 100: NOTE: if the CMS Story _had no image_ we still need to set this to nil.  Because that triggers the `images.destroy_all` functionality in the setter.

But, only sync the caption/credit if we still have an `image_file` when all is said and done.

app/models/episode_story_handler.rb
```diff
@@ -93,12 +94,15 @@ def update_audio
   end
 
   def update_image
-    image_url = begin
-      story.objects['prx:image'].links['original'].href
-    rescue StandardError
-      nil
+    cms_image = story.objects['prx:image'] rescue nil
```
cavis, line 97: Note: I sort of ignored Rubocop in these CMS-handler-y files.  Instead I stuck to existing themes and went for minimal diff.

These will all go away with CMS anyways!

app/views/podcasts/show.rss.builder
```diff
@@ -32,10 +32,10 @@ xml.rss 'xmlns:atom' => 'http://www.w3.org/2005/Atom',
     xml.image do
       xml.url @feed_image.url
       xml.title @feed.title || @podcast.title
-      xml.link @feed_image.link || @podcast.link
+      xml.link @podcast.link
```
cavis, line 35: [The docs](https://www.rssboard.org/rss-specification#ltimagegtSubelementOfLtchannelgt) indicate:

> Note, in practice the image <title> and <link> should have the same value as the channel's <title> and <link>

So I ended up dropping `link` from all the images (they were all null anyways).

app/views/podcasts/show.rss.builder
```diff
@@ -32,10 +32,10 @@ xml.rss 'xmlns:atom' => 'http://www.w3.org/2005/Atom',
...
       xml.width @feed_image.width
       xml.height @feed_image.height
-      xml.description @feed_image.description unless @feed_image.description.blank?
+      xml.description @feed.subtitle || @podcast.subtitle
```
cavis, line 38: Again, the docs say:

> <description> contains text that is included in the TITLE attribute of the link formed around the image in the HTML rendering.

I'm not entirely sure what we should be putting there?  But definitely not anything describing the image ... this is a link to the show's website after all.  So I thought... maybe subtitle instead?

In prod, all our image descriptions are null.  So currently not even including this field.

Open to suggestions.

db/migrate/20230113203436_fixup_images.rb
```diff
@@ -0,0 +1,14 @@
+class FixupImages < ActiveRecord::Migration[7.0]
+  def change
+    %i[episode_images feed_images itunes_images].each do |tbl|
+      add_column tbl, :alt_text, :string
+      add_column tbl, :caption, :string
+      add_column tbl, :credit, :string
```
cavis, line 6: Per designs, I'm proactively adding this metadata.  Even though there's nowhere to put it in the RSS yes.

test/controllers/api/auth/feeds_controller_test.rb
```diff
@@ -102,18 +102,18 @@ class << @controller; attr_accessor :prx_auth_token; end
         _(feed.feed_images.count).must_equal 0
         fua = feed.updated_at
 
-        update_hash = { feedImage: { href: url1, description: 'd1' } }
+        update_hash = { feedImage: { href: url1, caption: 'd1' } }
```
cavis, line 105: Some tests were using now gone fields to track which image is which in the API.

Switched them to fields that exist after these migrations.

kookster (approved):
lgtm!

app/models/episode_story_handler.rb
```diff
@@ -63,6 +63,7 @@ def update_story_attributes
     episode.description = sa[:description]
     episode.content = sa[:description]
     episode.categories = sa[:tags]
+    episode.production_notes = sa[:production_notes]
```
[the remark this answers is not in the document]
cavis: It did make its way into the Publish UI - and there were a few non-blank values.  But yeah, I can't remember exactly why we decided to add that.

PR #515, opened 2023-01-19
Setup asset pipeline

### LINKED ISSUE #473: Asset pipeline + Bootstrap
Get the asset pipeline (sprockets) working, and install bootstrap 5.  A bit similar to https://github.com/PRX/id.prx.org/issues/209.

- [x] Get asset pipeline working
- [x] Install the bootstrap gem
- [x] Copy over some the "shared" stylings from Augury/ID (`bootstrap-variables.scss` and all the shared files)

cavis:
Merge after #513.

Closes #473.

- [x] Adds bootstrap gem
- [x] Copied over all the shared scss from Augury
- [x] Stubbed out the basic layout for this app (relying on `content_for`)
- [x] Made a `/fake` index/show route to prove sass/stimulus/turbo were working


Dockerfile
```diff
@@ -4,7 +4,7 @@ LABEL org.prx.app="yes"
 LABEL org.prx.spire.publish.ecr="RAILS_APP"
 
 RUN apk -U upgrade && apk add --no-cache \
-    tzdata postgresql-dev postgresql-client build-base bash coreutils git
+    tzdata postgresql-dev postgresql-client build-base bash coreutils git nodejs
```
cavis, line 7: For some reason, this is required by execjs, which is required by autoprefixer-rails, which is part of bootstrap.

I could have picked a [different runtime](https://github.com/rails/execjs) instead ... but I sort of figured we'd need node in here at some point (for linting).

svevang (approved):

PR #520, opened 2023-01-20
Standardrb + erb-lint formatting

cavis:
Install and lint _everything_ with [standard](https://github.com/testdouble/standard) and [erb-lint](https://github.com/Shopify/erb-lint) (configured to [use standard](https://thoughtbot.com/blog/a-standard-way-to-lint-your-views)).

Setup a [.git-blame-ignore-revs](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view) file.


svevang (approved):
Looks good.  +1 for linting and not choosing rubocop rules.

Per the rev ignore list, it's helpful to set up a global config if you're using your local git client:

```
git config --global blame.ignoreRevsFile .git-blame-ignore-revs
```

With this approach I am 100% in favor of enormous linting commits. 🎉

PR #524, opened 2023-01-24
Format everything with prettier

cavis:
Installed prettier, and formatted everything.

NOTE: prettier does not touch html.erb files right now.  We have `erb_lint` doing that.  I'm a bit disappointed `erblint` doesn't fix HTML/ruby-block indent issues... but I think I'm willing to punt on that for now.  They have an open issue for it.


.gitignore
```diff
@@ -43,3 +43,5 @@ notes*
 /build
 
 .DS_Store
+node_modules
+package-lock.json
```
cavis, line 47: I've opted to ignore `package-lock.json` (for now).  The only thing in there is prettier, I have the version explicitly set in `package.json`, and I don't really want it locking the node version right now.

(Even though I did lock nodejs to latest LTS in `.tool-versions`).

svevang (approved):

PR #527, opened 2023-01-24
Init toasts for flash messages

cavis:
Use toasts to display flash messages.  Setup a demo for `flash.now[:notice]`.

image


brandonhundt (approved):
LGTM :shipit: @cavis

PR #528, opened 2023-01-24
Init pundit error handling

cavis:
Feeder already has the pundit gem for its API endpoints.

This adds it for the top-level controller, and adds some custom handling.  Users will normally never see these pages unless they're doing something nefarious.  But good to handle them.

image


app/views/errors/forbidden.html.erb
```diff
@@ -0,0 +1,4 @@
+<h5 class="text-center p-5"><%= t "pundit.title" %></h5>
```
brandonhundt, line 1: Suggestion:

`<h1 class="text-center p-5 pb-0 fw-bold"> ...`

brandonhundt (approved):
1 very minor display thought in the comments. Otherwise, this looks good to me. :shipit:

PR #537, opened 2023-01-25
Card example

brandonhundt:
Adding a fully styled card with a lot of different options that will be used in some cards, but not all. 

Is it too heavy on bootstrap classes?

### DIFF
--- a/app/assets/stylesheets/card.scss
+++ b/app/assets/stylesheets/card.scss
@@ -1,0 +1,57 @@
+.card {
+  .card-header,
+  .card-footer {
+    background-color: transparent;
+    display: flex;
+    justify-content: space-between;
+    align-items: center;
+
+    &.bg-info,
+    &.bg-primary,
+    &.bg-success,
+    &.bg-dark,
+    &.bg-warning,
+    &.bg-danger {
+      border-bottom: none;
+    }
+  }
+
+  .card-title {
+    font-weight: bold;
+    text-transform: uppercase;
+    margin: 0;
+  }
+
+  .status-text {
+    font-size: 0.875rem;
+    line-height: 1.1;
+    margin-bottom: 0;
+  }
+
+  .btn {
+    padding-top: 0.25rem;
+    padding-bottom: 0.25rem;
+    display: flex;
+    align-items: center;
+
+    .material-icons {
+      align-self: center;
+    }
+
+    &.btn-group {
+      display: block;
+    }
+  }
+}
+
+.bg-info,
+.bg-warning,
+.bg-danger {
+  color: black;
+}
+
+.bg-success,
+.bg-primary,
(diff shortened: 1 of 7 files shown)


app/assets/stylesheets/shared/card.scss
```diff
@@ -1,6 +1,5 @@
 .card {
   margin-bottom: 1em;
-  border-radius: 5px;
```
brandonhundt, line 3: The border radius felt too big.

app/assets/stylesheets/card.scss
```diff
@@ -0,0 +1,57 @@
+.card {
```
brandonhundt, line 1: I didn't write this intending it to be shareable across apps, but it might be able to. For now, keeping it feeder only.

app/views/fake/index.html.erb
```diff
@@ -13,20 +13,74 @@
 
     <div class="col-lg-8">
       <div class="card shadow">
+        <div class="card-header">
+          <h5 class="card-title">Look a card</h5>
+          <a href="#" class="btn btn-primary">Action <span class="material-icons" aria-label="hidden">keyboard_arrow_right</span></a>
+        </div>
```
brandonhundt, line 19: Styles are now applied in the cards.scss file. Can still use light bootstrap classes for alignment only when needed.

app/assets/stylesheets/card.scss
```diff
@@ -0,0 +1,57 @@
...
+  .card-header,
+  .card-footer {
+    background-color: transparent;
+    display: flex;
+    justify-content: space-between;
```
brandonhundt, line 6: If you have a title AND button, this will ensure the 2nd element right aligns.

app/assets/stylesheets/shared/bootstrap-variables.scss
```diff
@@ -73,3 +73,5 @@ $form-floating-input-padding-t: $input-btn-padding-y;
 $form-floating-input-padding-b: $input-btn-padding-y;
 $form-floating-label-opacity: 1;
 $form-floating-label-transform: scale(0.85) translateY(-1.75rem) translateX(0.15rem);
+
+$border-radius: 3px;
```
cavis, line 77: Just noting that the base bootstrap variables define these [in rems](https://github.com/twbs/bootstrap-rubygem/blob/main/assets/stylesheets/bootstrap/_variables.scss#L495).  But if you're cool with pixels, so am I!

cavis (approved):
👍 looks good to me!

app/assets/stylesheets/shared/bootstrap-variables.scss
brandonhundt: fixed

PR #539, opened 2023-01-26
Formbuilder

cavis:
Copies (and does some fixups) on the form field related stuff from Augury.

Inits a demo of single, multiple, and timestamp fields.

To make the demo editable (you're not considered logged in yet) ... open `episode_policy.rb` and change `def create?` to just return true.


app/helpers/feeder_form_builder.rb
```diff
@@ -0,0 +1,153 @@
...
+    add_data(opts, :action, BLANK_ACTION)
+  end
+
+  def add_search_action(opts)
+    add_data(opts, :action, SEARCH_ACTION)
+  end
+
+  def add_slim_select_controller(opts)
+    # TODO: slimselect 2 styles messed up at the moment
+    # add_data(opts, :controller, SLIM_SELECT_CONTROLLER)
```
cavis, line 129: To demo slimselect, uncomment this line.

brandonhundt (approved):
LGTM! I'll work on the slim select styles in a separate branch

PR #540, opened 2023-01-26
updated styles for slim select v2

brandonhundt:
Wooo! Slim Select V2 is not backwards compatible from a styling perspective. But I think I got all the differences.

This will be a big diff as slim-select got rid of some wrapper classes, so many things have been moved around the file.

When we port this over to the other sites, we will need to upgrade slim select, otherwise many selects will have weird styling.

### DIFF
--- a/app/assets/stylesheets/shared/form.scss
+++ b/app/assets/stylesheets/shared/form.scss
@@ -39,115 +39,130 @@
   }
 }
 
-// TODO: these don't really with with slimselect v2
+// These styles work with Slim Select v2, not tested with v1
 .ss-main {
   &.form-select {
-    padding: 0;
+    padding: 0.75rem 0.75rem;
     min-height: $input-height;
     height: auto;
   }
 
-  .ss-single-selected,
-  .ss-multi-selected {
-    padding: $input-btn-padding-y $input-btn-padding-x;
-    line-height: $line-height-base;
-    border: 0;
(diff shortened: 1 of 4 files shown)


app/views/fake/show.html.erb
```diff
@@ -36,7 +36,7 @@
 
           <div class="col-xl-6 mb-4">
             <div class="form-floating">
-              <%= form.select :itunes_type, [["Full", "full"], ["Bonus", "bonus"], ["Invalid", "invalid"]], include_blank: true %>
```
brandonhundt, line 39: I needed to test this because there was some weird arrow appearing in the background. Needed to ensure it wasn't part of the actual select option. (It wasn't!)

app/assets/stylesheets/shared/form.scss
```diff
@@ -39,117 +39,132 @@
...
+  .ss-values .ss-value {
+    animation: none;
+    background-color: $primary;
+    margin: 2px 5px 0 0;
+
+    .ss-value-delete,
+    .ss-value-delete svg {
+      width: 16px;
+      height: 16px;
+      border-left: none;
```
brandonhundt, line 60: Note that previously everything was wrapped in the `.ss-main` selector. Its still a selector, but doesn't apply to everything anymore. 

So the diffs won't line up perfectly here.

PR #544, opened 2023-01-27
Deprecate podcasts move feeds api

cavis:
Prepping for frontend dev -

- [x] Remove `/podcasts/123` RSS endpoint.  I think this is completely unused now?  Hopefully.
- [x] Move `/feeds` (_all the feeds JSON for dovetail router_) to `/api/v1/feeds`.  But also keep serving it at the original endpoint, until we get DTR updated.


kookster (approved):
lgtm

PR #547, opened 2023-01-30
Create styles for episode list cards

brandonhundt:
Realized, I forgot about these when creating the Card PR last week. 

- This sets up the patterns for the two column episode list. `.episode-card` (Check it own on the fake/index path)
- It also touches nav files a little by moving away from bootstrap classes to styles defined in stylesheets as we resolved to do that in the cards PR. Should be cleaner.

### DIFF
--- a/app/views/layouts/_subnav.html.erb
+++ b/app/views/layouts/_subnav.html.erb
@@ -1,8 +1,8 @@
-<nav class="prx-app-nav navbar bg-white d-flex flex-row flex-wrap justify-content-start m-0 p-0">
+<nav class="prx-app-nav navbar bg-white">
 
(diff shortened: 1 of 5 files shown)


app/views/layouts/_subnav.html.erb
```diff
@@ -1,8 +1,8 @@
-<nav class="prx-app-nav navbar bg-white d-flex flex-row flex-wrap justify-content-start m-0 p-0">
+<nav class="prx-app-nav navbar bg-white">
```
brandonhundt, line 1: moved to header.scss

cavis (approved):
👍

PR #549, opened 2023-01-31
Skip setting session for api

svevang:
For api requests, this PR prevents setting a session.

To verify:

- `/fake`: does set a cookie and insert a row in the sessions table
- `/api/v1`: (api base controller) does not set a cookie or insert a row into the sessions table
- `/api/v1/feeds`: feeds controller (application controller), does not set a cookie or insert a row into the sessions table


cavis (approved):
👍

PR #550, opened 2023-01-31
Fix/ops scripts skip session

svevang:
A couple of minor changes:

- skip the sessions table when dumping databases
- make sure to connect to the local dev database with the psql.sh helper


cavis (commented):
👍 good catch!

PR #552, opened 2023-01-31
Wire up nav

### LINKED ISSUE #504: Build Global Nav
After #503.  Build out the [global nav bar at the top](https://www.figma.com/proto/Yewv83Q8btHcaOBWCtGgKs/PRX-Dovetail-with-Navigation?page-id=80%3A2947&node-id=2934%3A80025&viewport=798%2C-3032%2C0.25&scaling=min-zoom&starting-point-node-id=80%3A3420&hide-ui=1), which is (fairly) static, and will eventually be used across apps.

image

- [ ] Build it (leave out "Collections", which we don't have yet)
- [x] Make it _fairly_ bootstrap-only-ish. As far as is reasonable. Since we'll be sharing this across apps.
- [x] Make it responsive
- [x] Build the help dropdown
- [ ] Build the user-avatar + dropdown
  - This data will have to be pulled from the `current_user`... which is a `prx_auth-rails` thing
  - We may need some additional helpers in that gem to get all these apps/links/etc
  - But if we don't have that yet, fake it!  We'll fill in the real data later.
- [ ] Don't hardcode links, so this works in staging and prod.  They should come from ID, I think.
- [ ] I18n the text

cavis:
Closes #504.

Wires up actual ID app data into the user dropdown.  And adds translations for everything.

I also opted to drop the big list of apps in the dropdown, and just added "Metrics" up top for now.

### DIFF
--- a/app/assets/stylesheets/shared/header.scss
+++ b/app/assets/stylesheets/shared/header.scss
@@ -48,7 +48,6 @@
       color: $white;
       text-transform: uppercase;
       font-weight: bold;
-      display: block;
       padding: 0.25rem 1.5rem;
       min-width: 240px;
 
--- a/app/assets/stylesheets/shared/header.scss
+++ b/app/assets/stylesheets/shared/header.scss
@@ -57,15 +57,6 @@
       }
     }
 
-    .dropdown-divider {
-      color: $cyan;
-    }
-
-    .dropdown-header {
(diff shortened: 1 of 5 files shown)


brandonhundt (approved):
Fixed some styling. The rest LGTM.

PR #553, opened 2023-02-01
Wire up subnav

### LINKED ISSUE #505: Build Feeder Subnav
After #503.  Build out the [sub nav bar](https://www.figma.com/proto/Yewv83Q8btHcaOBWCtGgKs/PRX-Dovetail-with-Navigation?page-id=80%3A2947&node-id=2934%3A80025&viewport=798%2C-3032%2C0.25&scaling=min-zoom&starting-point-node-id=80%3A3420&hide-ui=1), which has some dynamic content to whatever page you're on.

image

- [x] Make it _fairly_ bootstrap-ish, and responsive.
- [x] Add a before action to application controller (or whatever we're using as the parent to the HTML controllers).
  - Load the complete list of policy-readable podcasts
  - Order by title here
  - Assign to a less-conflict-prone variable like `@subnav_podcasts`
  - Check the current route to see if one of them should be selected (TODO: what about `/episodes/1234` type paths?)
  - Links should go to the `/podcasts/1234` "show" path
- [x] Make the search box work, but frontend only.  Stimulus!
- [x] Check if there is any content for the subnav - render them only if they exist.

cavis:
Closes #505.

Might be slightly overkill, but I opted to make the dropdown a [lazy loaded frame](https://turbo.hotwired.dev/reference/frames#lazy-loaded-frame).  So it only queries the db after you open it.

After typing in the search box, it makes a POST for a turbo-stream, which replaces the list-of-5-podcasts.


app/helpers/podcasts_helper.rb
```diff
@@ -29,4 +29,11 @@ def itunes_summary(model)
       sanitize_links_only(model.description || "")
     end
   end
+
+  def podcast_settings_active?
+    parts = request.path.split("/").select(&:present?)
+
+    # don't include podcast#show, or any episodes/feeds paths
+    parts[0] == "podcasts" && parts[2] && parts[2] != "episodes" && parts[2] != "feeds"
```
cavis, line 37: The hierarchy of podcast routes doesn't _quite_ line up with when the `Settings` subnav link is active.  Have to filter for:

- `podcasts/123` the "podcast dashboard" thing
- `podcasts/123/episodes` listing episodes
- `podcasts/123/feeds` listing feeds

app/javascript/controllers/search_controller.js
```diff
@@ -0,0 +1,22 @@
...
+    clearTimeout(this.timeout)
+    this.timeout = setTimeout(() => {
+      this.submitForm()
+    }, 200)
+  }
+
+  // NOTE: form.submit() doesn't seem to play well with Turbo at the moment -
+  // so just click the hidden submit button
+  submitForm() {
+    this.submitTarget.click()
```
cavis, line 20: I double checked ... and indeed calling `this.element.submit()` on the `<form>` still doesn't work with Turbo.  It submits a non-turbo-stream-`Accept` request.  Which isn't what I want.

app/policies/application_policy.rb
```diff
@@ -32,4 +32,17 @@ def account_id
   def account_id_was
     nil
   end
+
+  class Scope
+    attr_reader :token, :scope
```
cavis, line 37: Ported from Augury.  But... the pattern in Feeder seems to be calling it a `token` vs a `user`.  So stuck with that here.  (It's the same `Prx::Auth::Token` so doesn't really matter).

app/policies/podcast_policy.rb
```diff
@@ -1,4 +1,8 @@
 class PodcastPolicy < ApplicationPolicy
+  def new?
+    token&.authorized_account_ids(:podcast_create).present?
```
cavis, line 3: The `create?` method requires the Podcast have an account set on it already.  So needed a check for "can you create a podcast for _any_ account ids?".

app/policies/podcast_policy.rb
```diff
@@ -20,4 +24,12 @@ def account_id
   def account_id_was
     resource.account_id_was
   end
+
+  class Scope < Scope
+    def resolve
+      # TODO: this is hacky
+      uris = token.authorized_account_ids(:read_private).map { |id| "/api/v1/accounts/#{id}" }
```
cavis, line 31: I feel like someday (after deprecating CMS)... we should go through and change all these API-account-strings to just integers.

app/views/podcast_switcher/_dropdown.html.erb
```diff
@@ -0,0 +1,61 @@
...
+          <%= f.label :q, t(".search") %>
+        </div>
+        <%= f.submit class: "d-none", data: {search_target: "submit"} %>
+      <% end %>
+    </li>
+
+    <%# lazy-loaded list of podcasts %>
+    <%= turbo_frame_tag "podcast-switcher-list", src: main_app.podcast_switcher_path, loading: "lazy" do %>
+      <li class="d-flex justify-content-center">
+        <div class="spinner-border text-primary m-5" role="status"></div>
```
cavis, line 34: Shows a spinner while the list-of-5 lazy loads.  Saves us querying the DB for a hidden dropdown every page load.

brandonhundt (commented):
@cavis Made one minor styling change for handling when an podcast image isn't present. 

The rest of the functionality is great! Thanks!!! 

Ready for Code Review.

app/views/podcast_switcher/_dropdown.html.erb
```diff
@@ -1,19 +1,18 @@
-<div class="prx-podcast-switcher dropdown" data-controller="focus">
+<div class="prx-podcast-switcher | dropdown" data-controller="focus">
```
brandonhundt, line 1: @rpeterman-gp What's the rationale behind the `|`, just curious?

app/views/podcast_switcher/_dropdown.html.erb
svevang: Screen Shot 2023-02-02 at 10 19 30 AM

Seeing some titles hanging off the edge here.

svevang (approved):

app/policies/application_policy.rb
cavis: It seems they already are?  I've actually been using `current_user` in other places (outside of the Policies).

PR #555, opened 2023-02-01
Make sure that pg_restore has creds

svevang:


ops/bin/_env.sh
```diff
@@ -9,4 +9,6 @@ set -a
 source $DOTENV_PATH
 set +a
 
-
+export PGPASSWORD=$POSTGRES_PASSWORD
+export PGUSER=$POSTGRES_USER
+export PGHOST=$POSTGRES_HOST
```
cavis, line 14: Noticed some missing newlines at the end of files... maybe missing that editor config?

cavis (approved):
👍 looks like 2 PRs with this same content... but looks good to me!

PR #558, opened 2023-02-02
Nav and Subnav Style clean up

brandonhundt:
- Fixes issue when there's a really long podcast name in the podcast switcher and dropdown menu (the menu will expand)
- Cleans up responsive styles quite a bit.

To Review:
- Run the server
- Open code inspector and confirm the nav just looks better at all screensizes.

### DIFF
--- a/app/assets/stylesheets/shared/header.scss
+++ b/app/assets/stylesheets/shared/header.scss
@@ -39,6 +39,7 @@
   }
 
   .dropdown-menu {
+    position: absolute;
     background: $blue-dark;
     margin: 0;
     padding: 0;
@@ -105,14 +106,11 @@
   width: 100%;
   border-bottom: 1px solid $gray-200;
 
-  @media (min-width: 576px) {
+  @media (min-width: 992px) {
+    min-width: 350px;
     width: auto;
     border-right: 1px solid $gray-300;
     border-bottom: none;
-  }
-
-  @media (min-width: 992px) {
-    min-width: 350px;
   }
 
   .dropdown-toggle {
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/shared/header.scss
```diff
@@ -39,6 +39,7 @@
   }
 
   .dropdown-menu {
+    position: absolute;
```
brandonhundt, line 42: Bootstrap was doing a thing where if you were on a small screen, the positioning switched to `static` and that opened up the dropdown within itself. We want it to be positioned absolutely at all times.

app/assets/stylesheets/shared/header.scss
```diff
@@ -105,16 +106,13 @@
   width: 100%;
   border-bottom: 1px solid $gray-200;
 
-  @media (min-width: 576px) {
+  @media (min-width: 992px) {
+    min-width: 350px;
```
brandonhundt, line 110: To avoid weird spacing, we need the podcast switcher to be full width until it hits 992px (or the large screensize). it may seem like a lot, but this works best!

app/views/layouts/_nav.html.erb
```diff
@@ -3,18 +3,18 @@
     <%= link_to t(".dovetail"), main_app.podcasts_path, class: "navbar-brand" %>
 
     <ul class="navbar-nav">
-      <li class="nav-item">
+      <li class="nav-item d-none d-md-block">
```
brandonhundt, line 6: I decided to hide these links on mobile devices. I suspect if you are accessing Feeder on your phone, you will only need to navigate feeder.

NOTE: When we get to implementing this on ID, we may have to do some more work to get these links to play nicer in mobile (but ID won't have a subnav)

cavis (approved):
👍 looks good to me!

PR #561, opened 2023-02-03
Use standard PRX favicon

farski:
My recommendation would be to standardize on the PRX icon for the time being. If we want to develop a meaningful set of icons for each app (Feeder, Augury, etc), or functions (publishing, metrics, etc) down the road, that could be beneficial. But until then, I think just using the vanilla icon gets us in a good spot.


brandonhundt (approved):
+1 LGTM Thanks @farski

PR #563, opened 2023-02-06
Extract a base enclosure url

svevang:
This PR Fixes a couple problems with the urls that the Apple Subscriptions code uses:

1) It removes the `prefix` from the enclosures that the apple api uses.  This decouples the GET from the Dovetail CDN from the typical redirect flow that end users experience. The 'reduced' enclosure url should directly reference the DTR redirect endpoint.
2) It fixes up the feed enclosure to make sure that we're building _authed_ enclosures for the episodes -- the same as what are referenced in the private feed.  This ensures that we are getting the stitched audio from the private feed.  Previously this was getting the enclosure for the default (public) feed.

### DIFF
--- a/app/models/apple/episode.rb
+++ b/app/models/apple/episode.rb
@@ -47,14 +47,6 @@ module Apple
       end
     end
 
-    def self.add_no_imp_param(url)
-      # use the URI class to add a 'noImp' query param
-      url = URI.parse(url)
-      decoded_query = URI.decode_www_form(url.query.to_s) << ["noImp", "1"]
-      url.query = URI.encode_www_form(decoded_query)
-      url.to_s
-    end
-
     def initialize(show:, feeder_episode:, api: nil)
       @show = show
       @feeder_episode = feeder_episode
@@ -69,8 +61,18 @@ module Apple
       feeder_episode.id
     end
 
+    def private_feed
+      show.private_feed
+    end
+
+    def podcast
+      show.podcast
+    end
+
     def enclosure_url
-      self.class.add_no_imp_param(feeder_episode.enclosure_url)
+      url = EnclosureUrlBuilder.new.base_enclosure_url(podcast, feeder_episode, private_feed)
+      url = EnclosureUrlBuilder.mark_no_imp(url)
+      EnclosureUrlBuilder.mark_authorized(url, show.private_feed)
     end
 
     def enclosure_filename
--- a/app/models/enclosure_url_builder.rb
+++ b/app/models/enclosure_url_builder.rb
@@ -76,6 +76,7 @@ class EnclosureUrlBuilder
 
   def enclosure_prefix_url(u, prefix)
     return u if prefix.blank?
+
     pre = Addressable::URI.parse(prefix)
     orig = Addressable::URI.parse(u)
     orig.path = File.join(pre.path, orig.host, orig.path)
@@ -83,8 +84,6 @@ class EnclosureUrlBuilder
     orig.host = pre.host
     orig.to_s
   end
-
-  private
 
   def self.add_query_param(url, key, value)
     url = URI.parse(url)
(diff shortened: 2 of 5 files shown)


app/models/apple/episode.rb
```diff
@@ -69,8 +61,18 @@ def feeder_id
...
+
+    def podcast
+      show.podcast
+    end
+
     def enclosure_url
-      self.class.add_no_imp_param(feeder_episode.enclosure_url)
+      url = EnclosureUrlBuilder.new.base_enclosure_url(podcast, feeder_episode, private_feed)
+      url = EnclosureUrlBuilder.mark_no_imp(url)
+      EnclosureUrlBuilder.mark_authorized(url, show.private_feed)
```
svevang, line 75: This is generating enclosure URLs from the private feed's template so that we get the private (typically adfree) audio. 

Also, using a "pluggable" approach to modify the enclosure.

It's possible that the `noImp` and `auth` params could be part of the template.  But the way we're using them currently is that these params are supra / distinct from the templated enclosure -- more akin to query params.

app/models/enclosure_url_builder.rb
```diff
@@ -2,12 +2,32 @@
 require "addressable/template"
 
 class EnclosureUrlBuilder
+  def self.mark_authorized(enclosure_url, feed)
+    return enclosure_url unless feed.private?
+    raise "Missing tokens for private feed #{feed.id}" if feed.private? && !feed.tokens.any?
```
svevang, line 7: This runtime error works as a contract with the apple publishing to halt the publishing flow if we don't have a way to pull down stitched audio.

test/models/apple/episode_test.rb
```diff
@@ -85,14 +85,4 @@
...
-
-  describe ".add_no_imp_param" do
-    it "should add a noImp query param" do
-      assert_equal "http://example.com?noImp=1", Apple::Episode.add_no_imp_param("http://example.com")
-    end
-
-    it "should preserve existing query params" do
-      assert_equal "http://example.com?foo=bar&noImp=1", Apple::Episode.add_no_imp_param("http://example.com?foo=bar")
-    end
-  end
```
svevang, line 97: Moved these over to the enclosure url builder to illustrate the coupling of apple api related code and enclosure nuance.

app/models/enclosure_url_builder.rb
```diff
@@ -57,11 +76,19 @@ def enclosure_template_url(template, expansions)
...
+
     pre = Addressable::URI.parse(prefix)
     orig = Addressable::URI.parse(u)
     orig.path = File.join(pre.path, orig.host, orig.path)
     orig.scheme = pre.scheme
     orig.host = pre.host
     orig.to_s
   end
+
+  def self.add_query_param(url, key, value)
```
cavis, line 88: Maybe move this up next to the other class methods?

app/models/apple/episode.rb
cavis: I could _maybe_ see adding `auth` to the template, as it's in the API.  But I don't think there's a huge advantage - this works.

`noImp` I generally don't want out in the wild, so makes sense to keep it out of the template.

app/models/enclosure_url_builder.rb
```diff
@@ -2,12 +2,32 @@
...
+    raise "Missing tokens for private feed #{feed.id}" if feed.private? && !feed.tokens.any?
+
+    token = feed.tokens.first.token
+    add_query_param(enclosure_url, "auth", token)
+  end
+
+  # Marks the url as a `noImp`
+  # Used by Dovetail to skip the impression tracking
+  def self.mark_no_imp(enclosure_url)
+    add_query_param(enclosure_url, "noImp", "1")
```
cavis, line 16: This is fine for now - but I feel like at some point, we'll want to not do this.  We track "bot/crawler" traffic for other apps (Spotify, Stitcher, Luminary, Adswizz, etc).  And just mark it as a bot via prx-podagent, and store the `is_duplicate = true` rows in BigQuery.

cavis (approved):
One thought about grouping class methods, but otherwise looks good to me!

PR #565, opened 2023-02-06
feat/ show podcast

### LINKED ISSUE #508: Show Podcast
After #505.  Build out what we can of the "[podcast dashboard](https://www.figma.com/proto/Yewv83Q8btHcaOBWCtGgKs/PRX-Dovetail-with-Navigation?page-id=80%3A2947&node-id=2934%3A80025&viewport=798%2C-3032%2C0.25&scaling=min-zoom&starting-point-node-id=80%3A3420&hide-ui=1)" (`podcasts#show`).

image

- [x] Use `podcasts#show` for this
- [x] Omit the chart for now - condense the other cards
- [x] Use bootstrap grid (and other bootstrap components) as much as possible!
- [ ] Create a shared "podcast subnav" that does the content_for the subnavigation
- [x] Fake any links that don't exist yet

radical-ube:
resolves #508


app/controllers/podcasts_controller.rb
```diff
@@ -9,6 +9,9 @@ def index
   # GET /podcasts/1
   def show
     authorize @podcast
+
+    @recently_published = @podcast.episodes.published.first
+    @next_scheduled = @podcast.episodes.where("released_at > ?", Time.now).order(:released_at).first
```
cavis, line 14: Might be worth making this a scope?  `@podcast.episodes.scheduled.first`.  I imagine we'll need that scope elsewhere.

app/views/podcasts/show.html.erb
```diff
@@ -1,10 +1,105 @@
-<p style="color: green"><%= notice %></p>
+<% if notice %>
+  <p style="color: green"><%= notice %></p>
```
cavis, line 2: You can probably drop this - we've been using toasts.  And I'm not sure one would ever show up on this page?

app/views/podcasts/show.html.erb
```diff
@@ -1,10 +1,105 @@
...
-<%= render @podcast %>
+<div class="container my-4">
+  <div class="d-grid d-sm-flex align-items-center">
+    <%= render @podcast %>
+    <div class="d-grid d-lg-flex gap-2 align-items-center">
+      <%= link_to new_episode_path, class: "btn btn-primary btn-sm" do %>
+        <%= t ".create_episode" %>
+        <span class="material-icons ms-2" aria-hidden="true">mic</span>
+      <% end %>
+      <%= link_to podcast_path(@podcast.id), class: "btn btn-primary btn-sm" do %>
```
cavis, line 13: I think there's a route directly to the plan-episodes tab?

cavis (commented):
Couple minor code suggestions.  But looking good to me!

brandonhundt:
@radical-ube Thanks for the styling updates! LGTM!

app/models/episode.rb
```diff
@@ -57,6 +57,8 @@ class Episode < ApplicationRecord
 
   scope :published_by, ->(offset) { where("published_at IS NOT NULL AND published_at <= ?", Time.now + offset) }
 
+  scope :scheduled, -> { where("released_at > ?", Time.now) }
```
cavis, line 60: Bit of nuance in these labels, but following the Publish/CMS logic...

- An episode is "draft" when `published_at` is nil. But a draft can still have a dropdate set, with a non-nil `released_at`
- An episode is "scheduled" when we copy `released_at -> published_at`.  (And we also verify the episode has all its audio and is "ready" to publish, etc).  It's only scheduled while that date is > Time.now.
- An episode is "published" when we reach that `published_at` timestamp

So here, I think you want to check the `published_at` field instead?  And you'll also need to check that it's not null - in fact, you may want to just copy the `scope :published` and modify that.

app/models/episode.rb
```diff
@@ -54,10 +54,10 @@ class Episode < ApplicationRecord
   after_save :publish_updated, if: ->(e) { e.published_at_previously_changed? }
 
   scope :published, -> { where("published_at IS NOT NULL AND published_at <= now()") }
-
   scope :published_by, ->(offset) { where("published_at IS NOT NULL AND published_at <= ?", Time.now + offset) }
-
-  scope :scheduled, -> { where("released_at > ?", Time.now) }
+  scope :draft, -> { where("published_at IS NULL") }
+  scope :scheduled, -> { where("published_at IS NOT NULL AND published_at > now()") }
+  scope :draft_or_scheduled, -> { draft.or(scheduled) }
```
cavis, line 60: Nice use of scope chaining!

cavis (approved):
👍 looks good to me

PR #567, opened 2023-02-07
Unsaved form

cavis:
Inits a new controller for "unsaved form data", and adds it to the Podcast form.

- Styles elements with the orange-outline
- Tracks changes to the original field value
- Shows/hides a discard button
- Prevents navigating away or refreshing when you have unsaved changes


app/views/podcasts/_form_status.html.erb
```diff
@@ -5,6 +5,7 @@
   <div class="card-footer">
     <% if podcast.persisted? %>
       <p class="status-text"><strong><%= t(".updated_at_hint") %></strong> <br><%= l(podcast.updated_at) %></p>
+      <%= link_to t(".discard"), edit_podcast_path(podcast), class: "btn btn-light me-2 invisible", data: {action: "unsaved#discard", unsaved_target: "discard"} %>
```
brandonhundt, line 8: lets change `btn-light` to an orange button.

Issue #546, opened 2023-01-30
Read-only publishing pass 


svevang:
Set up some testing around a podcast in a subscription channel.

- [x] Get access credentials via podcasts connect
- [x] Run through the publishing routine (in a read only-fashion)

PR #572, opened 2023-02-09
Feat/apple read only polling

svevang:
closes https://github.com/PRX/feeder.prx.org/issues/546

This PR sets up a way to poll the remote apple data in a read-only fashion.  With this, we can run `Apple::Publisher#poll!` to build up some basic stats about episodes and pull down local copies of the podcast containers and related resources.

This also guards against creating new deliveries and uploads for podcast containers with existing podcast audio. Previously we were using the presence of the delivery as a guard. However, with audio uploaded to podcasts connect, there are not deliveries / delivery files associated with the podcast container.

This also reformats the log messages and adds json attributes.

### DIFF
--- a/app/models/apple/podcast_container.rb
+++ b/app/models/apple/podcast_container.rb
@@ -30,7 +30,7 @@ module Apple
       end
     end
 
-    def self.update_podcast_container_state(api, episodes)
+    def self.poll_podcast_container_state(api, episodes)
       results = get_podcast_containers_via_episodes(api, episodes)
 
       join_on_apple_episode_id(episodes, results).each do |(ep, row)|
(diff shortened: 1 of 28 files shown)


app/models/apple/podcast_container.rb
```diff
@@ -186,5 +195,27 @@ def podcast_deliveries_url
...
+      end
+    end
+
+    def missing_podcast_audio?
+      !has_podcast_audio?
+    end
+
+    def needs_delivery?
+      # TODO: Overwriting the podcast audio with another file
+      missing_podcast_audio? && podcast_deliveries.empty?
```
cavis, line 218: Okay, so here's the gatekeeper - it only needs delivery if the audio is missing _and_ we haven't tried deliveries yet.

app/models/apple/podcast_delivery.rb
```diff
@@ -35,16 +40,30 @@ def self.update_podcast_deliveries_state(api, episodes)
...
-      # TODO: Support multiple deliveries per episode
-      podcast_containers = episodes.map(&:podcast_container)
-
-      podcast_containers = podcast_containers.reject do |container|
-        # Don't create deliveries for containers that already have deliveries.
-        # An alternative workflow would be to swap out the existing delivery and
-        # upload different audio.
-        container.podcast_deliveries.present?
-      end
+      podcast_containers = episodes.map do |ep|
```
cavis, line 50: Ooh, we're on ruby 2.7 now ... so how about `.filter_map` ?

cavis (approved):
Seems like a good changeset, accumulated from real world QA!

One minor filter-map suggestion, but 👍 looks good to me.

PR #573, opened 2023-02-09
Feat/gatekeep apple publishing with boolean

svevang:
Sets up an additional boolean to ensure that we don't run the Apple related publishing unless we check that box.


db/migrate/20230209162324_feeds_not_nullable.rb
```diff
@@ -0,0 +1,6 @@
+class FeedsNotNullable < ActiveRecord::Migration[7.0]
+  def change
+    change_column :apple_credentials, :public_feed_id, :bigint, null: false
+    change_column :apple_credentials, :private_feed_id, :bigint, null: false
```
cavis, line 4: I'm guessing just belt-and-suspenders here?  (Looks like the model _should_ validate they exist already).

cavis (approved):
👍

PR #574, opened 2023-02-09
Rename apple_credentials -> apple_configs

svevang:
Fallout from https://github.com/PRX/feeder.prx.org/issues/546 This PR renames the  `Apple::Credentials` object to `Apple::Config`

Since it manages a lot more than credentials. It handles:
- Multi-tenant apple api credentials
- Structural associations needed for the publisher routines
- enable / disable publishing flag (per https://github.com/PRX/feeder.prx.org/pull/573)

So this PR renames the table / classes / tests / factories to match that "config" intent.


cavis (approved):
👍 makes good sense!  It was a little odd that an `AppleCredential` would allow you to have blank credential-fields - it only really requires the public/private feed ids.

PR #578, opened 2023-02-15
Paranoid media files

cavis:
Objectives:

1. Make all our files (media_resources / images) paranoid, so we're not losing track of real files in S3.
2. Ensure we're only using `status == "complete"` images/audio in our RSS and publicly-served-API-files (Dovetail Router)
3. Simplify relations, so we're only accessing plain ol' `has_many`, which can be used via the formbuilder.


app/models/content.rb
```diff
@@ -28,9 +28,6 @@ def update_content_attributes(content)
   end
 
   def replace_resources!
-    episode.with_lock do
-      # delete enclosures in the same episode created before this
-      episode.all_contents.where("position = ? AND created_at < ? AND id != ?", position, created_at, id).destroy_all
-    end
+    Content.where(episode_id: episode_id, position: position).where.not(id: id).touch_all(:replaced_at, :deleted_at)
```
cavis, line 31: Okay here's the fun part...

For all the audio/image files, there's an `after_create` hook.  It finds any other files we're replacing, and marks them both as `deleted_at` and `replaced_at`.  Which allows us to distinguish these "replaced" files vs just deleted ones in the `complete_or_replaced` scope.

app/models/episode.rb
```diff
@@ -133,28 +129,75 @@ def author=(a)
...
+      enclosures.each(&:mark_for_destruction)
+    elsif enc&.replace?(enclosure)
+      enclosures.build(enc.attributes.compact)
+    else
+      enc.update_resource(enclosure)
     end
   end
 
+  def ready_contents
+    contents.complete_or_replaced.group_by(&:position).values.map(&:first)
```
cavis, line 173: Here's another pattern I'm following:

The public RSS (or the API JSON that DTR uses) should always return "ready" files.  Which may be the latest/greatest `contents` here.  Or it might be an older file we just replaced.

app/models/episode.rb
```diff
@@ -133,28 +129,75 @@ def author=(a)
...
+    existing = contents.group_by(&:position)
+
+    files&.each_with_index do |file, idx|
+      position = idx + 1
+      con = Content.build(file, position)
+
+      if !con
+        existing[position]&.each(&:mark_for_destruction)
+      elsif con.replace?(existing[position]&.first)
+        contents.build(con.attributes.compact)
```
cavis, line 186: Trying hard not to take actions that will persist changes right now.  So got rid of any `<<` on the collection, or `destroy_all` calls.

app/models/episode.rb
```diff
@@ -244,15 +287,32 @@ def path
   end
 
   def include_in_feed?
-    !media? || media_ready?
+    (segment_count.nil? && !media?) || media_ready?
```
cavis, line 290: Bit nuanced - but if you set a non-nil segment_count, you need to actually have that many files.

app/models/media_resource.rb
```diff
@@ -4,10 +4,36 @@ class MediaResource < ApplicationRecord
...
   before_validation :initialize_attributes, on: :create
 
+  after_create :replace_resources!
+
+  scope :complete_or_replaced, -> do
+    with_deleted
+      .complete
+      .where("deleted_at IS NULL OR replaced_at IS NOT NULL")
+      .order("created_at DESC")
+  end
```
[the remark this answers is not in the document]
svevang: Does this need to be wrapped in e.g. read committed semantics when saving the form and reading via this scope?  The nested form model save _should_ be in a read-committed transaction, but I'm not sure about the reading side here.

app/models/content.rb
svevang: One thing that's not clear is when validation / transcoding could happen.  Is that via the `complete` enum?  Or maybe we are already downstream of the porter tasks.

app/models/episode.rb
```diff
@@ -263,23 +323,17 @@ def media_status
...
-      !!enclosure
-      # if this episode has contents, ready when each position is ready
-    elsif !all_contents.blank?
-      max_pos = all_contents.map(&:position).max
-      contents.size == max_pos
-      # if this episode has no audio, the media can't be ready, and `media?` will be false
+    if contents.blank?
+      ready_enclosure.present?
+    elsif segment_count.nil?
+      ready_contents.count == contents.maximum(:position)
```
svevang, line 329: Is there a separate validation for 'holes' in the position sequence on either side here?

app/models/episode.rb
cavis: I guess my thinking was... we generally want our APIs and HTML pages to show users the latest state.  What files are in this episode, are they processing, etc.  And also, it made more sense to have the default `has_many` be a normal (doesn't include deleted) relation?  I was trying to get away from the more magical multi-relations we had before, as they were causing issues.

And the exception is for the Feed RSS and DTR - where we want to gatekeep only completed files.

I guess it depends on perspective?

app/models/episode.rb
cavis: No - new contents start out as `created`.  (And we don't allow manually setting the status, via API or worker).

app/models/episode.rb
cavis: I agree, but I think we need to divorce this from CMS first?  I've been struggling to not add validations that will break incoming CMS announcements.  Like this one, which is probably `validate :published_or_scheduled_episodes_match_segment_count`?

(As I type this, I'm now wondering if this change was too heavy-handed, for CMS updates...)

app/models/media_resource.rb
```diff
@@ -23,6 +49,9 @@ def url
     self[:url] ||= media_url
   end
 
+  def replace_resources!
+  end
```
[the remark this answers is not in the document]
cavis: Was trying to fix the `media_resource_test.rb` that directly instantiates this.

app/models/episode.rb
```diff
@@ -133,28 +129,75 @@ def author=(a)
...
+      enc.update_resource(enclosure)
     end
   end
 
+  def ready_contents
+    contents.complete_or_replaced.group_by(&:position).values.map(&:first)
+  end
+
+  def contents=(files)
+    existing = contents.group_by(&:position)
```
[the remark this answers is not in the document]
cavis: Currently, we do have multiple contents per position.  And we just grab the newest-completed one to say "this should be in the feed/API".  (We're a bit inconsistent even on that, though).

The `after_create` hook will start to sort that out, as episodes get re-saved (each Content will delete/replace its predecessors).

I think that will be okay?  This code also handles that case - serving up the newest content in that position.  That was my intent anyways.  Anything newly saved transitions to the new state, and old data continues to work as-is.

Was hoping to avoid "integer versioning" of files if possible.  That just feels like a bit overkill for us.  File-consistency only matters after an episode is published, and even then, users are generally only swapping out files.  If users are doing something more drastic, they should be unpublishing the episode while they fix it.

I may need to think a bit more about about handling old data...

app/models/episode.rb
cavis: Yeah, that's another validation we'll need to add at some point (which could potentially conflict with CMS).  Currently there is no tie/validation between segment_count and your contents.  That all lives on the CMS side.

I may need to think through if we can/should add all those validations sooner ... but that can be a separate issue from this.

app/models/episode.rb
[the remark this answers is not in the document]
cavis: But the `has_many :contents` is effectively append-only, right?  We're just setting deleted_at/replaced_at timestamps on all but the 1st item (by position).

app/models/episode.rb
cavis: Hmm... you're asking what happens if we replace this position with a `Content` that already exists (somehow was saved), and has been processed by Porter?

I'm not sure how that could happen logically (I _think_ Content have to be assigned to an Episode to be created?)... but this code would copy its `original_url` again to a brand new Content (`status: 'created'`), and re-process it.

app/models/episode.rb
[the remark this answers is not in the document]
svevang: Ahh ok,  I thought that guid was keyed on the episode.  I updated the code example.

One other thing to consider here is that if validations push against CMS too much, we could shim in validations that only log errors.

app/models/media_resource.rb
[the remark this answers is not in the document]
svevang: Yep you are right!  The episode save wraps the nested models' saves [in a transaction with read-committed isolation](https://api.rubyonrails.org/v7.0.4.2/classes/ActiveRecord/Callbacks.html#module-ActiveRecord::Callbacks-label-Transactions). Otherwise readers would see the positions shifting around as each Content's after_create hooks run.

app/models/media_resource.rb
```diff
@@ -4,10 +4,36 @@ class MediaResource < ApplicationRecord
...
+  after_create :replace_resources!
+
+  scope :complete_or_replaced, -> do
+    with_deleted
+      .complete
+      .where("deleted_at IS NULL OR replaced_at IS NOT NULL")
+      .order("created_at DESC")
+  end
+
+  def self.build(file, position = nil)
```
svevang, line 22: Does this need to override the rails default build semantics to do what you want?  This could be a helper method `build_for_replacement`. ?

app/models/media_resource.rb
```diff
@@ -4,10 +4,36 @@ class MediaResource < ApplicationRecord
...
+    media =
+      if file&.is_a?(Hash)
+        new(file)
+      elsif file&.is_a?(String)
+        new(original_url: file)
+      else
+        file
+      end
+
+    media.try(:position=, position)
```
svevang, line 32: This code here is tied to the contents setter on the Episode right?  Does this need to be happening for all the subclasses?

svevang (approved):

PR #579, opened 2023-02-17
533 list episodes

### LINKED ISSUE #533: List Episodes

radical-ube:
resolves #533

### DIFF
--- a/app/views/episodes/index.html.erb
+++ b/app/views/episodes/index.html.erb
@@ -7,13 +7,18 @@
     <% @published_episodes.each do |episode| %>
       <%= render episode %>
     <% end %>
+    
+    <%= paginate @published_episodes %>
   </div>
 
   <div class="col">
     <% @scheduled_episodes.each do |episode| %>
       <%= render episode %>
     <% end %>
+    
+    <%= paginate @scheduled_episodes %>
   </div>
 </div>
+
 
 <%= link_to "New episode", new_episode_path %>
--- a/app/views/episodes/index.html.erb
+++ b/app/views/episodes/index.html.erb
@@ -2,13 +2,18 @@
 
 <h1>Episodes</h1>
 
-<div id="episodes">
-  <% @episodes.each do |episode| %>
-    <%= render episode %>
-    <p>
-      <%= link_to "Show this episode", episode %>
-    </p>
-  <% end %>
+<div class="row">
+  <div class="col">
+    <% @published_episodes.each do |episode| %>
+      <%= render episode %>
+    <% end %>
+  </div>
+
+  <div class="col">
+    <% @scheduled_episodes.each do |episode| %>
+      <%= render episode %>
+    <% end %>
+  </div>
 </div>
 
 <%= link_to "New episode", new_episode_path %>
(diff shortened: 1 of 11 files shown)


app/views/episodes/index.html.erb
```diff
@@ -1,14 +1,46 @@
-<p style="color: green"><%= notice %></p>
```
cavis, line 1: This page should probably set a title?  So you can distinguish between searches in your history.  (Maybe also include the podcast title, if you're on that page).

app/views/episodes/index.html.erb
```diff
@@ -1,14 +1,46 @@
...
 
-<div id="episodes">
-  <% @episodes.each do |episode| %>
-    <%= render episode %>
-    <p>
-      <%= link_to "Show this episode", episode %>
-    </p>
-  <% end %>
+<%= form_with(
+      url: episodes_path,
```
cavis, line 4: I think this needs to vary, depending if you're viewing episodes within a podcast or not?

db/schema.rb
```diff
@@ -115,8 +115,8 @@
     t.string "feedburner_orig_link"
     t.string "feedburner_orig_enclosure_link"
     t.boolean "is_perma_link"
-    t.datetime "source_updated_at", precision: nil
     t.string "keyword_xid"
+    t.datetime "source_updated_at", precision: nil
```
cavis, line 119: Extraneous change? (I sometimes just manually undo these, so they're not part of the PR).

cavis (changes requested):
Generally all is working, just a few minor fixes/suggestions.

I'm going to take a TODO to re-think the `search_controller.js`.  Because it's not _quite_ working for both this page and the podcast-selector-dropdown.

app/javascript/controllers/search_controller.js
```diff
@@ -10,13 +10,31 @@ export default class extends Controller {
   submit() {
     clearTimeout(this.timeout)
     this.timeout = setTimeout(() => {
-      this.submitForm()
+      const next = this.encodeForm()
+      if (next !== window.location.href) {
+        this.submitForm()
+        this.setHistory(next)
```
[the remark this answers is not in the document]
radical-ube: i think in the midst of learning how turbo frames worked, we assumed that this might have something to do with an input refresh issue we were running into, but couldn't explain. we can probably leave this out of the PR if it doesn't make sense here.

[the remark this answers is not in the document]
cavis: So normally, turbo-frames do _not_ change your browser history/url.  They only update the interior of that turbo-frame tag on the page.

I think when i wrote this controller for Augury, I wanted the history to change.  So I cobbled together some stuff.  That may have been a bad choice - it feels very javascript-heavy.

I need to do more reading/trying-things, but turbo _streams_ might be a better solution.

app/views/episodes/index.html.erb
```diff
@@ -1,14 +1,64 @@
-<p style="color: green"><%= notice %></p>
+<% content_for :title, t(".title") %>
```
cavis, line 1: It would be nice if this actually had the current podcast's title in it (with a fallback if you're not within a podcast).  But we can tackle later.

cavis (approved):
👍 looks good to me!

PR #580, opened 2023-02-21
Feat/apple publishing depends on asset state

svevang:
Fixes up a race where the episode audio asset state was not complete after the files were uploaded and processed. This was preventing episode publishing.

- Sets up a new polling procedure for checking the episode audio asset linking state
- Guards against waiting for episode assets that will never be processed.


app/models/apple/episode.rb
```diff
@@ -2,14 +2,59 @@
...
+    AUDIO_ASSET_FAILURE = "FAILURE"
+    AUDIO_ASSET_SUCCESS = "SUCCESS"
+
+    def self.wait_for_asset_state(api, eps)
+      wait_for(eps) do |remaining_eps|
+        Rails.logger.info("Probing for episode audio asset state")
+        unwrapped = get_episodes(api, remaining_eps)
+
+        remote_ep_by_id = unwrapped.map { |row| [row["request_metadata"]["guid"], row] }.to_h
+        remaining_eps.each { |ep| ep.api_response = remote_ep_by_id[ep.guid] }
```
svevang, line 18: This is the new polling procedure on the episode.

It was a bit problematic to set this up given how we were fetching apple episode state. Originally we were querying the _show_ to get the episode json.  This PR moves the episode state from the Apple::Show model into the Apple::Episode. But in order to do that I had to "disconnect" the apple episode json from the show's accessor methods.

Now the Apple::Episode has an `api_response` attribute (exactly the same as PodcastContainer, PodcastDelivery, and PodcastDeliveryFile) -- though it is not persisted to the database as the Apple::Episode is not an active record model.

app/models/apple/episode.rb
```diff
@@ -2,14 +2,59 @@
...
+        rem =
+          remaining_eps.filter do |ep|
+            if ep.waiting_for_asset_state?
+              Rails.logger.info("Waiting for audio asset state", {episode_id: ep.feeder_id,
+                                                                      delivery_file_count: ep.podcast_delivery_files.count,
+                                                                      delivery_files_processed_errors: ep.podcast_delivery_files.all?(&:processed_errors?),
+                                                                      delivery_files_processed: ep.podcast_delivery_files.all?(&:processed?),
+                                                                      delivery_files_delivered: ep.podcast_delivery_files.all?(&:delivered?),
+                                                                      asset_state: ep.audio_asset_state,
+                                                                      has_podcast_audio: ep&.podcast_container&.has_podcast_audio?})
```
svevang, line 29: If we do end up waiting an an episode, make sure to log out _why_.

app/models/apple/episode.rb
```diff
@@ -2,14 +2,59 @@
...
+    def self.get_episodes_via_show(api, show_id)
+      bridge_params = Apple::Show.get_episodes_json(api, show_id).map do |ep_json|
+        id = ep_json["id"]
+        guid = ep_json["attributes"]["guid"]
+        Episode.get_episode_bridge_params(api, id, guid)
+      end
+
+      api.bridge_remote_and_retry!("getEpisodes", bridge_params)
+    end
+
```
svevang, line 57: Now the episodes have the same pattern as the PodcastContainer/PodcastDelivery/PodcastDeliveryFile where you can see the pattern as `.get_<children>_via_<parent>`.

1) Use the 'parent' model as a means to get the index view (e.g. `show/<show_id>/relationships/episodes`)
2) Then use singular api endpoint (e.g. `episodes/<ep_id>`) to get a standardized view of the resource data

app/models/apple/episode.rb
```diff
@@ -35,25 +80,26 @@ def self.publish(api, episodes)
...
-    def initialize(show:, feeder_episode:, api: nil)
+    def initialize(show:, feeder_episode:, api:, api_response: nil)
       @show = show
       @feeder_episode = feeder_episode
+      @api_response = api_response
       @api = api || Apple::Api.from_env
     end
 
-    def item_guid
+    def guid
```
svevang, line 102: Trying to clean this up as we're presenting a unified `guid` to the join against apple's episode guid attributes (which is parsed from the RSS potentially)

app/models/apple/episode.rb
```diff
@@ -199,6 +245,12 @@ def apple?
       feeder_episode.apple?
     end
 
+    def apple_json
+      return nil unless api_response.present?
+
+      apple_data
+    end
```
svevang, line 252: Shoehorned in the existing api data accessors into the `Apple::ApiResponse` call structure -- part of integrating the `api_response` attribute.

cavis (approved):
👍 looks good to me!

PR #581, opened 2023-02-21
Update standardrb gem

svevang:
Needed for compatibility with the new [testdouble vscode extension](https://github.com/testdouble/vscode-standard-ruby).

This bumps the standard ruby gem to the latest version 1.24.3.


cavis (approved):
👍

PR #583, opened 2023-02-23
Audio file uploads

### LINKED ISSUE #568: Audio/Image uploader (to S3)
Implement audio and image uploaders throughout the app.

cavis:
Fixes #568.

image

Changes:

- [x] Moves audio/images into panels in main form body, per latest designs
- [x] Signed uploads to temporary S3 location
- [x] Audio state changes / playback UI

Fallout:

- [ ] Developing this locally is a bit of a pain
- [ ] Still need to work through the UI/error-messaging for audio validation errors.  Or removing audio from a published episode.
- [ ] Probably still needs some UI polish
- [ ] Need to setup the temporary uploads bucket in CFN


app/models/episode.rb
```diff
@@ -378,6 +385,22 @@ def feeder_cdn_host
     ENV["FEEDER_CDN_HOST"]
   end
 
+  def segment_range
+    1..segment_count.to_i
+  end
+
+  def build_contents
+    segment_range.map do |p|
+      contents.find { |c| c.position == p } || contents.build(position: p)
```
cavis, line 394: Used for the frontend ... return just the segment_range filled in with new/unsaved contents.

The other half is `destroy_out_of_range_contents` below - after the episode is saved, deletes content that we're no longer using.

app/models/tasks/copy_media_task.rb
```diff
@@ -1,15 +1,18 @@
 class Tasks::CopyMediaTask < ::Task
   before_save do
     if media_resource && status_changed?
-      media_resource.update!(status: status)
```
cavis, line 4: I was hitting a bug, where the "processing" view would switch to "complete" too early.  Before the audio metadata (duration, bitrate, etc) was there.

Turns out - this method was making 2 different `update!` calls.  Refactored a bit so it only makes 1.  Otherwise equivalent.

app/views/episodes/_form_audio.html.erb
```diff
@@ -1,10 +1,47 @@
...
+            <button type="button" class="prx-btn-help" data-bs-toggle="popover" data-bs-content="<%= t(".help.segment_count") %>">
+              <span class="material-icons">help</span>
+            </button>
+          </div>
 
-  <button type="button" class="prx-btn-help" data-bs-toggle="popover" data-bs-content="<%= t(".help.segment_count") %>"><span class="material-icons">help</span></button>
+          <%= form.fields_for :contents, episode.build_contents, include_id: false do |fields| %>
+            <% content = fields.object %>
+
+            <%= turbo_frame_tag "episode-form-audio-#{content.position}" do %>
```
cavis, line 27: Okay, and this partial has all the turbo-magic in it:

1. An outer parent turbo-frame, which the `segment_count` field updates (which adds/removes contents)
2. Many inner content turbo-frames, for each `position` in this segment_count.  Uploading/deleting/replacing the content in that position only impacts that inner turbo-frame.

app/controllers/episodes_controller.rb
```diff
@@ -19,28 +19,34 @@ def index
...
   def show
-    authorize @episode
+    redirect_to edit_episode_url(@episode)
   end
 
   # GET /episodes/new
   def new
     @episode = Episode.new(episode_params)
+    @episode.strict_validations = true
+    @episode.valid? if turbo_frame_request?
```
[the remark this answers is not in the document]
a-maci29: Wanna make sure @radical-ube and I are understanding this right:
So when the turbo frame receives a change to update the DOM, we want to check if the episode/audio is valid or not, and if it isn't, display the error.

app/controllers/uploads_controller.rb
```diff
@@ -0,0 +1,28 @@
+class UploadsController < ApplicationController
+  # GET /uploads/signature
+  def signature
+    date_stamp = Date.strptime(params[:datetime], "%Y%m%dT%H%M%SZ").strftime("%Y%m%d")
+
```
a-maci29, line 5: We think that OpenSSL is encrypting these keys for AWS. True? False? @radical-ube

app/helpers/feeder_form_builder.rb
```diff
@@ -140,7 +140,10 @@ def add_changed(method, opts)
 
     if object.present?
       if object.respond_to?("#{method}_changed?")
-        add_class(opts, CHANGED_CLASS) if object.public_send("#{method}_changed?")
+        # ignore nils being set to blanks by text fields
+        if !object.respond_to?("#{method}_was") || (object.public_send("#{method}_was").nil? && object.public_send(method) == "")
```
a-maci29, line 144: @radical-ube We're not sure what line 144 checks for, when on line 145-6`'s `object.public_send("#{method}_changed?")` seems to also be checking for a change in attribute.

app/helpers/uploads_helper.rb
```diff
@@ -0,0 +1,69 @@
+module UploadsHelper
+  def uploads_meta_tags
+    [
+      tag(:meta, name: :upload_bucket_name, content: ENV["UPLOAD_BUCKET_NAME"]),
+      tag(:meta, name: :upload_bucket_prefix, content: uploads_prefix),
+      tag(:meta, name: :upload_s3_endpoint_host, content: ENV["UPLOAD_S3_ENDPOINT_HOST"]),
+      tag(:meta, name: :upload_signing_service_key_id, content: uploads_key_id),
+      tag(:meta, name: :upload_signing_service_url, content: uploads_service_url)
+    ].join("\n    ").html_safe
+  end
```
a-maci29, line 10: What are the `meta_tags`? Is this purely an AWS thing?

app/javascript/controllers/audio_player_controller.js
```diff
@@ -0,0 +1,104 @@
...
+
+export default class extends Controller {
+  static targets = ["audio", "duration", "progress", "progressBar"]
+
+  connect() {
+    this.playing = false
+    this.seeking = false
+    this.originalDuration = this.durationTarget.innerHTML
+    this.bindMouseMove = this.mouseMove.bind(this)
+    this.bindMouseUp = this.mouseUp.bind(this)
```
a-maci29, line 12: Which `this` do each `this` refer to?

app/javascript/controllers/audio_player_controller.js
```diff
@@ -0,0 +1,104 @@
...
+
+  disconnect() {
+    window.removeEventListener("mousemove", this.bindMouseMove)
+    window.removeEventListener("mouseup", this.bindMouseUp)
+  }
+
+  async play() {
+    this.playing = true
+    if (!this.audioTarget.duration) {
+      this.setProgress(0, 1) // initial render
```
a-maci29, line 23: Are the `0` and `1` acting as boolean values for time values?

a-maci29: We might have figured it out:
Is it measuring progress in percentage? `0` is the beginning, and `1` represents 100%, meaning the end?

app/javascript/controllers/audio_upload_controller.js
```diff
@@ -0,0 +1,46 @@
...
+  // skip some updates/deletes for segment_count changes
+  beforeFrameRender(event) {
+    const childrenOnly = true
+    const onBeforeElUpdated = this.skipSegmentChanges
+    const onBeforeNodeDiscarded = this.skipSegmentChanges
+
+    event.detail.render = (currentElement, newElement) => {
+      if (this.isParentTurboFrame(event.target)) {
+        if (this.isInvalidSegmentCount(event.detail.newFrame)) {
+          morphdom(currentElement, newElement, { childrenOnly, onBeforeElUpdated, onBeforeNodeDiscarded })
```
a-maci29, line 23: We may need a rundown of some `morphdom` basics.

app/javascript/controllers/audio_upload_controller.js
```diff
@@ -0,0 +1,46 @@
...
+        }
+      } else {
+        morphdom(currentElement, newElement, { childrenOnly })
+      }
+    }
+  }
+
+  // skip changes to segment turbo-frames (they update themselves)
+  skipSegmentChanges(el) {
+    return (el.id || "").match(/episode-form-audio-[0-9]+/) ? false : true
```
a-maci29, line 35: Where is the `el` being passed in from? The only other two usages of `skipSegmentChanges` don't pass in an element from what we see, so we're not sure how it's evaluating.

app/javascript/controllers/unsaved_controller.js
```diff
@@ -98,4 +101,22 @@ export default class extends Controller {
       }
     }
   }
+
+  hasChangedFields() {
+    // scan for changes in form (ignoring slim select, which takes a bit to update)
+    const changed = this.element.querySelectorAll(".is-changed:not(.ss-main):not([style='display:none'])")
```
a-maci29, line 107: This will select all elements with the `is-changed` class that does not also have the `.ss-main` class and is not hidden with `display:none`
Then, we loop through all of those, and if any are not `slimselect` or `offsetParent`, then it is true that there changed fields.

app/views/episodes/_form.html.erb
```diff
@@ -1,22 +1,23 @@
-<%= form_with(model: episode, html: {autocomplete: "off"}, data: {controller: "unsaved", unsaved_changed_value: episode.changed?, unsaved_confirm_value: t(".confirm")}) do |form| %>
+<% model = episode.new_record? ? [episode.podcast, episode] : episode %>
```
a-maci29, line 1: What does the `[episode.podcast, episode]` array accomplish in the form?

Also curious why it's defined here instead of inside a helper or a controller.

app/controllers/episodes_controller.rb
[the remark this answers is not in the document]
cavis: So the only reason I need to call that ... is to get the validation errors to show up on this field:

image

When you change that field to blank, Turbo makes a request for:

```
GET http://localhost:3002/episodes/<guid>/edit?episode%5Bsegment_count%5D=
```

That's the `episodes#show` route.  (Or `episodes#new` - if you're creating a new episode).

You usually don't call `.valid?` anywhere in the `def show` or `def new` methods.  Since you don't need/want a bunch of validation messages to show up in the form before the user has even entered any data.  But here, for the turbo-request only ... I did want that validation error to show up.

app/helpers/feeder_form_builder.rb
cavis: I was hitting an issue where, if you went to create a new episode, left the title blank, and hit save...

You'd get the angry red "title can't be blank" message on the page.

But you'd also get a ton of orange-outlined changed fields.  Because the form just changed a bunch of attributes (`clean_title`, for instance) from `nil` to `""`.

Bit cosmetic, but I'm leaving off the orange-outline `is-changed` class if all of:

- `object.clean_title_changed? == true`
- `object.clean_title == ""`
- `object.clean_title_was == nil`

cavis: Oh wait... I see what you're saying.

Yeah, I botched this one.  Will clean this up.

app/helpers/uploads_helper.rb
cavis: These insert tags into the HTML `<head>`, from the Application layout.  The page source will have a bunch of things like:

```html
<meta name="upload_bucket_prefix" content="uploads/feeder" />
```

Honestly, I can't remember why I did that, instead of using [Stimulus values](https://stimulus.hotwired.dev/reference/values) to pass these to the controller.

Maybe just trying to save some characters, vs typing out a bunch of `data_upload_bucket_prefix_value=`?  Whatever past-ryan thought, i'm not feeling opinionated on it.  Would it be clearer if I moved them?

app/javascript/controllers/audio_player_controller.js
cavis: So that method is defined as `setProgress(currentTime, totalDuration)`.  At this point, I don't actually know the duration of the audio file (the browser hasn't loaded it yet).  So I'm just arbitrarily passing 0 and 1, to set it back at 0% playback.

app/javascript/controllers/audio_player_controller.js
cavis: `this` is always the instance of the controller we're in.  I'm creating a new version of that method, where within it, `this` still refers to the controller.  Same as right here.

If I didn't do that, and called:

```js
window.addEventListener("mousemove", this.mouseMove)
```

Then inside mouse move:

```js
mouseMove(event) {
  // an error would happen on the next line, because "this" is no longer a stimulus controller
  this.seekProgressBar(event.x)
}
```

app/javascript/controllers/audio_upload_controller.js
cavis: I'm passing this method (`this.skipSegmentChanges`) in as the `onBeforeElUpdated` and `onBeforeNodeDiscarded` options to morphdom.

If you search for those options in the [morphdom docs](https://www.npmjs.com/package/morphdom) ... you'll see they both get set to functions.  And the functions take in either 1 or 2 elements ... since I don't care about the 2nd element, I just have the 1st one in the method definition here.

app/javascript/controllers/unsaved_controller.js
cavis: Ooh, I left several unnecessary things in this call.  Should just be `this.element.querySelectorAll(".is-changed")`.  Will fix.

But you're right - i loop through to check them all.

app/views/episodes/_form.html.erb
cavis: This is varying the form url/method, depending if this is a new episode (`POST /podcasts/123/episodes`) or an existing one (`PATCH /episodes/abcd1234`).

I added it to a helper, but that seemed like unnecessary obfuscation.  And the `model:` option was the easiest way to make it work in both create/update.

app/views/episodes/_form_audio.html.erb
```diff
@@ -1,10 +1,47 @@
...
+              <% if content.new_record? || content.marked_for_destruction? %>
```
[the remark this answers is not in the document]
cavis: If you're uploading a new segment (Content), or replacing an existing (you just marked-for-destruction via Turbo) ... we want to show the file-uploader in that partial.

app/controllers/uploads_controller.rb
cavis: Can of worms ... but this is implementing the [AWS v4 signature algorithm](https://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-authenticating-requests.html).

And because the javascript upload-file-to-S3 library we use ([EvaporateJS](https://github.com/TTLabs/EvaporateJS)) is real picky, I couldn't use the AWS SDK to do this.  I had to manually port some of their [example signing code](https://github.com/TTLabs/EvaporateJS/blob/master/example/signing_example_awsv4_controller.rb) - though not exactly.

Honestly, can't remember how/why Ryan landed on this.  But it works!

app/helpers/uploads_helper.rb
a-maci29: Also not too opinionated. It's understandable as it is and why it lives here.

app/javascript/controllers/upload_controller.js
```diff
@@ -0,0 +1,206 @@
+import { Controller } from "@hotwired/stimulus"
+import Evaporate from "evaporate"
+import SparkMD5 from "spark-md5"
+import sha256 from "sha256"
+
+export default class extends Controller {
```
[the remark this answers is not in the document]
a-maci29: What is EvaporateJS and SparkMD5 that are being imported?

config/initializers/shoryuken.rb
```diff
@@ -11,4 +11,5 @@
 
 Shoryuken.configure_server do |_config|
   Shoryuken::Logging.logger = Rails.logger
+  Shoryuken::Logging.logger.level = :info
```
a-maci29, line 14: What does Shoryuken do?

test/controllers/episodes_controller_test.rb
```diff
@@ -59,8 +53,11 @@ class EpisodesControllerTest < ActionDispatch::IntegrationTest
   end
 
   test "should update episode" do
-    patch episode_url(episode), params: {episode: params}
-    assert_redirected_to edit_episode_url(episode)
+    Episode.stub_any_instance(:copy_media, true) do
```
a-maci29, line 56: Where does `:copy_media` come from?

a-maci29 (approved):
@radical-ube and I went through everything to the best of our ability and found no errors or discrepancies. We've played around with it locally and ran tests, everything is working as expected.

config/initializers/shoryuken.rb
cavis: [Shoryuken](https://github.com/ruby-shoryuken/shoryuken) is the gem that hooks our workers up to the SQS queues.  So any messages on those queues get picked up and processed.

test/controllers/episodes_controller_test.rb
cavis: A bunch of our models (Podcast, Episode, Feed, Images, MediaResources) have a `def copy_media` method on them.  Which triggers Porter to inspect/copy S3 files to their destinations.

I had to mock it out in the test now - since I added `@episode.copy_media` to the controller.  Otherwise the test suite complains that some code made an un-expected outgoing HTTP request during the test run.

PR #586, opened 2023-02-28
Fix nil check on episode card date in podcast#show

radical-ube:


app/views/podcasts/show.html.erb
```diff
@@ -42,7 +42,13 @@
         <div class="card-body episode-card h-auto border-start-0">
           <% if @next_scheduled.present? %>
             <div class="episode-card-info">
-              <p class="episode-card-date"><%= @next_scheduled.released_at.to_date.strftime("%m/%d") %></p>
+              <p class="episode-card-date">
+                <% if @next_scheduled.published_at.present? %>
```
cavis, line 46: I think we should just make this a method in Episode, because we'll need it many places.  `def published_or_released_at` or something?

app/views/podcasts/show.html.erb
```diff
@@ -42,7 +42,13 @@
...
+                  <%= @next_scheduled.published_at.to_date.strftime("%m/%d") %>
```
cavis, line 47: Now that I'm thinking about it - this should use the `l()` localization method. 

```ruby
l(@next_scheduled.published_or_released_at, format: :short)
```

(May need to add/change that format to en.yml too)

cavis (changes requested):

cavis (approved):
👍 looks good!

PR #590, opened 2023-03-04
576 list podcasts

a-maci29:


app/controllers/podcasts_controller.rb
a-maci29, line 1: [It was decided](https://prx.slack.com/archives/G0FKLK2Q5/p1678225789907889?thread_ts=1678221265.495749&cid=G0FKLK2Q5) to leave off the filter from the original design for now

app/views/podcasts/index.html.erb
```diff
@@ -1,14 +1,40 @@
...
-<%= link_to "New podcast", new_podcast_path %>
+    <% end %>
+    </div>
+</section>
+
+<%= paginate @podcasts, window: 2, outer_window: 1 %>
+<div class="text-end">
+</div>
+  <%= page_entries_info @podcasts %>
+  </div>
```
kookster, line 39: indents look kinda weird here, is that right?

app/views/podcasts/index.html.erb
```diff
@@ -1,14 +1,40 @@
...
 
-<h1>Podcasts</h1>
+      <a data-bs-toggle="dropdown">
+      <button class="btn btn-outline-light btn-sm dropdown-toggle">
+        Sort <span class="material-icons" style="color: green" aria-hidden="true" aria-hidden="true">sort</span>
+      </button>
+        <ul class="dropdown-menu"><a class="dropdown-item active" href="<%= podcasts_path(sort: "Recent") %>">Recent Activity</a></li>
+          <li><%= link_to "# of Episodes", podcasts_path(sort: "episode_count"), class: "dropdown-item" %></li>
+          <li><%= link_to "A-Z", podcasts_path(sort: "A-Z"), class: "dropdown-item" %></li>
+          <li><%= link_to "Z-A", podcasts_path(sort: "Z-A"), class: "dropdown-item" %></li>
```
kookster, line 16: Could also get these values from the hash in the controller, instead of duplicating them

app/views/podcasts/index.html.erb
```diff
@@ -1,14 +1,40 @@
+<% content_for :title, "My Podcasts" %>
+
 <p style="color: green"><%= notice %></p>
+<div class="row mt-4 mb-4">
+  <div class="dashboard header">
+    <div class="col d-flex align-items-center ">
+      <h1 class="text-black fw-bold me-4 mb-0">My Podcasts</h1>
```
kookster, line 7: Maybe a design question, but "My" may not be true - this is podcasts they can access, not some kind of implied ownership

app/controllers/podcasts_controller.rb
kookster: makes sense to me!

kookster (changes requested):
I'd take a look at the erb formatting, seems like those div indents are off?
Maybe thinking about refactoring the ordering options to draw from the same data, but that's not critical.

app/views/podcasts/index.html.erb
a-maci29: I deferred to the design, but I can change it if you like

app/views/podcasts/index.html.erb
a-maci29: Yep, definitely needs some fixes; I'm on it.

cavis (approved):
:+1: This looks/works good to me!

I kind of want to see a test for this... but our integration/systems tests in this repo aren't in good shape right now.  So that may have to wait.  So onwards!

For fallout-tickets, we should maybe open:

1. A ticket to i18n-ify these pages (moving all english strings and timestamp/date formatting into `en.yml`)
2. A ticket to support the (optional) `?per` query param
3. A ticket to fixup N+1 queries.  When I load this page locally, I see ~5 SQL queries per podcast displayed on the page.  That's fine to start, but we should optimize that so it doesn't need to run a bunch of similar-looking SQL for each podcast.

brandonhundt:
Hey @a-maci29 great work! Its nice to see the sort function working!  Now time for one very picky bit of feedback 🎁!!!

The class `.card-body` is applying a padding to the left and right of each item causing an extra gap. (Note in the screenshot). Since Bootstrap is applying a gutter to the flex-item, we don't need the extra padding. Feel free to zero it out!

Screenshot 2023-03-21 at 5 00 53 PM

Thank you!!!

a-maci29:
I pushed up a fix! Go ahead and check it out. :)

PR #591, opened 2023-03-07
Feat/plan podcast episodes

### LINKED ISSUE #532: Plan Podcast Episodes
After #505.  Build the Podcast -> Settings -> [Episode Planner](https://www.figma.com/proto/Yewv83Q8btHcaOBWCtGgKs/PRX-Dovetail-with-Navigation?page-id=80%3A2947&node-id=2461%3A27199&viewport=798%2C-3032%2C0.25&scaling=min-zoom&starting-point-node-id=80%3A3420&hide-ui=1)

image

- [x] Setup routes/tab, using `episode_planner#index` and `/podcasts/123/planner`
- [x] Use a bootstrap grid and other bootstrap-y things for the layout.
- [x] This is probably going to take a lot of Turbo, and interesting param management.
- [x] The "include in feeds" thing doesn't exist yet, probably.

radical-ube:
resolves #532 

This is currently an MVP for the Podcast Planner tool, recreated based on [Publish's current planner tool.](https://publish.staging.prx.tech/series/37560/plan)


app/helpers/feeder_form_builder.rb
```diff
@@ -113,6 +114,11 @@ def search_select(method, choices, params, html_options = {})
     select(method, choices, {include_blank: true, selected: selected}, add_search_action(html_options))
   end
 
+  def toggle_field_select(method, choices, params, html_options = {})
```
radical-ube, line 117: in hindsight, I'm not sure if a new kind of select was totally necessary to create. i can't remember if/why i had issues adding this toggle action to a normal `select` field.

app/helpers/podcast_planner_helper.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
```
radical-ube, line 2: all text in here still needs to go into `en.yml`.

app/javascript/controllers/flatpickr_controller.js
```diff
@@ -19,6 +19,8 @@ export default class extends Controller {
       parseDate: (str) => this.parseDate(str),
       formatDate: (date) => this.formatDate(date),
       onValueUpdate: () => this.validate(),
+      minDate: this.element.min,
+      maxDate: this.element.max,
```
radical-ube, line 23: added to capture any `min` or `max` attributes on `date_field`s.

app/javascript/controllers/podcast_planner_controller.js
```diff
@@ -0,0 +1,9 @@
+import { Controller } from "@hotwired/stimulus"
+
+export default class extends Controller {
```
radical-ube, line 3: i think i modeled this after the `allocations` controller in Augury, but obviously very pared down and simplified to what i needed it for, which is to update the dates "preview" in the view.

app/javascript/controllers/toggle_field_controller.js
```diff
@@ -17,4 +17,16 @@ export default class extends Controller {
       this.checkTarget.parentElement.classList.remove("d-none")
     }
   }
+
```
radical-ube, line 20: added this in this already existing controller; i wasn't sure it needed to be in a completely different controller, and it shares a little bit of conceptual similarities to what is already here, which is why i put it in here.

app/models/podcast_planner.rb
```diff
@@ -0,0 +1,152 @@
+class PodcastPlanner
```
radical-ube, line 1: this is a new model that i'm using to package all the date calculation and draft generation logic, in order to abstract it away from the controller.

app/models/podcast_planner.rb
```diff
@@ -0,0 +1,152 @@
...
+
+  def periodic?
+    @week_condition == "periodic"
+  end
+
+  def monthly?
+    @week_condition == "monthly"
+  end
+
+  def ready_to_select_weeks?
```
radical-ube, line 44: this method and the next two below are what i'm using as guards when generating dates. i'm wondering if i should just put these into actual validations, or if it makes sense to place the logic here? it feels like i won't need it other than as a guard, which feels like a hint to put it into a validation, but i'm not sure on which(or either/neither) is best practice.

app/models/podcast_planner.rb
```diff
@@ -0,0 +1,152 @@
...
+  def ready_to_generate_dates?
+    @selected_days.present? &&
+      ready_to_select_weeks? &&
+      ready_to_select_date_range?
+  end
+
+  def generate_dates!
+    return unless ready_to_generate_dates?
+
+    if date_range_ends_by_episodes?
```
radical-ube, line 73: i also have a bunch of these if/else conditionals because of the different decision trees that could happen from the date generator, and wondered if there was a better way to handle them. at the very least i've tried to abstract and name it as much as possible to provide clarity on where the logic is going.

app/models/podcast_planner.rb
```diff
@@ -0,0 +1,152 @@
...
+      generate_dates_by_remaining_episodes
+    elsif date_range_ends_by_end_date?
+      generate_dates_by_end_date
+    end
+  end
+
+  def clear_dates!
```
radical-ube, line 80: just realized i made this but don't use it anywhere. and i've also been reading up on attribute methods in the [rails docs](https://guides.rubyonrails.org/active_model_basics.html) and i could probably do this in a different way, if i needed it later.

app/models/podcast_planner.rb
```diff
@@ -0,0 +1,152 @@
...
+      dates_to_add.push(current_day) if @selected_days.include?(current_day.wday)
+      current_day += 1.day
+    end
+    dates_to_add
+  end
+
+  def generate_dates_by_remaining_episodes
+    current_day = @start_date
+
+    while episodes_remain?
```
radical-ube, line 96: the logic inside of this loop is the exact same as in the method below, which means i should be able to extract it. (something like, `add_dates_and_advance_date_pointer` or something?). the conditional bothers me, because the two branches are just different enough to require two different methods, and i'm wondering if adding so many different methods pollutes the logic clarity (or does it help)? is there another way to do this?

app/models/podcast_planner.rb
```diff
@@ -0,0 +1,152 @@
...
+    @dates.present? &&
+      @publish_time.present? &&
+      @segment_count.present?
+  end
+
+  def generate_drafts!
+    return unless ready_to_generate_dates?
+    return unless ready_to_generate_drafts?
+
+    @dates.each do |date|
```
radical-ube, line 143: the logic here is that the differentiating aspect of these drafts is the date, so that's why i loop through the dates, and then just return general objects for the different attributes i'll want when actually making the db query in the controller to bulk insert.

app/views/podcast_planner/_form_date_calculator.html.erb
```diff
@@ -0,0 +1,83 @@
...
+    </div>
+  </div>
+
+  <div class="mb-4" data-controller="toggle-field">
+    <h3 class="fw-bold">Week selection</h3>
+    <div class="mb-4">
+      <p>Which weeks of each month do you plan to publish? Select all that apply.</p>
+      <div class="col-lg-12" data-toggle-field-target="field" field="periodic">
+        <div class="form-floating">
+          <%= form.select :period, periodic_weeks_options, {}, data: {action: "podcast-planner#preview"} %>
```
radical-ube, line 26: so i've added all of these `onChange` actions to all the inputs in this partial, so that the preview updates with changes on any of the inputs.

app/views/podcast_planner/_form_date_calculator.html.erb
```diff
@@ -0,0 +1,83 @@
...
+    </div>
+  </div>
+
+  <div class="mb-4" data-controller="toggle-field">
+    <h3 class="fw-bold">Date Range selection</h3>
+    <div class="row mb-4 g-3">
+      <p>How long should episodes recur? Give a date range with a start date, and choose how to calculate the end date.</p>
+      <div class="col-lg-6">
+        <div class="form-floating">
+          <%= form.date_field :start_date, min: DateTime.tomorrow, max: DateTime.tomorrow + 6.months, value: DateTime.tomorrow, data: {action: "podcast-planner#preview"} %>
```
radical-ube, line 55: min and max feels arbitrary, but following what's in Publish, i figured a 6 month time span would be good enough?

app/views/podcast_planner/_form_date_calculator.html.erb
```diff
@@ -0,0 +1,83 @@
...
+          <%= form.label :start_date %>
+        </div>
+      </div>
+      <div class="col-lg-6" data-toggle-field-target="field" field="episodes">
+        <div class="form-floating">
+          <%= form.number_field :number_of_episodes, {min: 1, max: 100, step: 1, value: 10, data: {action: "podcast-planner#preview"}} %>
```
radical-ube, line 61: same question here with min/max on episode count. i'm not sure there was a max limit in Publish. but probably don't want people bulk creating 1000s of episodes?

and then the default was 10, which i've recreated here.

app/views/podcast_planner/_form_main.html.erb
```diff
@@ -0,0 +1,19 @@
+<%= form_with(
```
radical-ube, line 1: so using Turbo, the function of this controller is just to update the preview, since on submit all it does is `get` back to `index`. this is the "main" middle section of the view, and the target is the "side" section with the preview.

app/views/podcast_planner/_form_options.html.erb
```diff
@@ -0,0 +1,33 @@
+<section>
```
radical-ube, line 1: these particular options have no impact on date generation so i separated them out into their own partial for conceptual clarity. i even wonder if they can just go into the side panel and just be selected in the preview?

app/views/podcast_planner/_form_options.html.erb
```diff
@@ -0,0 +1,33 @@
...
+    <h2 class="fw-bold">Audio Segments</h2>
+    <p>
+      Make sure you choose a template associated with your series
+      Podcast Info.
+      This will help us forecast ad inventory available in each episode.
+    </p>
+
+    <div class="col-lg-4">
+      <div class="form-floating">
+        <%= form.number_field :segment_count, {min: 1, max: 10, step: 1, value: 2, data: {action: "podcast-planner#preview"}} %>
```
radical-ube, line 28: do we have a min/max audio segment count? i'm using 1 and 10 for now, but i feel like most don't go above 4? and  the highest "outlier" i've seen was maybe 6 or 7?

app/views/podcast_planner/_form_side.html.erb
```diff
@@ -0,0 +1,45 @@
+<div class="card">
+  <div class="card-header-primary">
+    <h5 class="card-title">Schedule Preview</h5>
+  </div>
+  <%= turbo_frame_tag "preview", target: "_top" do %>
```
radical-ube, line 5: this is the preview that gets updated by Turbo when the date selections change.

app/views/podcast_planner/_form_side.html.erb
```diff
@@ -0,0 +1,45 @@
...
+    <% if @planner.dates.present? %>
+      <%= form_with(
```
radical-ube, line 7: this form is what sends a `put` query to hit the `update` action to actually generate the episodes.

app/views/podcast_planner/_form_side.html.erb
```diff
@@ -0,0 +1,45 @@
...
+            url: podcast_planner_path(@podcast),
+            method: :put
+          ) do |form| %>
+          <div class="card-body">
+            <p><%= "Publish Time: #{l(@planner.publish_time, format: :time_12_hour)}" %></p>
```
radical-ube, line 12: yeah, i'm wondering if i just put this and the `segment_count` inputs here? instead of updating them in the "main" partial.

app/views/podcast_planner/_form_side.html.erb
```diff
@@ -0,0 +1,45 @@
...
+            <p><%= "Number of Segments: #{@planner.segment_count}" %></p>
+            <p><%= "Number of Drafts: #{@planner.dates.count}" %></p>
+            <ul class="list-group list-group-numbered">
+              <% @planner.dates.each do |date| %>
+                <li class="list-group-item"><%= l(date, format: :day_and_date) %></li>
+              <% end %>
+            </ul>
+          </div>
+          <div class="card-footer">
+            <%= form.hidden_field :week_condition, value: @planner.week_condition %>
```
radical-ube, line 22: ok, i'm not sure if it is weird to basically re-do all of these params in these `hidden_field`s, or if there was a better way to pull the params from the other form into here without causing Turbo weirdness?

app/models/podcast_planner.rb
cavis: I think this is fairly readable.  If you're interested in a slightly different implementation, you could look at the [Publish code](https://github.com/PRX/publish.prx.org/blob/main/src/app/series/directives/series-plan.component.ts#L109).

Kind of similar to a while.  But the main difference is I had it walk through the entire calendar one day at a time, rather than adding weeks to it.  And for each day, it checks:

- Are we past max episodes? Exit early.
- Are we past the max date? Exit early.
- Is the date one of the days-of-week we should add?  And is it one of the weeks-of-month we should add?  Then add it.

app/models/podcast_planner.rb
cavis: I'm kind of surprised `insert_all` doesn't let you pass in an array of `Episode.new` unsaved models.

app/views/podcast_planner/_form_date_calculator.html.erb
cavis: I think we generally push users to plan _at least_ 1 year of future episodes?  In Publish, (with the walk-through-every-day starting tomorrow code), I have `MAX_PLAN_DAYS = 730`.  So that's 2 years.

app/views/podcast_planner/_form_date_calculator.html.erb
cavis: 100 is probably sane.  In Publish, I had to artificially limit things because creating episodes is slow (a bunch of API requests running in series).

app/views/podcast_planner/_form_main.html.erb
cavis: Nice!  Seems to work well.

app/views/podcast_planner/_form_options.html.erb
cavis: This feels sane.  I can't imagine having more than 5 segments... but who knows.  Maybe someone wants just a ton of midroll ads.

app/views/podcast_planner/_form_side.html.erb
cavis: Once this merges, may be worth another pass on the UX.  I was also thinking the segment-count input should be a lot higher on the page, so they don't miss it.

(It would be a pain to generate dozens of episodes with the wrong segment_count, and have to go update them all one by one).

app/views/podcast_planner/_form_side.html.erb
cavis: There's probably some way to reuse the turbo fields?

But Turbo seems to work best when you write simple (if redundant) HTML, without a bunch of magic.  So I think this is a good approach.

app/views/podcast_planner/_form_side.html.erb
```diff
@@ -0,0 +1,45 @@
...
+            <%= form.hidden_field :period, value: @planner.period %>
+            <%= form.hidden_field :start_date, value: @planner.start_date %>
```
cavis, line 24: Why is it necessary to pass all the how-to-generate-the-calendar params again?  You already have the array of timestamps right?

I'd think it would be sufficient to pass just the segment_count and the array of UTC timestamps. Which would also let the user edit those dates before submitting (Publish actually lets you select/unselect arbitrary days on the calendar - and I think eventually we'll want that here).

cavis (changes requested):
Initial comments!

The turbo stuff works pretty slick.  Generally, I think you've got a good approach here.  Mostly looking for some tests on the calendar generator.  And a suggestion on what hidden fields to pass.

app/javascript/controllers/click_controller.js
```diff
@@ -0,0 +1,9 @@
+import { Controller } from "@hotwired/stimulus"
+
+export default class extends Controller {
+  static targets = ["submit"]
+
+  submit() {
+    this.submitTarget.click()
```
radical-ube, line 7: genericized this controller and method for future reuse.

app/models/podcast_planner.rb
```diff
@@ -0,0 +1,163 @@
...
+  end
+
+  def week_of_the_month(date)
+    (date.day.to_f / 7).ceil
+  end
+
+  def days_between(start_date, end_date)
+    (end_date - start_date).to_i
+  end
+
```
radical-ube, line 88: 3 "validations" below to abstract the selection logic into singular methods

app/models/podcast_planner.rb
```diff
@@ -0,0 +1,163 @@
...
+      elsif monthly? && valid_week_to_add?(current_day)
+        @dates.push(current_day)
+      end
+    end
+  end
+
+  def past_max_date?(current_day)
+    current_day.after?(@start_date + 730.days)
+  end
+
```
radical-ube, line 114: i've decided to follow closer to the original code, and iterate by day instead of chunking by week. it did help with readability (i think).

app/models/podcast_planner.rb
radical-ube: updated to use `Episode.new` here, and then the controller saves the models.

test/models/podcast_planner_test.rb
```diff
@@ -0,0 +1,423 @@
+require "test_helper"
+
```
radical-ube, line 2: new tests for the `PodcastPlanner` model. did not test every single small method, but focused on the larger more complex ones that utilize the smaller ones.

test/models/podcast_planner_test.rb
```diff
@@ -0,0 +1,423 @@
...
+      planner.dates = []
+      assert_equal planner.ready_to_generate_drafts?, false
+
+      planner.dates = dates
+      planner.publish_time = nil
+      assert_equal planner.ready_to_generate_drafts?, false
+    end
+  end
+
+  describe "generators" do
```
radical-ube, line 98: for each test in this suite, i opted to add comments of the dates to assist in readability, since the numbers alone don't necessarily make it apparent.

test/models/podcast_planner_test.rb
```diff
@@ -0,0 +1,423 @@
...
+          planner.dates.each do |date|
+            assert_equal date > planner.start_date, true
+            assert_equal date <= planner.end_date, true
+          end
+        end
+      end
+    end
+
+    describe "generate_drafts!" do
+      before do
```
radical-ube, line 370: instead of using `generate_dates!` here (and having to set all of the different attributes necessary), i'm just assigning `dates` for brevity.

test/models/podcast_planner_test.rb
```diff
@@ -0,0 +1,423 @@
...
+          DateTime.new(2001, 7, 2),
+          # Wednesday, July 4
+          DateTime.new(2001, 7, 4)
+        ]
+        # 9:30 AM
+        planner.publish_time = DateTime.new(2001, 2, 3, 9, 30)
+        planner.segment_count = 2
+      end
+
+      describe "Episodes" do
```
radical-ube, line 398: i actually wasn't sure what i needed to test in this suite, so i just check that the `Episode`s look correct according to the `dates`, `publish_time`, and `segment_count` supplied.

farski (commented):
image

I'm finding these controls a bit confusing. I think it's mostly just a matter of verbiage and labels. "Calculate the weeks by periodic", for example, is probably not intuitive for a lot of users. Even I have to think pretty hard to entirely parse what it means. "Week condition" similarly is a very jargony term. I think right now it would be a lot of trial and error with people changing settings to see how the preview changes, but I also think a few tweaks to the UI would give users more confidence when interacting with the controls.

brandonhundt:
> I'm finding these controls a bit confusing. I think it's mostly just a matter of verbiage and labels. "Calculate the weeks by periodic", for example, is probably not intuitive for a lot of users. Even I have to think pretty hard to entirely parse what it means. "Week condition" similarly is a very jargony term. I think right now it would be a lot of trial and error with people changing settings to see how the preview changes, but I also think a few tweaks to the UI would give users more confidence when interacting with the controls.

@farski I agree there are some verbiage tweaks needed. I didn't want word choices to gum up the PR. I'm going to write a ticket to circle back on it soon!

brandonhundt:
Thanks @radical-ube, lets not sweat the verbiage here! I opened #606 to address after the fact!

app/helpers/podcast_planner_helper.rb
```diff
@@ -0,0 +1,37 @@
...
+module PodcastPlannerHelper
+  PERIODIC_WEEKS = I18n.t([:every_one, :every_two, :every_three, :every_four], scope: ".podcast_planner.helper.period_options").freeze
```
farski, line 4: I think these need to be like `scope: [:podcast_planner, :helper, :period_options]`

farski (commented):
When I run this locally, some of the selects are dropping the first value. `Sunday` and `First` are not visible.

image
image

radical-ube:
> When I run this locally, some of the selects are dropping the first value. `Sunday` and `First` are not visible.
> 

when you inspect the element, do you still see the options (and they are just not displaying for some reason?), or are those particular options cut out completely? i'm not sure how to reproduce this bug.

Screenshot 2023-03-28 at 3 45 55 PM

farski:
> > When I run this locally, some of the selects are dropping the first value. `Sunday` and `First` are not visible.
> 
> when you inspect the element, do you still see the options (and they are just not displaying for some reason?), or are those particular options cut out completely? i'm not sure how to reproduce this bug.
> 
> Screenshot 2023-03-28 at 3 45 55 PM

Yes, they are disabled with a `d-none` class.

farski (commented):
I realize this work is based on what exists in Publish, but that UX has never really felt optimal to me.

In my mind, there are two different modes that should be addressed with this tool.

The first I'll call "frequency-based". In this mode you would select a period `P` (daily, weekly, monthly, etc) and a frequency `X`. e.g., "Every X P"
- Every 2 weeks (aka every other week)
- Every 7 weeks
- Every 1 month (aka every month)
- Every 19 days

And then you would select which members of that period are included:
- For daily, there's no selection, it's purely based on the frequency
- For weekly, you would select which days of the week to include (e.g., M W F)
- For monthly, you would select which days of the month to include (e.g., 8th, 21st)

Frequency-based is the general case.

The edge case is the week-of-month mode, because it's a highly opinionated mode. There's no clear definition of what a "week" is in this case. The way publish currently works, it considers the "first week of the month" to be the first 7 days of the month, **not** the first calendar week of the month. For scheduling purposes, to me at least, that's unexpected. For calendaring and scheduling purposes, if someone said to me "let's aim for the first week of June", I wouldn't look at the somewhat arbitrary 7 day period that starts on a Thursday and ends on a Wednesday. I may look at the 1st through the 3rd, or I may look at the first full Sunday-to-Saturday week (the 4th to 10th). Some people I've talked to would even include the end of May in that week. It's ambiguous without more clarification. This mode needs to explain clearly what it means by "week", or it needs to provide the various options.

I think mode selection should be the first choice made, rather than Weekdays, since there are cases where you never end up selected Weekdays. I think switching that order means the other controls can be more purpose-built, rather than doing slightly different things in different modes, and the resulting UI ends up being simpler at any given time.

A few examples of these different branches:

image
image
image
image

radical-ube:
i think i solved the option d-none issue with the new commit push. but i'm not sure as i wasn't able to reproduce the bug locally for some reason, although i had an inkling of what was causing it.

farski:
> i think i solved the option d-none issue with the new commit push. but i'm not sure as i wasn't able to reproduce the bug locally for some reason, although i had an inkling of what was causing it.

That fixed it for me

radical-ube:
> I realize this work is based on what exists in Publish, but that UX has never really felt optimal to me.
> 
> In my mind, there are two different modes that should be addressed with this tool.
> 
> The first I'll call "frequency-based". In this mode you would select a period `P` (daily, weekly, monthly, etc) and a frequency `X`. e.g., "Every X P"
> 
> * Every 2 weeks (aka every other week)
> * Every 7 weeks
> * Every 1 month (aka every month)
> * Every 19 days
> 
> And then you would select which members of that period are included:
> 
> * For daily, there's no selection, it's purely based on the frequency
> * For weekly, you would select which days of the week to include (e.g., M W F)
> * For monthly, you would select which days of the month to include (e.g., 8th, 21st)
> 
> Frequency-based is the general case.
> 
> The edge case is the week-of-month mode, because it's a highly opinionated mode. There's no clear definition of what a "week" is in this case. The way publish currently works, it considers the "first week of the month" to be the first 7 days of the month, **not** the first calendar week of the month. For scheduling purposes, to me at least, that's unexpected. For calendaring and scheduling purposes, if someone said to me "let's aim for the first week of June", I wouldn't look at the somewhat arbitrary 7 day period that starts on a Thursday and ends on a Wednesday. I may look at the 1st through the 3rd, or I may look at the first full Sunday-to-Saturday week (the 4th to 10th). Some people I've talked to would even include the end of May in that week. It's ambiguous without more clarification. This mode needs to explain clearly what it means by "week", or it needs to provide the various options.
> 
> I think mode selection should be the first choice made, rather than Weekdays, since there are cases where you never end up selected Weekdays. I think switching that order means the other controls can be more purpose-built, rather than doing slightly different things in different modes, and the resulting UI ends up being simpler at any given time.
> 

FWIW, i agree and also believe that frequency-based mode is the general case (it's what i normally think of with scheduling, and also where my logic for "period" selection came from). "week-of-the-month" doesn't make too much sense to me either, but i am told that producers opt for this more often? (but also, their only other option in Publish is "every other week", so the data naturally skews toward using "week-of-the-month" if the frequency option doesn't do the desired effect)

development-wise, i didn't want to challenge the legacy design on this first pass. however, i would be up for (in next ticket / next pass) refactoring to take the planner in this direction.

app/views/podcast_planner/_form_side.html.erb
```diff
@@ -0,0 +1,48 @@
...
+                <%= form.label :segment_count %>
+              </div>
+            </div>
+
+            <div class="col">
+              <p><%= t(".header.number_of_drafts", count: @planner.dates.count) %></p>
+              <ul class="list-group list-group-numbered">
+                <% @planner.dates.each do |date| %>
+                  <li class="list-group-item"><%= l(date, format: :day_and_date) %></li>
+                  <%= form.hidden_field :generated_dates, multiple: true, value: date %>
```
cavis, line 33: +10 ... and I think this will help when you get to the calendar UI.  Since the user should be able to select/unselect dates here _after_ generating the initial schedule.  (Maybe there are a few days they'll skip, or some additional bonus episodes to add).

test/controllers/podcast_planner_controller_test.rb
```diff
@@ -10,7 +10,7 @@ class PodcastPlannerControllerTest < ActionDispatch::IntegrationTest
   end
 
   test "should update podcast_planner" do
-    patch podcast_planner_url(@podcast), params: {podcast_planner: {}}
-    assert_redirected_to podcast_planner_url(@podcast)
+    # patch podcast_planner_url(@podcast), params: {podcast_planner: {}}
```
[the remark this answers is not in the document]
cavis: For controller tests, I like to test very basic things (validates, authorizes, etc).  You probably want to do something similar to the podcast-controller test edit/update:

- Make sure `#show` renders successfully
- Make sure `#show` authorizes (renders forbidden), if you're not a user with access to this podcast
- Maaaaybe also send some query params to `#show`?
- Patch `#create` with valid params
- Patch `#create` with invalid params
- Patch `#create` with invalid auth/account

test/models/podcast_planner_test.rb
cavis: Makes sense.  You could also have skipped `.save!`-ing these episodes, and just look at the `planner.drafts` array (would be slightly faster to not hit the db ... just assert that they look `.valid?`).

cavis (approved):
Nice testing!

Okay okay, I think this is nearly there.  This PR's comment thread is getting long, so tempted to just merge this and have you make a new PR for a few minor things.

PR #592, opened 2023-03-07
Fix/guard on remote apple state

svevang:
Fixes a couple of problems:

- Apple validates attributes after they are persisted.  The RSS scraper apparently does not validate any data before persisting it in the Apple episode model. Then if we're PATCHing a single, unrelated, field. We get back errors on the unvalidated.  So for example when updating the audio container reference on the episode, we can get back validation errors on an associated email address
- In debugging the prior issue,  I discovered that the my assumption that delivery and delivery files associated with the PodcastsConnect UI are not exposed through the UI.  And in my initial testing, I was not getting any back for UI uploaded subscriber audio.  In testing the mix of API and UI uploads, it turns out that the UI _does_ create delivery and delivery files that are exposed via the API.

To fix the above two issues, this PR:

- Makes the error handling reentrant across container/delivery/delivery-file update and create actions (something we needed anyways). So this sets up checks on the results / errors and makes sure to update local state for those requests that succeeded.
- Handles the `has_many` relations on `podcast_container->deliveries` and `delivery->delivery_files` by extending the system that joins local resources (state) to some set of remote api responses. 
   - Adds a new `join_on_many` helper in the api_response module.
   - Checks for join key overlap in the local resources and the request responses
   - Checks for join key equality in the case of the 'ordinary' one to one join
   - adds a `left_join` option for those resources that do not expect a row to be returned from set of (chained) remote api calls

### DIFF
--- a/test/models/apple/api_response_test.rb
+++ b/test/models/apple/api_response_test.rb
@@ -52,29 +52,78 @@ describe Apple::ApiResponse do
     end
   end
 
-  describe ".join_on_apple_episode_id" do
-    it "uses the request metadata to join a set of resources responging to `apple_episode_id`" do
-      row_results = [{request_metadata: {apple_episode_id: "1"}, api_response: {foo: "111"}},
-        {request_metadata: {apple_episode_id: "2"}, api_response: {foo: "222"}},
-        {request_metadata: {apple_episode_id: "3"}, api_response: {foo: "333"}}]
-      row_results = row_results.map(&:with_indifferent_access)
+  describe "joins" do
+    let(:row_results) do
+      [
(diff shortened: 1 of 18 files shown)


test/models/apple/api_response_test.rb
```diff
@@ -52,29 +52,79 @@ class TestResponse
...
+      [
         OpenStruct.new(apple_episode_id: "1", bar: "111"),
         OpenStruct.new(apple_episode_id: "2", bar: "222"),
         OpenStruct.new(apple_episode_id: "3", bar: "333")
       ]
+    end
+
+    describe "many to one joins" do
+      it "joins on apple_episode_id" do
+        zipped = TestResponse.join_many_on("apple_episode_id", resources, row_results)
```
cavis, line 74: Joins!

(And tests++... I actually started with the tests and worked backwards looking at this PR)

PR #593, opened 2023-03-07
Setup spring binstubs

svevang:
This PR will:
- update spring
- setup binstubs to use spring

Most noticeable when you are running a single test repeatedly, for instance. This will speed up rails boot time.

```bash
# Invoking rails irb repeatedly on `main`
bash-5.1.16$ time : $(echo "1" | ./bin/rails c)                                                                
                                                                                                                       
real    0m5.128s                                                                                                       
user    0m2.220s    
sys     0m1.302s 

```

versus

```bash
# Invoking rails irb repeatedly on `chore/spring-for-binstubs`
bash-5.1.16$ time : $(echo "1" | ./bin/rails c)                                       
DEBUGGER[spring app    | feeder.prx.org | started 10 secs ago | development mode#61061]: Attaching after process 60898 fork to child process 61061
Running via Spring preloader in process 61061                                                                                                                                                                                                 
                                                                                                                       
real    0m0.678s                                                                                                       
user    0m0.218s                                           
sys     0m0.181s

### DIFF
--- a/bin/rails
+++ b/bin/rails
@@ -1,4 +1,5 @@
 #!/usr/bin/env ruby
+load File.expand_path("spring", __dir__)
 APP_PATH = File.expand_path('../config/application', __dir__)
 require_relative '../config/boot'
 require 'rails/commands'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -404,7 +404,7 @@ GEM
     simplecov-html (0.12.3)
     simplecov_json_formatter (0.1.4)
     smart_properties (1.17.0)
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

File: bin/rails [hunk 1]
```diff
@@ -1,4 +1,5 @@
 #!/usr/bin/env ruby
+load File.expand_path("spring", __dir__)
```
cavis, line 2: Are these changes just the result of `bundle exec spring binstub --all`?

cavis (approved):
👍 Looks good to me!

PR #598, opened 2023-03-24
597 podcast sort dropdown fixes

### LINKED ISSUE #597: Podcast sort dropdown fixups
Couple minor things to fix on the Podcast#index sort picker.

image

- [x] Don't underline on hover/focus
- [x] Decide/design some UI for how the _current_ sort order is indicated.  Currently you'd have to expand the dropdown to see your current sort order.  But probably some better/other options.
- [x] Implement the ruby code to indicate what the current sort is (adding the `active` class, or whatever)

a-maci29:
Closes #597


app/views/podcasts/index.html.erb
a-maci29, line 1: As an aside, I noticed that the `Recent Activity` selection wasn't responding because it didn't match the hash key in `DISPLAY_ORDER` back in the controller, looks like it fixed it?

app/controllers/podcasts_controller.rb
```diff
@@ -64,13 +64,13 @@ def create
 
   def sort_podcasts_by
     if params[:sort] == "Recent Activity"
-      "Recent Activity"
+      render html: '<span class="material-icons">timer</span>'.html_safe
```
a-maci29, line 67: Something isn't quite working with the `material-icons-outlined` class as is, looking for solution

cavis:
Dup: #621

PR #602, opened 2023-03-27
Fix/break out enclosure url

svevang:
Hit a case where the private branch redirect was alternating across a set of arrangements. 

This PR sets up some guards on the redirect, and tracks the resultant CDN url so that the delivery files are based on that single arrangement.

This also sets up a way to recover from bad audio data in Apple's system:

- Allow replacing audio by marking existing audio as deleted
- Mark the episode's podcast container to immediately disconnect from Apple hosted audio (if needed).

### DIFF
--- a/app/models/apple/podcast_container.rb
+++ b/app/models/apple/podcast_container.rb
@@ -6,7 +6,7 @@ module Apple
 
     serialize :api_response, JSON
 
-    has_many :podcast_deliveries
+    has_many :podcast_deliveries, dependent: :destroy
     has_many :podcast_delivery_files, through: :podcast_deliveries
     belongs_to :episode, class_name: "::Episode"
 
@@ -42,6 +42,8 @@ module Apple
 
       join_on_apple_episode_id(episodes, results, left_join: true).each do |(ep, row)|
         next if row.nil?
+        apple_id = row.dig("api_response", "val", "data", "id")
+        raise "missing apple id!" unless apple_id.present?
 
         upsert_podcast_container(ep, row)
       end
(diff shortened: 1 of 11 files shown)


app/models/apple/podcast_container.rb
```diff
@@ -68,7 +74,7 @@ def self.upsert_podcast_container(episode, row)
           vendor_id: episode.audio_asset_vendor_id).first)
 
           pc.update(api_response: row,
-            source_url: episode.enclosure_url,
+            enclosure_url: episode.enclosure_url,
```
cavis, line 77: Is this literally the `<item><enclosure url="">` from the private feed?  (Aka `Episode.enclosure_url`).  Or something else?

And if so - does this copy ever deviate from the private-Feed episode enclosure_url?

svevang: Naming is hard.  And confusingly we have a `Apple::Episode` in addition to the base `Episode`.

This is coming from the apple model here:
https://github.com/PRX/feeder.prx.org/blob/main/app/models/apple/episode.rb#L130

If I were to rename this method, maybe `#private_feed_base_enclosure_url` might make sense?

This does deviate from the  private-feed episode enclosure_url in that this should strip out all third party additions to the enclosure.

svevang: err,  or maybe I didn't understand your question?

cavis: Oh, i see.  It's the private-feed-episode's enclosure URL, but for a _specific_ private-feed-token (i think).  And whatever other changes it needs to make - the `?noImp` for instance

Okay, I think the naming is fine.  Refers to the specific fully-expanded url you're using to download the media.

cavis (approved):
👍 looks good to me!

PR #603, opened 2023-03-27
Fix/constrain batch sizes

svevang:
Merge after https://github.com/PRX/feeder.prx.org/pull/602

Counterpart to https://github.com/PRX/api-bridge-lambda/pull/17, this sets up a new batch size parameter to constrain the number of workers hitting the remote API.


cavis (approved):
👍

PR #613, opened 2023-04-05
simple calendar

### LINKED ISSUE #612: Development: calendar for episode scheduler

radical-ube:
resolves #612


app/controllers/podcast_planner_controller.rb
```diff
@@ -1,10 +1,15 @@
...
   before_action :set_podcast
+  before_action :set_beginning_of_week
 
   # GET /podcasts/1/planner
   def show
     @planner = PodcastPlanner.new(planner_params)
     @planner.generate_dates!
+
+    if @planner.dates.present?
+      @planner_months = @planner.dates_by_month
```
radical-ube, line 11: could probably just make this its own attribute, but essentially this splits the dates into arrays by month.

app/javascript/controllers/calendar_controller.js
```diff
@@ -0,0 +1,28 @@
+import { Controller } from "@hotwired/stimulus"
+
+export default class extends Controller {
+  static targets = ["date"]
+
+  connect() {
```
radical-ube, line 6: connect happens with each monthly calendar that is created in the DOM. so this cycles through each date and adds a toggle to each date. the `date` is the hidden field, and the parent is the td box itself, which is the interactable element in the UI.

app/javascript/controllers/calendar_controller.js
```diff
@@ -0,0 +1,28 @@
...
+    this.dateTargets.forEach((date) => {
+      date.parentElement.addEventListener("click", this.toggleSelect)
+
+      if (date.hasAttribute("selected")) {
```
radical-ube, line 10: this renders the blue background on the dates that have been generated through the calculator.

app/javascript/controllers/calendar_controller.js
```diff
@@ -0,0 +1,28 @@
...
+        date.parentElement.setAttribute("selected", true)
+        date.parentElement.classList.add("bg-primary", "text-light")
+      }
+    })
+  }
+
+  toggleSelect(event) {
```
radical-ube, line 17: since the `event.target` is the parent td box, the `disabled` attribute is added/removed on the child, which is the actual `hidden_field` form element that needs to be enabled/disabled to make through to the controller params.

app/models/podcast_planner.rb
```diff
@@ -1,8 +1,8 @@
 class PodcastPlanner
-  attr_accessor :podcast_id, :dates, :selected_days, :week_condition, :period, :monthly_weeks, :start_date, :date_range_condition, :number_of_episodes, :end_date, :publish_time, :segment_count, :generated_dates, :drafts
+  attr_accessor :podcast_id, :dates, :selected_days, :week_condition, :period, :monthly_weeks, :start_date, :date_range_condition, :number_of_episodes, :end_date, :publish_time, :segment_count, :selected_dates, :drafts
```
radical-ube, line 2: selected dates seems more accurate now.

app/models/podcast_planner.rb
```diff
@@ -160,4 +160,23 @@ def apply_publish_time(date)
   def generate_default_title(date)
     I18n.l(date, format: :day_and_date).to_s
   end
+
```
radical-ube, line 163: helper methods to separate the dates into months

app/views/podcast_planner/_calendar.html.erb
```diff
@@ -0,0 +1,13 @@
+<div class="col" data-controller="calendar">
```
radical-ube, line 1: i essentially wrap the simple calendar with a controller for this partial.

app/views/podcast_planner/_calendar.html.erb
```diff
@@ -0,0 +1,13 @@
...
+  <%= month_calendar(start_date: month.first) do |date| %>
+    <% if date_is_within_month(date, month) %>
+      <%= date.day %>
+
+      <% if is_preselected_date?(date, month) %>
```
radical-ube, line 6: feels a little wordy, but essentially pre-selects the hidden_field if it's one of the calculated dates, and disables the field if not.

app/controllers/podcast_planner_controller.rb
cavis: Is there any way to make the monthly-calendars show up, even if no generator-dates are selected?  So a user can sort of mix/match generating the dates and manually selecting them.

(Maybe this could be a different issue?  No need to clutter this PR)

app/javascript/controllers/calendar_controller.js
```diff
@@ -0,0 +1,28 @@
+import { Controller } from "@hotwired/stimulus"
+
+export default class extends Controller {
+  static targets = ["date"]
+
+  connect() {
+    this.dateTargets.forEach((date) => {
+      date.parentElement.addEventListener("click", this.toggleSelect)
```
cavis, line 8: Generally, if you add an event listener somewhere, you need to also remove it when the stimulus controller disconnects.

Might be better to move this into the parent elements `data-action` instead?  Then stimulus handles all the event listening/disconnecting for you.

app/views/podcast_planner/_calendar.html.erb
cavis: Alternatively ... it might cleanup your controller code to have 1 controller _per day_ rather than 1 global.  Then you wouldn't need to do a bunch of parent/child checking - you could just tie the 1 controller explicitly to the parent and form-field elements.

app/views/podcast_planner/_calendar.html.erb
cavis: +1 makes sense to me.

You could also drop the if/else, and just make this `form.hidden_field ..., disable: !is_preselected_date?`

app/javascript/controllers/calendar_controller.js
```diff
@@ -0,0 +1,28 @@
...
+
+      if (date.hasAttribute("selected")) {
+        date.parentElement.setAttribute("selected", true)
+        date.parentElement.classList.add("bg-primary", "text-light")
+      }
+    })
+  }
+
+  toggleSelect(event) {
+    if (event.target.hasAttribute("selected")) {
```
cavis, line 18: Similarly, this `<input type="hidden">` "selected" attribute seems redundant - couldn't you just check if the element is disabled or not?

app/models/podcast_planner.rb
cavis: +1

app/views/simple_calendar/_week_calendar.html.erb
```diff
@@ -0,0 +1,37 @@
...
+        <% end %>
+      </tr>
+    </thead>
+
+    <tbody>
+      <% date_range.each_slice(7) do |week| %>
+        <tr>
+          <% week.each do |day| %>
+            <%= content_tag :td, class: calendar.td_classes_for(day) do %>
+              <% if defined?(Haml) && respond_to?(:block_is_haml?) && block_is_haml?(passed_block) %>
```
cavis, line 26: Haml!  /me shakes fist at haml!

I'm guessing these `simple_calendar` views are generated?

app/controllers/podcast_planner_controller.rb
cavis: Another way to think about it:

If you decouple the calendar-generating code from the "which days on the calendar are selected" code, it may simplify things?  Just loop through N months starting at `Date.today.beginning_of_month` to generate the calendars.  And for each rendered day, ask "is this selected?"

cavis (changes requested):
This is looking/working good!

Couple suggestions on Stimulus controllers, and optionally a way to make the calendars show up before the user fills out the generator fields.

app/views/simple_calendar/_week_calendar.html.erb
cavis: HAML is just an alternative some people use to html.erb views, in their rails projects.

Since it's generated from the gem, i'd say it's fine.  Boilerplate gonna boiler plate.

app/assets/stylesheets/shared/shared.scss
```diff
@@ -22,3 +22,4 @@
 @import "table";
 @import "tabs";
 @import "toast";
+@import "calendar";
```
radical-ube, line 25: for some reason, adding a `.scss` file to the `/app` directory wasn't properly being read, so i've added it here in the `/shared` directory for now. could also be a future discussion for if we want to have shared calendar styling across apps.

app/controllers/podcast_planner_controller.rb
```diff
@@ -1,10 +1,12 @@
 class PodcastPlannerController < ApplicationController
   before_action :set_podcast
+  before_action :set_beginning_of_week
 
   # GET /podcasts/1/planner
   def show
     @planner = PodcastPlanner.new(planner_params)
     @planner.generate_dates!
+    @draft_dates = @podcast.episodes.draft.pluck(:released_at)
```
radical-ube, line 9: a new feature i've implemented, to show drafts that already exist for the podcast. these will show in yellow on the calendar.

app/helpers/podcast_planner_helper.rb
```diff
@@ -36,4 +40,32 @@ def time_options
...
+    else
+      false
+    end
+  end
+
+  def date_is_in_month?(date, month)
+    date.month == month
+  end
+
+  def calendar_day_tag(day:, month:, calendar:, &block)
```
radical-ube, line 60: custom tag that modifies the `content_tag` from the gem generated partials. adds a `dateController` to dates within that month and avoids adding to "blank" spaces that are not within that month.

app/javascript/controllers/date_controller.js
```diff
@@ -0,0 +1,67 @@
+import { Controller } from "@hotwired/stimulus"
+
+export default class extends Controller {
+  static targets = ["preselected"]
+
+  connect() {
```
radical-ube, line 6: calendars are re-rendered every time the generator makes a change that triggers a turbo refresh, so i check to see which dates have been pre-selected on `connect()`

app/javascript/controllers/date_controller.js
```diff
@@ -0,0 +1,67 @@
...
+    if (!this.isDisabled(this.preselectedTarget)) {
+      if (this.isDraft(this.preselectedTarget)) {
+        this.element.classList.add("bg-danger", "text-light")
+      } else {
+        this.element.classList.add("bg-primary", "text-light")
+      }
+    }
+
+    if (this.preselectedTarget.getAttribute("draft") === "true") {
```
radical-ube, line 15: existing drafts render in yellow.

app/javascript/controllers/date_controller.js
```diff
@@ -0,0 +1,67 @@
+import { Controller } from "@hotwired/stimulus"
+
+export default class extends Controller {
+  static targets = ["preselected"]
+
+  connect() {
+    if (!this.isDisabled(this.preselectedTarget)) {
+      if (this.isDraft(this.preselectedTarget)) {
+        this.element.classList.add("bg-danger", "text-light")
```
radical-ube, line 9: "double-booked" drafts render in red, otherwise normal blue.

app/javascript/controllers/date_controller.js
```diff
@@ -0,0 +1,67 @@
...
+      if (this.isDraft(event.target.firstElementChild)) {
+        event.target.classList.add("bg-danger", "text-light")
+      } else {
+        event.target.classList.add("bg-primary", "text-light")
+      }
+      event.target.firstElementChild.disabled = null
+    }
+  }
+
+  highlight(event) {
```
radical-ube, line 37: `highlight` and `unhighlight` trigger on mouseover/mouseoff, to highlight the date the mouse is currently on.

app/javascript/controllers/pager_controller.js
```diff
@@ -0,0 +1,44 @@
...
+  static targets = ["page", "prev", "next"]
+  static values = {
+    currentPosition: { type: Number, default: 0 },
+    range: { type: Number, default: 6 },
+    step: { type: Number, default: 1 },
+  }
+
+  currentPositionValueChanged() {
+    this.pageTargets.forEach((el) => {
+      if (this.inRange(parseInt(el.getAttribute("index")))) {
```
radical-ube, line 13: so many parentheses lol. wasn't sure if `parseInt` was necessary here. sometimes i forget if strings/ints have to be strictly compared or not in JS.

app/views/podcast_planner/_form_side.html.erb
```diff
@@ -2,47 +2,55 @@
   <div class="card-header-primary">
     <h5 class="card-title"><%= t(".header.preview") %></h5>
   </div>
-  <%= turbo_frame_tag "preview", target: "_top" do %>
```
radical-ube, line 5: moved the turbo frame just around the calendars, which also helps to prevent the `publish_time` and `audio_segments` values being refreshed unnecessarily.

app/views/podcast_planner/_form_side.html.erb
```diff
@@ -2,47 +2,55 @@
...
-    <% if @planner.dates.present? %>
```
radical-ube, line 6: also removed this conditional so that the calendars don't depend on the calculator/generator to render.

app/views/podcast_planner/_form_side.html.erb
```diff
@@ -2,47 +2,55 @@
...
+              <%= form.number_field :segment_count, {min: 1, max: 10, step: 1, value: 2} %>
+              <%= form.label :segment_count %>
             </div>
           </div>
-          <div class="card-footer">
-            <%= form.submit t(".label.submit"), class: "btn btn-primary" %>
+        </div>
+
+        <div class="col-lg">
+          <div data-controller="pager" data-pager-range-value="6">
```
radical-ube, line 30: new `pagerController` added here, wrapping the calendars.

app/views/simple_calendar/_month_calendar.html.erb
```diff
@@ -0,0 +1,31 @@
...
+          <th><%= t("date.abbr_day_names")[day.wday] %></th>
+        <% end %>
+      </tr>
+    </thead>
+
+    <tbody>
+      <% date_range.each_slice(7) do |week| %>
+        <tr>
+          <% week.each do |day| %>
+            <%= calendar_day_tag(day: day, month: start_date.month, calendar: calendar) do %>
```
radical-ube, line 19: custom tag implemented here, otherwise all default gem code (minus their original pagers)

app/assets/stylesheets/shared/shared.scss
cavis: Hmm... so you added the file `app/assets/stylesheets/app/calendar.scss`, then add an `@import "calendar"` to the `app/index.scss` right?

I would have expected that to work - things should generally work now that i've added explicit imports of the "barrel files" (`@import "app/index"` instead of just `@import "app"`).

app/controllers/podcast_planner_controller.rb
cavis: Nice!

app/javascript/controllers/pager_controller.js
cavis: You could also ditch the index attribute, and just get the index from the forEach. Can be inferred from their order in the DOM anyways.

```
this.pageTargets.forEach((el, index) => {
```

app/views/podcast_planner/_form_side.html.erb
cavis: +1 good call

app/javascript/controllers/date_controller.js
cavis: I'd actually suggest doing this in pure CSS with `:hover`.

app/javascript/controllers/date_controller.js
cavis: This seems redundant with the `disabled` property - couldn't you just check if the element is disabled or not here?

app/views/podcast_planner/_calendar.html.erb
```diff
@@ -0,0 +1,9 @@
+<div class="col <%= "d-none" if index >= range %>" data-pager-target="page" index="<%= index %>">
+  <%= month_calendar(start_date: start_date) do |date| %>
+    <% if date_is_in_month?(date, start_date.month) %>
+      <%= date.day %>
+
+      <%= form.hidden_field :selected_dates, multiple: true, data: {date_target: "preselected"}, value: date.to_datetime, disabled: !is_preselected_date?(date, @planner.dates), draft: is_preselected_date?(date, @draft_dates) %>
```
cavis, line 6: Yeah, the `draft` attribute appears redundant with `!disabled`.

app/javascript/controllers/date_controller.js
radical-ube: possibly naming confusion happening here: `drafts` are the episodes that have already been scheduled and will appear in yellow on the calendar. not necessarily tied to or mutually exclusive from the `disabled` property which is _any date_ chosen, so that's the reasoning behind it being a different attribute.

app/javascript/controllers/pager_controller.js
radical-ube: oh right, JS iterating magic. will update.

app/assets/stylesheets/shared/shared.scss
radical-ube: i will `merge main` and try again, but i did try `@import "calendar"` in the `index` file and added the `calendar` file and it didn't seem to work. i even tried to follow the audio segmenter patterning and make a new directory but it didn't seem to work either.

app/javascript/controllers/date_controller.js
cavis: Ohhhh... you're right.  I misread that.  Disregard!

app/assets/stylesheets/shared/shared.scss
cavis: I think [this is the commit](https://github.com/PRX/feeder.prx.org/commit/4934663958af36073b7bdbd37b26c09b3bb41d98) that fixed it.

Doing an `@import "audio-segmenter"` was bad at getting any updated files.  It needed the `/index` on it.

cavis (approved):
👍 looks good to me!

PR #614, opened 2023-04-06
Fix form state styling

brandonhundt:
- Restyles help text (creates a few custom classes)
- Renamed help-text button
- Reinstates changed state styling (`.is-changed`)
- Moves the help-text button outside of `form-floating` class (this is the bulk of the code change you'll see)
- Creates the `prx-field-group` class that uses CSS grid to allow `invalid-feedback` text appears on its own row without affecting the rest of the styling

h/t @rpeterman-gp

### DIFF
--- a/app/assets/stylesheets/shared/form.scss
+++ b/app/assets/stylesheets/shared/form.scss
@@ -308,20 +308,22 @@ small {
   font-size: 1rem;
 }
 
-.field-help-text {
-  display: flex;
-  justify-content: space-between;
+/* TODO: Apply prx prefix */
+
+.prx-field-group {
+  display: grid;
+  grid-template-columns: 1fr min-content;
+  grid-template-rows: $input-height;
+  grid-auto-rows: auto;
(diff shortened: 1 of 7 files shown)


app/assets/stylesheets/shared/form.scss
```diff
@@ -97,14 +81,61 @@
...
+  .is-changed ~ .form-tag-select {
+    border-color: $orange;
+  }
+  .form-tag-select.form-control-focus {
+    color: $input-focus-color;
+    background-color: $input-focus-bg;
+    border-color: $input-focus-border-color;
+    outline: 0;
+    box-shadow: $input-focus-box-shadow;
+  }
```
cavis, line 137: Still seeing some issues on the episode -> categories field.  When going from an initial blank state to adding the 1st tag.  But... that can be cleaned up later!  (I think that field still needs some non-UI changes too).

cavis (approved):
👍 looks good to me

Issue #618, opened 2023-04-19
Audio/Image handling cleanup


cavis:
Keeping a laundry list of audio/image TODOs, as I encounter them.  So as not to bog down the first pass at them.

- [x] Disable saving while upload in progress
- [x] Display metadata + download-file links
- [x] Disable uploader/buttons/etc when user is not authorized to edit the Podcast/Feed/Episode
- [x] Implement async validations on audio/image/video
- [x] Video file uploads
- [x] Enforce audio file validations on published episodes (and better UI when invalid)
- [x] Handle broken image urls
- [x] Style drag-n-drop uploads
- [x] Alt/caption/credit disabled on new podcasts
- [x] Dropdowns flicker while uploads are processing
- [x] ~Should we be resizing images?  (Scaling down itunes_image to use as thumbnail/feed_image?)~ not yet
- [x] ~Should we be transcoding audio?  (Wav to mp3 for within-Feeder playback, etc?)~  no
- [x] Uploading image, replacing, _then_ deleting 2nd image results in the 1st one still being the `ready_image`
- [x] Move all text to en.yml
- [x] Delete button styling (hover/focus)
- [x] The `?_retry` + "polling" js controller has some weird turbo-frame interaction, which causes infinite retrying/SNS every time it polls?
- [x] Add a shared method for all the repeated `accepts_nested_attributes_for` / `reject_if` code
- [x] Do we need to ask for confirmation when deleting episode media?
- [x] Saving with a validation error nukes unsaved uploaded files

PR #619, opened 2023-04-20
Image uploads

### LINKED ISSUE #620: Podcast Images Card

cavis:
Closes #620.

image

- Adds uploader UI to podcast images and episode image
- Fixups to the podcast/episode `#new` pages
- Fallout/bugs that I'm not going to tackle here listed in #618.


app/controllers/podcasts_controller.rb
```diff
@@ -37,10 +37,16 @@ def show
   # GET /podcasts/new
   def new
     @podcast = Podcast.new(podcast_params)
+    @podcast.set_default_feed
+
+    # TODO: get the default account from ID somehow
+    @podcast.prx_account_uri = helpers.podcast_account_name_options(@podcast).first.last
+    @podcast.clear_attribute_changes(%i[prx_account_uri])
```
cavis, line 44: I needed to set the initial selected account_uri on `#new` - to some account that this user has edit access to.  And then mark it as unchanged, so you don't get the orange outline right away.

But would be much better to somehow get their default account from ID.  Someday...

app/models/concerns/image_file.rb
```diff
@@ -106,7 +106,12 @@ def reset_image_attributes
   end
 
   def detect_image_attributes
-    return if !original_url || (width && height && format)
+    # skip if we've already detected width/height/format
+    return if width.present? && height.present? && format.present?
+
+    # s3 urls cannot be fastimage'd - must wait for async porter inspection
+    return if original_url.blank? || original_url.starts_with?("s3://")
```
cavis, line 113: Still some TODOs in #618 for how we handle async validation of these `s3://` uploads from Feeder.  But this at least makes it work ... don't try to use FastImage to detect image height/width synchronously, if the url starts with s3.

app/models/tasks/copy_image_task.rb
```diff
@@ -1,17 +1,24 @@
 class Tasks::CopyImageTask < ::Task
   before_save do
     if image_resource && status_changed?
-      image_resource.update!(status: status)
       if complete?
-        image_resource.update!(url: image_resource.published_url)
-        podcast.try(:publish!)
+        meta = porter_callback_image_meta
+        if meta
+          update_image!(meta.merge(url: image_resource.published_url))
```
cavis, line 7: Copied from the Content (audio) task with 1 exception...

There's a chance that the image fails validation, because its height/width are too small.  So for now, just catching that and setting status to "error".  Will figure out a better way to relay to the user later.

app/views/images/_complete.html.erb
```diff
@@ -0,0 +1,26 @@
+<div class="card prx-uploads">
+  <div class="card-header-info d-flex align-items-center">
+    <h5 class="card-title flex-grow-1"><%= t("images.title.#{image.model_name.singular}") %></h5>
```
cavis, line 3: All episode uploaders are cards.  And we share the markup for their various states.

But because the content of the card-header changes based on the image status (showing/hiding the info and destroy icons) ... I had to put the entire card inside the turbo-frame.

Just FYI, why there's a bit of repeated code!

svevang (approved):

PR #621, opened 2023-04-21
Sort Dropdown Fixes

### LINKED ISSUE #597: Podcast sort dropdown fixups

a-maci29:
Closes #597

### DIFF
--- a/app/controllers/podcasts_controller.rb
+++ b/app/controllers/podcasts_controller.rb
@@ -4,14 +4,16 @@ class PodcastsController < ApplicationController
   # Translate the user selected sort to a query order argument
   DISPLAY_ORDER = {"A-Z" => {title: :asc},
                    "Z-A" => {title: :desc},
-                   "" => {updated_at: :desc}}.freeze
+                   "" => {updated_at: :desc},
+                   "Recent Activity" => {updated_at: :desc}}.freeze
 
   DEFAULT_PAGE_SIZE = 10
 
   # GET /podcasts
   def index
     base_query = policy_scope(Podcast).page(params[:page]).per(DEFAULT_PAGE_SIZE).includes(default_feed: :feed_images)
-    @podcasts = add_sorting(base_query).filter_by_title(params[:q])
+    filtered_podcasts = base_query.filter_by_title(params[:q])
+    @podcasts = add_sorting(filtered_podcasts)
 
     @published_episodes_counts = Episode.where(podcasts: @podcasts).published.group(:podcast_id).count
     @scheduled_episodes_counts = Episode.where(podcasts: @podcasts).scheduled.group(:podcast_id).count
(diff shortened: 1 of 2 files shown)


app/controllers/podcasts_controller.rb
```diff
@@ -4,14 +4,16 @@ class PodcastsController < ApplicationController
...
+                   "Recent Activity" => {updated_at: :desc}}.freeze
 
   DEFAULT_PAGE_SIZE = 10
 
   # GET /podcasts
   def index
     base_query = policy_scope(Podcast).page(params[:page]).per(DEFAULT_PAGE_SIZE).includes(default_feed: :feed_images)
-    @podcasts = add_sorting(base_query).filter_by_title(params[:q])
+    filtered_podcasts = base_query.filter_by_title(params[:q])
+    @podcasts = add_sorting(filtered_podcasts)
```
cavis, line 16: Noting this change is functionally equivalent to what was there before.  (Calling both `add_sorting` and `.filter_by_title` on the base query, assigning the result to `@podcasts`).  But this is fine, if you think it's more readable this way.

app/views/podcasts/index.html.erb
```diff
@@ -6,15 +6,23 @@
...
+          <% else %>
+            <span class="material-icons" aria-hidden="true">timer</span>
+          <% end %>
         </button>
         <ul class="dropdown-menu">
-          <li><%= link_to "Recent Activity", podcasts_path(sort: "Recent"), class: "dropdown-item active" %></li>
-          <li><%= link_to "# of Episodes", podcasts_path(sort: "episode_count"), class: "dropdown-item" %></li>
-          <li><%= link_to "A-Z", podcasts_path(sort: "A-Z"), class: "dropdown-item" %></li>
-          <li><%= link_to "Z-A", podcasts_path(sort: "Z-A"), class: "dropdown-item" %></li>
+        <li><%= active_link_to "Recent Activity", podcasts_path(sort: "Recent Activity"), active: :exact, class: "dropdown-item text-decoration-none" %></li>
```
cavis, line 22: I think these should stay indented, inside the `<ul>`?

cavis (approved):
One nitpick on indentation.  But otherwise looks good to me.

Should probably close #597 and mark this as a non-draft?  I got them mixed up since this one is still a draft.

app/controllers/podcasts_controller.rb
a-maci29: The latter way would maybe be easier for me to come back and understand what's happening, but I'm not insistent about it. I can change it back to the way it was before.

app/views/podcasts/index.html.erb
a-maci29: Huh. It's indented in my VS code. Would linter change that??

cavis (approved):
👍 looks good to me!

Issue #606, opened 2023-03-28
Podcast Planner help text and verbiage sweep


brandonhundt:
I think the podcast planner needs a verbiage sweep. The legacy help text isn't great and needs improvement and some of the field names could be clearer. 

Also, we may be able to use the new help text style for some of this.

PR #628, opened 2023-04-27
Podcast Planner verbiage and layout

brandonhundt:
closes: #606 

Now that we have a working calendar (!!!EXCITING!!!) I thought we could take a pass at the verbiage and the layout of the calendar. This PR's changes will get us closer to what exists in Publish currently where we've received a lot of neutral to positive feedback. (NOTE: I think we can open a new ticket to explore enhancements to this feature that go beyond the existing functionality in this PR.)

There were a lot of places where we had help text, headers, and field labels that felt a little redundant or overly technical in its use, so I attempted to clean It up.

Also, I reverted this to a single column layout like we have now. I think this will serve our producers finely enough until we get the time to really rethink the feature in the future.

### DIFF
--- a/app/assets/stylesheets/app/calendar.scss
+++ b/app/assets/stylesheets/app/calendar.scss
@@ -19,8 +19,9 @@
     }
 
     .day.current-month:hover {
-      background-color: $secondary !important;
+      background-color: $primary !important;
       color: $light;
+      cursor: pointer;
     }
   }
 }
--- a/app/helpers/podcast_planner_helper.rb
+++ b/app/helpers/podcast_planner_helper.rb
@@ -61,7 +61,6 @@ module PodcastPlannerHelper
     if date_is_in_month?(day, month)
       data[:controller] = DATE_CONTROLLER
       data[:action] = [TOGGLE_ACTION, RECOUNT_ACTION].join(" ")
-      data["count-target"] = "counter"
     end
 
     content_tag(:td, class: calendar.td_classes_for(day), data: data) do
--- a/app/helpers/podcast_planner_helper.rb
+++ b/app/helpers/podcast_planner_helper.rb
@@ -5,6 +5,7 @@ module PodcastPlannerHelper
   MONTHLY_WEEKS = I18n.t([:first, :second, :third, :fourth, :fifth], scope: [:podcast_planner, :helper, :monthly_options])
   DATE_CONTROLLER = "date"
   TOGGLE_ACTION = "click->date#toggleSelect"
+  RECOUNT_ACTION = "click->count#recount"
 
   def day_options
     DateTime::DAYNAMES.map.with_index { |day, i| [day, i] }
@@ -59,7 +60,8 @@ module PodcastPlannerHelper
     data = {}
     if date_is_in_month?(day, month)
       data[:controller] = DATE_CONTROLLER
-      data[:action] = TOGGLE_ACTION
+      data[:action] = [TOGGLE_ACTION, RECOUNT_ACTION].join(" ")
+      data["count-target"] = "counter"
     end
 
     content_tag(:td, class: calendar.td_classes_for(day), data: data) do
--- a/app/javascript/controllers/count_controller.js
+++ b/app/javascript/controllers/count_controller.js
@@ -3,7 +3,7 @@ import { Controller } from "@hotwired/stimulus"
 export default class extends Controller {
   static targets = ["counter", "display"]
   static values = {
-    selected: { type: Number, default: 0 }
+    selected: { type: Number, default: 0 },
   }
 
   connect() {
@@ -17,7 +17,7 @@ export default class extends Controller {
   }
 
   countSelectedTargets() {
-    const selectedTargets = this.counterTargets.filter(el => {
+    const selectedTargets = this.counterTargets.filter((el) => {
       return el.firstElementChild.disabled
     })
     return this.counterTargets.length - selectedTargets.length
--- a/app/javascript/controllers/count_controller.js
+++ b/app/javascript/controllers/count_controller.js
@@ -2,24 +2,15 @@ import { Controller } from "@hotwired/stimulus"
 
 export default class extends Controller {
   static targets = ["counter", "display"]
-  static values = {
-    selected: { type: Number, default: 0 },
-  }
-
-  connect() {
-    this.selectedValue = this.countSelectedTargets()
-    this.displayTarget.innerHTML = this.selectedValue
-  }
 
   recount() {
-    this.selectedValue = this.countSelectedTargets()
-    this.displayTarget.innerHTML = this.selectedValue
+    this.displayTarget.innerHTML = this.countSelectedTargets()
   }
 
   countSelectedTargets() {
     const selectedTargets = this.counterTargets.filter((el) => {
-      return el.firstElementChild.disabled
+      return !el.disabled
     })
-    return this.counterTargets.length - selectedTargets.length
+    return selectedTargets.length
   }
 }
--- a/app/javascript/controllers/count_controller.js
+++ b/app/javascript/controllers/count_controller.js
@@ -1,0 +1,25 @@
+import { Controller } from "@hotwired/stimulus"
+
+export default class extends Controller {
+  static targets = ["counter", "display"]
+  static values = {
+    selected: { type: Number, default: 0 }
+  }
+
+  connect() {
+    this.selectedValue = this.countSelectedTargets()
+    this.displayTarget.innerHTML = this.selectedValue
+  }
+
+  recount() {
+    this.selectedValue = this.countSelectedTargets()
+    this.displayTarget.innerHTML = this.selectedValue
+  }
+
+  countSelectedTargets() {
+    const selectedTargets = this.counterTargets.filter(el => {
+      return el.firstElementChild.disabled
+    })
+    return this.counterTargets.length - selectedTargets.length
+  }
+}
--- a/app/views/podcast_planner/_calendar.html.erb
+++ b/app/views/podcast_planner/_calendar.html.erb
@@ -1,4 +1,4 @@
-<div class="col <%= "d-none" if index >= range %>" data-pager-target="page">
+<div class="col col-xl-4 <%= "d-none" if index >= range %>" data-pager-target="page">
   <%= month_calendar(start_date: start_date) do |date| %>
     <% if date_is_in_month?(date, start_date.month) %>
       <%= date.day %>
--- a/app/views/podcast_planner/_calendar.html.erb
+++ b/app/views/podcast_planner/_calendar.html.erb
@@ -3,7 +3,7 @@
     <% if date_is_in_month?(date, start_date.month) %>
       <%= date.day %>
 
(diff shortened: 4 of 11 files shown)


app/views/podcast_planner/_form_date_calculator.html.erb
```diff
@@ -1,12 +1,6 @@
 <section>
   <div class="mb-4">
-    <h2 class="fw-bold"><%= t(".header.date_calculation") %></h2>
-    <p><%= t(".help.date_calculation") %></p>
-  </div>
-
-  <div class="mb-4">
-    <h3 class="fw-bold"><%= t(".header.day_select") %></h3>
-    <p><%= t(".help.day_select") %></p>
+    <h3 class="mb-3"><%= t(".header.day_select") %></h3>
```
brandonhundt, line 3: Example of simplifying the text presented to the user.

app/views/podcast_planner/_form_date_calculator.html.erb
```diff
@@ -46,25 +36,26 @@
...
-        <p><%= t(".help.select_range") %></p>
-          <div class="form-floating mb-2">
-            <%= form.date_field :start_date, min: DateTime.tomorrow, max: DateTime.tomorrow + 1.years, value: DateTime.tomorrow, data: {action: "click#submit"} %>
-            <%= form.label :start_date %>
-          </div>
+        <div class="form-floating mb-2">
+          <%= form.date_field :start_date, min: DateTime.tomorrow, max: DateTime.tomorrow + 1.years, value: DateTime.tomorrow, data: {action: "click#submit"} %>
+          <%= form.label :start_date %>
+        </div>
+      </div>
```
brandonhundt, line 56: I adjusted the layout of this field so that the selector field is full width, with two variable fields appearing below. This is more consistent with our UI patterns.

app/views/podcast_planner/_form_main.html.erb
```diff
@@ -8,7 +8,7 @@
       }
     ) do |form| %>
     <h1 class="text-black fw-bold d-flex flex-fill"><%= t(".header") %></h1>
-    <p><%= t(".help_html", url: new_podcast_episode_path(@podcast)) %></p>
+    <p class="lh-sm"><%= t(".help_html", url: new_podcast_episode_path(@podcast)) %></p>
```
brandonhundt, line 11: Line-height was really tall, looked weird.

app/views/podcast_planner/_form_side.html.erb
```diff
@@ -1,56 +1,51 @@
...
+  method: :post
+) do |form| %>
+  <div class="mb-4">
+    <h3 class="mb-3"><%= t(".header.publish_time") %></h3>
+    <div class="col-lg-12">
+      <div class="form-floating">
+        <%= form.select :publish_time, time_options %>
+        <%= form.label :publish_time %>
+      </div>
+    </div>
```
brandonhundt, line 12: This file was reworked since its no longer technically a sidebar card

brandonhundt: Functionally there shouldn't be any different. All presentational here.

config/locales/en.yml
```diff
@@ -230,13 +230,13 @@ en:
       success: Episode drafts generated.
     form_main:
       header: Plan Episodes
-      help_html: Save yourself some time by automatically creating multiple episode drafts to create a production schedule, it may also help your revenue as our sales teams will have access to dropdates to sell against. If you only wish to add a single episode, <a href="%{url}">go here</a>.
+      help_html: Add multiple episode drafts to create a production schedule. Episode drafts improve the accuracy of your podcast's forecasts. If you only need to add a single episode, <a href="%{url}">go here</a>.
```
brandonhundt, line 233: A verbiage review in this file is very appreciated!!!

app/views/podcast_planner/_form_side.html.erb
```diff
@@ -1,56 +1,51 @@
...
-                <% 24.times do |i| %>
-                  <%= render "calendar", form: form, start_date: Date.today.beginning_of_month + i.months, planner: @planner, index: i, range: 6 %>
-                <% end %>
-              </div>
-            <% end %>
-          </div>
+  <div class="card p-4 mt-4mb-4">
+    <div data-controller="pager" data-pager-range-value="6">
+      <div class="d-flex justify-content-between">
+        <h3 class="mb-3"><%= t(".header.number_of_drafts") %></h3>
```
radical-ube, line 28: this needs a `count` variable to pass through to the yml file in order to display the correct number.

you can probably pass through `@planner.dates.count` but that would only reflect the dates generated through the calculator, not selected manually. i could probably figure out a helper method to add to the stimulus controller that would update every time a manual date has been clicked as well. is this an important feature for parity?

radical-ube (commented):
in general i have no comments on verbiage since this is a return to parity verbiage and we're not planning on the improvements until post-launch. the only comment i made was regarding the `count` variable for that one header label. the only other comment i would make is figuring out a color key since that is one thing new to the calendar.

radical-ube (changes requested):
should have put previous comment under request changes for the `count` variable.

app/helpers/podcast_planner_helper.rb
```diff
@@ -5,6 +5,7 @@ module PodcastPlannerHelper
   MONTHLY_WEEKS = I18n.t([:first, :second, :third, :fourth, :fifth], scope: [:podcast_planner, :helper, :monthly_options])
   DATE_CONTROLLER = "date"
   TOGGLE_ACTION = "click->date#toggleSelect"
+  RECOUNT_ACTION = "click->count#recount"
```
radical-ube, line 8: this and the `counterTarget` below are additions for the new stimulus `count_controller`. Essentially adds a counter to each date. the controller counts the ones that are selected.

app/javascript/controllers/count_controller.js
```diff
@@ -0,0 +1,25 @@
...
+    this.selectedValue = this.countSelectedTargets()
+    this.displayTarget.innerHTML = this.selectedValue
+  }
+
+  recount() {
+    this.selectedValue = this.countSelectedTargets()
+    this.displayTarget.innerHTML = this.selectedValue
+  }
+
+  countSelectedTargets() {
```
radical-ube, line 19: i wasn't quite sure how to generalize this action, so for now it is reading the `disabled` property on the inputs within the dates. and then calculating which dates are selected by determining which have been enabled.

config/locales/en.yml
```diff
@@ -248,16 +248,19 @@ en:
...
       header:
-        preview: Schedule Preview
-        publish_time: Publish Time
-        audio_segments: Audio Segments
-        number_of_drafts: "Number of Draft Episodes: %{count}"
+        preview: Preview Draft Schedule
+        publish_time: What time of day do you plan to publish the episodes?
+        audio_segments: How many audio segments?
+        segment_count: Number of segments
+        number_of_drafts: "Drafts planned: "
```
radical-ube, line 261: changed this so it no longer needs to pass in a `count` variable. the count is displayed in the view in the innerHTML of a new span in the h3 that uses this label.

app/views/podcast_planner/_form_side.html.erb
```diff
@@ -1,56 +1,53 @@
...
-            <div class="form-floating">
-              <%= form.number_field :segment_count, {min: 1, max: 10, step: 1, value: 2} %>
-              <%= form.label :segment_count %>
-            </div>
+  <div class="card p-4 mt-4mb-4">
+    <%= turbo_frame_tag "preview", target: "_top" do %>
+      <div data-controller="pager count" data-pager-range-value="6">
+        <div class="d-flex justify-content-between">
+          <h3 class="mb-3"><%= t(".header.number_of_drafts") %>
+            <span data-count-target="display"></span>
```
radical-ube, line 30: new span here. the innerHTML of this will update with the number of selected items, as determined by the new `count_controller`.

app/javascript/controllers/count_controller.js
cavis: It is a little awkward to not just count the `counterTargets`.

Could you instead add this to the `form.hidden_field` in `_calendar.html.erb`?  Would basically look the same, except getting rid of the `firstElementChild`.  Maybe a bit cleaner?

Or it's nearly the same - your call.

app/javascript/controllers/count_controller.js
```diff
@@ -0,0 +1,25 @@
...
+    const selectedTargets = this.counterTargets.filter((el) => {
+      return el.firstElementChild.disabled
+    })
+    return this.counterTargets.length - selectedTargets.length
```
cavis, line 23: Naming is a bit misleading ... `selectedTargets` appears to be all the _deselected_ hidden fields?  Maybe invert the above filter to `!disabled`, and then just return `.length`?

app/views/podcast_planner/_form_side.html.erb
cavis: I would probably render a number in here to start?  So it has the correct value in HTML, before javascript even takes a pass at it.

app/javascript/controllers/count_controller.js
```diff
@@ -0,0 +1,25 @@
+import { Controller } from "@hotwired/stimulus"
+
+export default class extends Controller {
+  static targets = ["counter", "display"]
+  static values = {
+    selected: { type: Number, default: 0 },
```
cavis, line 6: Is this even being passed in anywhere, or are you getting the default?

I'd probably suggest omitting this entirely.  The server should render the correct number in the HTML to start with.  And then drop the `connect()` method here, and just wait until some action calls `recount()` to have this controller modify the innerHTML.

cavis (changes requested):
Couple minor suggestions @radical-ube.  But all pretty much your call - they're slightly stylistic.

@brandonhundt one slight issue on wide viewports 
image

It isn't going to be trivial to display 8 calendars magically as the user widens the viewport.  So may just want to restrict the max-width of those for now?  Or force maximum 3-per-row?

app/helpers/podcast_planner_helper.rb
radical-ube: moved `counterTarget` to the input in the `_calendar` partial.

radical-ube (approved):

brandonhundt:
@ube @ryan I've fixed the grid issue so that on the largest screens it will always display in 3 columns.

Also added a key for the calendar, and tweaked the hover state on the calendar dates.

Screenshot 2023-05-03 at 1 02 18 PM

app/assets/stylesheets/app/calendar.scss
```diff
@@ -19,8 +19,9 @@
     }
 
     .day.current-month:hover {
-      background-color: $secondary !important;
+      background-color: $primary !important;
       color: $light;
+      cursor: pointer;
```
brandonhundt, line 24: to let the user know they can click to add dates. chose the link color to emphasize clicking.

app/views/podcast_planner/_calendar.html.erb
```diff
@@ -1,9 +1,9 @@
-<div class="col <%= "d-none" if index >= range %>" data-pager-target="page">
+<div class="col col-xl-4 <%= "d-none" if index >= range %>" data-pager-target="page">
```
brandonhundt, line 1: Thats all it was to fix the grid issue.

cavis (approved):
👍

PR #629, opened 2023-04-28
Media status handling

cavis:
🚧 WIP - some pieces of #618.

### DIFF
--- a/app/helpers/uploads_helper.rb
+++ b/app/helpers/uploads_helper.rb
@@ -24,6 +24,21 @@ module UploadsHelper
     params
   end
 
+  def uploads_retry_params(form1, form2 = nil)
+    params = {}
+    params["#{form1.object_name}[id]"] = form1.object.id
+    params["#{form2.object_name}[id]"] = form2.object.id if form2
+
+    # retry only the right-most form object
+    if form2
+      params["#{form2.object_name}[_retry]"] = "1"
+    else
+      params["#{form1.object_name}[_retry]"] = "1"
+    end
+
+    params
+  end
+
   private
 
   def uploads_prefix
--- a/app/javascript/controllers/disable_controller.js
+++ b/app/javascript/controllers/disable_controller.js
@@ -1,0 +1,39 @@
+import { Controller } from "@hotwired/stimulus"
+
+export default class extends Controller {
+  connect() {
+    this.element.dataset.action = `${this.element.dataset.action || ""} submit->disable#disableForm`
+    this.uploadCount = 0
+  }
+
+  disableForm() {
+    this.submitButtons().forEach((el) => {
+      el.disabled = true
+      el.value = el.dataset.disableWith || el.value
+    })
+  }
+
+  uploading(disable = true) {
+    if (disable) {
+      this.uploadCount++
+    } else {
+      this.uploadCount--
+    }
+    this.submitButtons().forEach((el) => {
+      if (this.uploadCount > 0) {
+        el.disabled = true
+        el.dataset.disableOriginal = el.dataset.disableOriginal || el.value
+        el.value = el.dataset.uploadWith || el.value
+      } else {
+        el.disabled = false
+        el.value = el.dataset.disableOriginal
+      }
+    })
+  }
+
+  submitButtons() {
+    const submits1 = this.element.querySelectorAll('input[type="submit"]')
+    const submits2 = this.element.id ? document.querySelectorAll(`input[type="submit"][form="${this.element.id}"]`) : []
+    return Array.from(submits1).concat(Array.from(submits2))
+  }
+}
--- a/app/javascript/controllers/upload_controller.js
+++ b/app/javascript/controllers/upload_controller.js
@@ -4,6 +4,7 @@ import SparkMD5 from "spark-md5"
 import sha256 from "sha256"
 
 export default class extends Controller {
+  static outlets = ["disable"]
   static targets = [
     "field",
     "originalUrl",
@@ -140,6 +141,9 @@ export default class extends Controller {
     toggle(this.progressTargets, type === "progress")
     toggle(this.successTargets, type === "success")
     toggle(this.errorTargets, type === "error")
+
+    // disable form submit while upload in progress
+    this.disableOutlet.uploading(type === "progress")
 
     // disable id/_destroy fields if we're replacing that file
     this.replaceTargets.forEach((el) => (el.disabled = type === "success"))
--- a/app/models/media_resource.rb
+++ b/app/models/media_resource.rb
@@ -101,4 +101,17 @@ class MediaResource < ApplicationRecord
   def update_resource(res)
     # NOTE: media_resources have no user settable fields
   end
+
+  def retryable?
+    status_processing? && (Time.now - updated_at) > 30
+  end
+
+  def retry!
+    status_retrying!
+    copy_media(true)
+  end
+
+  def _retry=(_val)
+    retry!
+  end
 end
--- a/app/views/images/_processing.html.erb
+++ b/app/views/images/_processing.html.erb
@@ -15,8 +15,20 @@
         </div>
 
         <div class="text-secondary mx-2"><%= image.file_name %></div>
-        <small class="text-muted flex-grow-1">(Processing)</small>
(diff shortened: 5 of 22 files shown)


app/models/media_resource.rb
```diff
@@ -101,4 +101,29 @@ def replace?(res)
...
+      false
+    end
+  end
+
+  def retry!
+    status_retrying!
+    copy_media(true)
+  end
+
+  def _retry=(_val)
```
a-maci29, line 118: What does the underscore before the method name here do?

cavis: I was just following a convention from rails forms/formbuilder, where you use `_destroy` to magically mark a record for destruction.  But here, I'm magically marking this file's processing to get retried.

app/views/images/_processing.html.erb
```diff
@@ -15,8 +15,20 @@
         </div>
 
         <div class="text-secondary mx-2"><%= image.file_name %></div>
-        <small class="text-muted flex-grow-1">(Processing)</small>
-        <small class="text-muted">(<%= number_to_human_size(image.size) %>)</small>
+
+        <% if image.retryable? %>
+          <div class="flex-grow-1">
+            <%= link_to retry_path, class: "btn btn-sm btn-warning" do %>
```
a-maci29, line 21: Trying to follow the flow of this retry path, and want to describe what we believe what might be happening:

- We looked at the `episodes_helper`, and found the `episode_retry_image_path`, which sets this path based on whether it's a new upload, or editing an upload
- It also passes some sort of `params[:_retry] = "1"` (meaning `true`) for this retry param value based on which form is being retried

cavis: Yep - the whole goal is to generate a form `params` something like:

```
podcast[default_feed_attributes][id]: 123
podcast[default_feed_attributes][itunes_images_attributes][0][_retry]: 1
podcast[default_feed_attributes][itunes_images_attributes][0][id]: 456
```

So when rails parses the params, and you call `podcast.assign_attributes(podcast_params)` ... the `accepts_nested_attributes_for` mechanics in the models traverse through, and call `record._retry = 1` on the itunes image.

app/helpers/uploads_helper.rb
```diff
@@ -24,6 +24,21 @@ def uploads_destroy_params(form1, form2 = nil)
     params
   end
 
+  def uploads_retry_params(form1, form2 = nil)
+    params = {}
+    params["#{form1.object_name}[id]"] = form1.object.id
+    params["#{form2.object_name}[id]"] = form2.object.id if form2
+
+    # retry only the right-most form object
```
radical-ube, line 32: what is the case that there are 2 forms, and what constitutes the right-most form?

app/javascript/controllers/disable_controller.js
```diff
@@ -0,0 +1,39 @@
+import { Controller } from "@hotwired/stimulus"
+
+export default class extends Controller {
+  connect() {
+    this.element.dataset.action = `${this.element.dataset.action || ""} submit->disable#disableForm`
```
radical-ube, line 5: i see that this controller is added to `_form`. is there a reason for adding the action here instead of where the controller is defined? is this because the `dataset` may not be present yet or something? (in the case of a new upload, for example?)

app/javascript/controllers/disable_controller.js
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  disableForm() {
+    this.submitButtons().forEach((el) => {
+      el.disabled = true
+      el.value = el.dataset.disableWith || el.value
+    })
+  }
+
+  uploading(disable = true) {
```
radical-ube, line 16: ah i didn't understand `uploadCount` at first but i think this makes it clear. when an upload begins (which i assume has an action that triggers this method), we tick the count up. and because there can be multiple uploads in progress at the same time, we need to keep track so that even if one finishes early, we don't automatically enable all the submit buttons in case there is still another upload in progress.

app/javascript/controllers/disable_controller.js
```diff
@@ -0,0 +1,39 @@
...
+    if (disable) {
+      this.uploadCount++
+    } else {
+      this.uploadCount--
+    }
+    this.submitButtons().forEach((el) => {
```
radical-ube, line 22: is there a difference between when this loop runs and when `disableForm` runs?

app/javascript/controllers/upload_controller.js
```diff
@@ -141,6 +142,9 @@ export default class extends Controller {
     toggle(this.successTargets, type === "success")
     toggle(this.errorTargets, type === "error")
 
+    // disable form submit while upload in progress
+    this.disableOutlet.uploading(type === "progress")
```
radical-ube, line 146: i have only been reading about `outlets` but this is my first encounter of it in practice. so in this example, when an upload calls this `show` method, we also want to call the `uploading` method that is in the `disable_controller`?

app/helpers/uploads_helper.rb
cavis: So this was a gotcha that took me ~way too long~ a normal amount of time to track down.

So I'm trying to generate form params nested 2 levels in.  Inside a `fields_for :default_feed` _AND_ inside that a `fields_for :itunes_image`.

If I only use the last (itunes_image) form to generate the params, I'd get:

```
podcast[default_feed_attributes][itunes_images_attributes][0][_retry]: 1
podcast[default_feed_attributes][itunes_images_attributes][0][id]: 456
```

And Rails would fail to pass that param, with something like "can't find itunes image with id 456".  The trick was: Rails requires _both_ the feed's ID, and the itunes_image ID to do its magic `accepts_nested_attributes_for`.  So using both forms things work!

```
podcast[default_feed_attributes][id]: 123
podcast[default_feed_attributes][itunes_images_attributes][0][_retry]: 1
podcast[default_feed_attributes][itunes_images_attributes][0][id]: 456
```

app/javascript/controllers/upload_controller.js
cavis: You got it.  "Outlets" are a way for controllers to communicate.  In this case, the UploadController tells the DisableController "hey - an upload started ... increment your count of uploads-in-progress".

app/javascript/controllers/disable_controller.js
cavis: So this is a stimulus trick I've seen in the examples lately...

Instead of this:

```html
<form data-controller="disable" data-action="submit->disable#disable">
```

We can just write:

```html
<form data-controller="disable">
```

And then on `connect()` ... the disable controller adds that action to the controller itself.  Ends up being the same as the 1st example, saving a few characters.  And you don't have to mess with `addEventListener` / `removeEventListener` manually.  Stimulus still handles it.

cavis: (And yes, I'm trying to handle the case where the DOM already has some other action in it ... I don't want to nuke that, adding myself to the data-actions).

app/javascript/controllers/disable_controller.js
cavis: Exactly

app/javascript/controllers/disable_controller.js
cavis: `disableForm` runs when the `<form>` itself submits.  So it disables the submit button, preventing users from double-clicking / double-saving.  But a save is already in-progress.  If you throw a `sleep 2` in the controller `#update`, you'll see it.

TMI: [Rails UJS](https://github.com/rails/rails-ujs) typically handles this in Rails apps.  But something appears incompatible/broken with Rails 7... and I really wanted the `Save` button to change to `Saving...` because I'm picky.

app/javascript/controllers/upload_controller.js
cavis: I messed around with some other ways of accomplishing this... something like...

```html
<div data-disable-target="uploading">
```

And then in the `disable_controller.js`, something like:

```js
uploadingTargetConnected(element) {
  // increment and disable buttons
}

uploadingTargetDisconnected(element) {
  // decrement and enable buttons
}
```

But that was a lot more lines of code.  And I hit a bunch of gotchas... I was having to make the `upload_controller.js` manually add/remove `el.dataset.disable_target = "uploading"`.  The Outlet turned out to be cleaner for 2 JS controllers communicating.

PR #630, opened 2023-04-28
Polling and dropdowns

cavis:
Part of #629.

When audio/images were processing, we were clicking a link within a turbo-frame, to poll for changes.  Every time this happened, the slim-select `document.onClick` listener would see a click even occurred outside of the dropdown, and close the dropdown.

This switches to using the parent turbo-frame ["src" property](https://turbo.hotwired.dev/reference/frames#properties) instead.  (With a graceful fallback to clicking the link).

Also fixed a bug with the metadata-modals ... branches got a bit out of sync there.

### DIFF
--- a/app/assets/stylesheets/shared/form.scss
+++ b/app/assets/stylesheets/shared/form.scss
@@ -165,6 +165,10 @@
     // border: none;
   }
 
+  .ss-values .ss-single {
+    margin-left: 0;
+  }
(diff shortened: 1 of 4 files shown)


kookster (approved):
lgtm

kookster:
change makes sense, and the bug fix is small 👍

PR #631, opened 2023-05-01
Re-copying complete media

cavis:
Found a bunch of old files in prod, with `status: "complete"` but missing a bunch of metadata fields.  I'm guessing they just predate when we added the metadata to the DB, or something was missing in the Fixer callbacks?

Anyways, I need to fix them, but their original urls are long gone.

Generally, when re-copying media (for files that already succeeded) - we can and should use the `f.prxu.org` url as the source.  Instead of the `original_url`, which in many cases is long gone/broken.

### DIFF
--- a/app/models/concerns/image_file.rb
+++ b/app/models/concerns/image_file.rb
@@ -79,7 +79,7 @@ module ImageFile
   end
 
   def href
-    status_complete? ? url : original_url
+    (status_complete? || status_invalid?) ? url : original_url
   end
 
   def href=(h)
--- a/app/models/tasks/copy_media_task.rb
+++ b/app/models/tasks/copy_media_task.rb
@@ -22,7 +22,7 @@ class Tasks::CopyMediaTask < ::Task
   end
 
   def source_url(media_resource)
-    media_resource.original_url.sub(/\?.*$/, "")
+    media_resource.href.sub(/\?.*$/, "")
   end
 
   def destination_url(media_resource)
--- a/test/models/tasks/copy_image_task_test.rb
+++ b/test/models/tasks/copy_image_task_test.rb
@@ -8,8 +8,16 @@ describe Tasks::CopyImageTask do
   it "has task options" do
     opts = task.task_options
     assert_equal opts[:job_type], "image"
-    assert_equal opts[:source], image.original_url
+    assert_equal opts[:source], image.href
     assert_match(/s3:\/\/test-prx-feed\/#{path}\/ba047dce-9df5-4132-a04b-31d24c7c55a(\d+)\/images\/4e745a8c-77ee-481c-a72b-fd868dfd1c9(\d+)\/image\.png/, opts[:destination])
+  end
+
+  it "uses the url as source when complete" do
+    assert_equal image.status, "complete"
+    assert_equal task.task_options[:source], image.url
+
+    image.status = "started"
+    assert_equal task.task_options[:source], image.original_url
   end
 
   it "gets the image path" do
--- a/test/models/tasks/copy_media_task_test.rb
+++ b/test/models/tasks/copy_media_task_test.rb
@@ -16,14 +16,14 @@ describe Tasks::CopyMediaTask do
     assert_equal options.keys, %w[callback job_type source destination]
     assert_match(/^sqs:\/\//, options[:callback])
     assert_equal options[:job_type], "audio"
-    assert_equal options[:source], "s3://prx-testing/test/audio.mp3"
+    assert_equal options[:source], task.media_resource.href
     assert_match(/^s3:\/\//, options[:destination])
   end
 
   it "remove query string from audio url" do
     refute_nil task.media_resource
-    original = task.media_resource.original_url
-    task.media_resource.original_url = "#{original}?remove=this"
+    original = task.media_resource.href
+    task.media_resource.href = "#{original}?remove=this"
     assert_equal task.task_options[:source], original
   end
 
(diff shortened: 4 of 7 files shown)


app/models/tasks/copy_media_task.rb
```diff
@@ -22,7 +22,7 @@ def task_options
   end
 
   def source_url(media_resource)
-    media_resource.original_url.sub(/\?.*$/, "")
+    media_resource.href.sub(/\?.*$/, "")
```
cavis, line 25: So this "strip query params" was added in #135.  I'm not entirely clear if/why we'd still need it?  But left it in for now.

svevang (approved):

app/models/concerns/image_file.rb
```diff
@@ -79,7 +79,7 @@ def url
   end
 
   def href
-    status_complete? ? url : original_url
+    (status_complete? || status_invalid?) ? url : original_url
   end
 
   def href=(h)
```
a-maci29, line 85: `href=(h)`? What does the `=(h)` syntax do on a method definition?

app/models/tasks/copy_media_task.rb
a-maci29: Biggo Shruggo.
What could potentially happen if it's removed?

app/models/concerns/image_file.rb
cavis: That's how you define a "setter method" in Ruby.  `href` isn't actually the name of any database field.  But you can call both:

```ruby
# returns some string
image.href

# set it to something
image.href = "foo"
```

The first case calls the method up on line 81.

In that 2nd case, it calls this method, with `h` set to the string `"foo"`.

test/models/tasks/copy_image_task_test.rb
```diff
@@ -8,10 +8,18 @@
   it "has task options" do
     opts = task.task_options
     assert_equal opts[:job_type], "image"
-    assert_equal opts[:source], image.original_url
+    assert_equal opts[:source], image.href
```
a-maci29, line 11: What does `opts` do? It seems like it's coming from porter.

a-maci29: Maybe I'm making that up.

app/models/tasks/copy_media_task.rb
a-maci29: Probably better to leave it in then, if it's not hurting anything and adds some use.

test/models/tasks/copy_image_task_test.rb
cavis: I believe this what we're sending _to_ Porter.

Like previously, it was the `image.original_url` (database field).  And Porter would copy that.

But now it's the `image.href` ... which is a method, returning either the `original_url` _OR_ the image's final CDN `http://` location.  Based on what the images status is.

app/models/concerns/image_file.rb
```diff
@@ -79,7 +79,7 @@ def url
   end
 
   def href
-    status_complete? ? url : original_url
+    (status_complete? || status_invalid?) ? url : original_url
```
a-maci29, line 82: Checking for understanding: This is making sure that if either `status_complete` or `status_invalid` are valid, it goes on to hit `url`, checking first for the `f.prxu.org` as the source, but if not, will fall back on the `original_url`.

cavis: You got it -

If the status is `"complete"` or `"invalid"`, we know that Porter succeeded, and `"https://f.prxu.org/..."` (the string returned by `url`) will return the file.

If the status is something else (error, processing, started, etc) ... Porter hasn't managed to copy the file to `https://f.prxu.org/...` for some reason.  And trying to GET it will 404.  So we should use the `original_url`.

test/models/tasks/copy_image_task_test.rb
a-maci29: I only just now realize that opts was defined here.

test/models/tasks/copy_media_task_test.rb
```diff
@@ -16,14 +16,14 @@
     assert_equal options.keys, %w[callback job_type source destination]
     assert_match(/^sqs:\/\//, options[:callback])
     assert_equal options[:job_type], "audio"
-    assert_equal options[:source], "s3://prx-testing/test/audio.mp3"
+    assert_equal options[:source], task.media_resource.href
```
a-maci29, line 19: I see that instead of a direct file path, there's a change to using `task.media_resource.href` as a source. I have an idea of why, but could you explain it with some more detail?
(Is `s3` linking a path to the s3 bucket?)

test/models/tasks/copy_media_task_test.rb
```diff
@@ -45,8 +45,12 @@
     assert_equal url, "s3://test-prx-feed/path/guid/audio.mp3"
   end
 
-  it "use original url as the source url" do
+  it "use original url as the source url until complete" do
+    task.media_resource.status = "started"
```
a-maci29, line 49: I want to say that this would use the `original_url` until the actual process of re-evaluating the `task.media_resource` is finished so it can decide if it wants to use that `url` first as the source before falling back to the `original_url`.

test/models/tasks/copy_media_task_test.rb
cavis: Previously, we were alway telling Porter to use the `:source` of the `original_url`.

But I believe in this test, this specific audio (the `task.media_resource` the in the test factory) is in "completed" status.  So instead of the `original_url` (that s3:// file)... this is going to be the `url`.

Seemed better to test for what the `:source` is actually being set to now ... which is whatever `.href` returns.

PR #638, opened 2023-05-04
Feat/edit podcast feed

radical-ube:
this is part of #535 but does not close it.

I believe I've reached a stage of the PR that can be considered ready for review, although the ticket is not complete. This PR includes `update` functionality for most of the straight-forward fields: 
- :file_name,
- :slug,
- :title,
- :subtitle,
- :description,
- :summary,
- :private,
- :url,
- :new_feed_url,
- :enclosure_prefix,
- :display_episodes_count,
- :display_full_episodes_count,
- :episode_offset_seconds

as well as `update` functionality for Auth Tokens.

TODO:
- `update` functionality for Ad Zones
- `update` functionality for Audio Format
- `create` functionality for New Feed
- turbo functionality for private -> Auth Token display
- `create` functionality for Auth Tokens
- Feed Images


app/controllers/feeds_controller.rb
```diff
@@ -1,25 +1,17 @@
...
   before_action :set_podcast
 
-  # GET /feeds
-  def index
-    @feeds = Feed.all.limit(10)
-  end
-
   # GET /feeds/1
   def show
+    @custom_feeds = @podcast.feeds.custom
```
radical-ube, line 7: are non-default feeds named something specific?

app/models/feed.rb
```diff
@@ -43,6 +44,7 @@ class Feed < ApplicationRecord
   before_validation :sanitize_text
 
   scope :default, -> { where(slug: nil) }
+  scope :custom, -> { where.not(slug: nil) }
```
radical-ube, line 47: any other attribute signifiers for non-default feeds? or is it just opposite of default?

app/views/feeds/_form_main.html.erb
```diff
@@ -0,0 +1,203 @@
...
+<div class="col-12 mb-4 prx-field-group">
+  <div class="form-floating">
+    <%= form.text_field :file_name %>
+    <%= form.label :file_name %>
+  </div>
+
+  <button type="button" class="prx-btn-help" data-bs-toggle="popover" data-bs-content="<%= t(".help.file_name") %>"><span class="material-icons">help</span></button>
+</div>
+
+<% unless feed.default? %>
```
radical-ube, line 21: wondering if this chunk of fields should be abstracted into its own partial? or if it makes sense to stay here?

app/views/feeds/_form_main.html.erb
```diff
@@ -0,0 +1,203 @@
...
+      <div class="form-floating">
+        <%= form.text_field :title %>
+        <%= form.label :title %>
+      </div>
+    </div>
+
+    <div class="mb-4 prx-field-group">
+      <div class="form-floating">
+        <%= form.text_field :subtitle %>
+        <%= form.label :subtitle %>
```
radical-ube, line 47: `subtitle` and `summary` weren't part of the Feeder design so i wasn't sure if i needed to bring them over from Publish.

app/views/feeds/_form_main.html.erb
```diff
@@ -0,0 +1,203 @@
...
+    </div>
+  </div>
+
+  <div class="col-12">
+    <hr>
+    <h3><%= t(".headers.private") %></h3>
+    <div class="prx-field-group">
+      <div>
+        <%= form.check_box :private %>
+        <%= form.label :private %>
```
radical-ube, line 78: TODO: link this check box to displaying the below Auth Tokens partial using turbo

app/views/feeds/_form_main.html.erb
```diff
@@ -0,0 +1,203 @@
...
+    <%= render "form_auth_tokens", feed: feed, form: form %>
+  </div>
+
+  <div class="col-12">
+    <hr>
+    <h3><%= t(".headers.ad_zones") %></h3>
+    <p><%= t(".help.ad_zones") %></p>
+
+    <%= form.check_box :billboard, value: 0 %>
+    <%= form.label :billboard, class: "me-2" %>
```
radical-ube, line 98: TODO: i understand that the attribute for the ad zones is set under `include_zones`, but it felt complicated to try and break that into these check boxes for this PR.

app/views/feeds/_form_main.html.erb
```diff
@@ -0,0 +1,203 @@
...
+    <%= form.label :url %>
+  </div>
+
+  <button type="button" class="prx-btn-help" data-bs-toggle="popover" data-bs-content="<%= t(".help.url") %>"><span class="material-icons">help</span></button>
+</div>
+
+<div class="col-12 mb-4 prx-field-group">
+  <div class="form-floating">
+    <%= form.text_field :new_feed_url %>
+    <%= form.label :new_feed_url %>
```
radical-ube, line 127: New Feed Url is another field left off Feeder designs so not sure if i needed to bring it over.

app/views/feeds/_form_main.html.erb
```diff
@@ -0,0 +1,203 @@
...
+    <%= form.label :episode_offset_seconds %>
+  </div>
+
+  <button type="button" class="prx-btn-help" data-bs-toggle="popover" data-bs-content="<%= t(".help.episode_offset_seconds") %>"><span class="material-icons">help</span></button>
+</div>
+
+<div class="col-12 mb-4">
+  <hr>
+  <h3><%= t(".headers.audio_format") %></h3>
+  <p><%= t(".help.audio_format") %></p>
```
radical-ube, line 172: TODO: like the Ad Zones checkboxes, it felt complicated to break the `audio_format` attribute into each of these selects in this PR. can address as fallout.

app/views/feeds/_form_status.html.erb
```diff
@@ -0,0 +1,15 @@
...
+    <h5 class="card-title">TITLE</h5>
+  </div>
+  <div class="card-footer d-flex align-items-center">
+    <% if feed.persisted? %>
+      <p class="status-text flex-grow-1"><strong><%= t(".updated_at_hint") %></strong> <br><%= l(feed.updated_at) %></p>
+      <%= link_to t(".discard"), podcast_feed_path(podcast, feed), class: "btn btn-discard-changed invisible me-2", data: {action: "unsaved#discard", unsaved_target: "discard"} %>
+      <%= form.submit t(".update"), class: "btn btn-primary" %>
+    <% else %>
+      <p class="status-text flex-grow-1"><strong><%= t(".new_hint") %></strong></p>
+      <%= form.submit t(".create"), class: "btn btn-primary" %>
```
radical-ube, line 12: TODO: `create` is not yet included in the UI.

config/routes.rb
```diff
@@ -6,12 +6,12 @@
...
       resources :imports
       resource :planner, only: [:show, :create], controller: :podcast_planner
-      resources :feeds, only: [:index]
+      resources :feeds, except: [:index, :edit]
       resources :episodes, only: [:index, :create, :new]
     end
 
     resources :episodes, except: [:create, :new]
-    resources :feeds
+    # resources :feeds
```
radical-ube, line 14: do we want to keep the original `/feeds` routing? i'm not aware of the use case for it.

app/helpers/feeds_helper.rb
```diff
@@ -0,0 +1,61 @@
+module FeedsHelper
+  def episode_offset_options
+    [
+      ["5 minutes", 300],
```
cavis, line 4: I'd probably leave the values in here, and move the english strings to en.yml?  (Or i'm fine with tackling the i18n stuff in a separate ticket).

app/models/feed.rb
cavis: Just the opposite of default.  Each podcast must have exactly 1 default feed (where the slug is nil).

app/policies/feed_token_policy.rb
```diff
@@ -0,0 +1,9 @@
+class FeedTokenPolicy < ApplicationPolicy
+  def update?
```
cavis, line 2: I think you'll eventually need to chain `create?` and `destroy?` to this update method.

And you'll also need to add a `show?` to the `FeedPolicy` - it seems like right now, I can change the `Podcast.update(prx_account_uri: "/api/v1/accounts/999999")`, and the page still loads for me.

app/views/feeds/_form_main.html.erb
cavis: +1 ... we should keep parity with Publish, for the moment.

app/views/feeds/_form_main.html.erb
cavis: May be worth extracting _something_... this file is getting a bit long.  Abstract out the more advanced sections?

app/views/feeds/_form_main.html.erb
cavis: +1

Could try it with helpers?  Or make a concern on the `Episode` model that provides a getter/setter for each of these checkboxes.

app/views/feeds/_form_main.html.erb
cavis: Yes, we do need it.

app/views/feeds/_form_main.html.erb
```diff
@@ -0,0 +1,203 @@
...
+    <%= form.number_field :display_full_episodes_count %>
+    <%= form.label :display_full_episodes_count %>
+  </div>
+
+  <button type="button" class="prx-btn-help" data-bs-toggle="popover" data-bs-content="<%= t(".help.display_full_episodes_count") %>" data-bs-html="true"><span class="material-icons">help</span></button>
+</div>
+
+<div class="col-12 mb-4 prx-field-group">
+  <div class="form-floating">
+    <%= form.select :episode_offset_seconds, episode_offset_options %>
```
cavis, line 162: Missing some allow-blanks in a few of these dropdowns?

config/routes.rb
cavis: Yeah, go ahead and remove.  I used to host an API at that endpoint, that Dovetail Router used.

But - it seems I already moved that to `/api/v1/feeds`.  And in staging right now, that `/feeds` path returns an html list.

config/locales/en.yml
```diff
@@ -78,6 +87,22 @@ en:
         subtitle: Teaser
         summary: iTunes Summary
         title: Title
+      feed:
+        title: Feed Label
```
cavis, line 91: I feel like this is referred to as "Title" in another piece of help text?  Ah well, can cleanup later.

test/controllers/feeds_controller_test.rb
```diff
@@ -23,25 +20,20 @@ class FeedsControllerTest < ActionDispatch::IntegrationTest
...
-  test "should get edit" do
-    get edit_feed_url(@feed)
+    get podcast_feed_url(podcast, feed)
     assert_response :success
   end
 
   test "should update feed" do
-    patch feed_url(@feed), params: {feed: {}}
-    assert_redirected_to feed_url(@feed)
+    patch podcast_feed_url(podcast, feed), params: {feed: {}}
```
cavis, line 28: I'd probably try setting a couple of actual params here.

And should also add tests for:

- Authorizes `#show`
- Authorizes `#update`
- Validates `#update`

app/views/feeds/_form_status.html.erb
```diff
@@ -0,0 +1,15 @@
+<div class="card">
+  <div class="card-header-primary">
+    <h5 class="card-title">TITLE</h5>
```
cavis, line 3: TITLE?

app/views/feeds/_form_main.html.erb
```diff
@@ -0,0 +1,203 @@
+<div class="col-12 mb-4 prx-field-group">
+  <div class="form-floating">
+    <%= form.text_field :feed_link, value: feed.published_url, disabled: true %>
+    <%= form.label :feed_link %>
+  </div>
+
+  <%= link_to(feed.published_url, class: "prx-btn-help", target: :_blank) do %>
```
cavis, line 7: For private feeds, this should include a valid FeedToken.  (See `FeedToken.feed_published_url_with_token`)

config/locales/en.yml
```diff
@@ -1,3 +1,12 @@
+defaults:
+  form_status: &form_status
```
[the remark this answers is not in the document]
cavis: Ooh, I didn't actually know that was a thing.  I've got some additional button text things in #629 I should probably move to here at some point.

cavis (changes requested):
Coming together!

A few minor bugs/suggestions, but mainly this is looking good.

app/helpers/feeds_helper.rb
```diff
@@ -0,0 +1,37 @@
+module FeedsHelper
+  def episode_offset_options
+    values = [300, 900, 3600, 21600, 43200, 86400]
+
+    I18n.t([:min_5, :min_15, :hr_1, :hr_6, :hr_12, :hr_24], scope: [:feeds, :helper, :episode_offset_options]).map.with_index { |label, i| [label, values[i]] }
```
cavis, line 5: Hmm... can you not just have numbers (300, 900, 3600...) in en.yml?  It seems to slightly odd to have 2 versions of them in here.

app/helpers/feeds_helper.rb
```diff
@@ -0,0 +1,37 @@
...
+  def audio_channel_options
+    values = [1, 2]
+
+    I18n.t([:mono, :stereo], scope: [:feeds, :helper, :audio_channel_options]).map.with_index { |label, i| [label, values[i]] }
+  end
+
+  def audio_sample_options
+    values = [8000, 11025, 12000, 16000, 22050, 24000, 44100, 48000]
+
+    I18n.t([:khz_8, :khz_11, :khz_12, :khz_16, :khz_22, :khz_24, :khz_44, :khz_48], scope: [:feeds, :helper, :audio_sample_options]).map.with_index { |label, i| [label, values[i]] }
```
cavis, line 35: Yeah, with all of these - it seems redundant to have the `values`, then this slightly-more-human-readable symbol version of the values, and then the human-readable `en.yml` version.  I guess I'd prefer just the values, and their mappings in en.yml.

app/policies/feed_policy.rb
```diff
@@ -1,4 +1,8 @@
 class FeedPolicy < ApplicationPolicy
+  def show?
+    update?
```
cavis, line 3: Is this true, though?  What about readonly users?  I think you want to check if the user can `show?` the Podcast.

app/policies/feed_token_policy.rb
```diff
@@ -0,0 +1,21 @@
+class FeedTokenPolicy < ApplicationPolicy
+  def show?
+    update?
```
cavis, line 3: Same as with the `Feed` ... I think readonly users should be able to see these.  (But the FormBuilder will disable the fields, so you can't edit them).

cavis (changes requested):
Looking good - just 2 things.

1. Are those intermediate i18n keys necessary?  Why not just map the numeric values directly to en.yml keys?
2. Need to handle readonly users (though I'm fine if you want to punt this one to a different PR)

app/helpers/feeds_helper.rb
radical-ube: i will look at it again. i think i had trouble using integers as keys in the yml file, but i didn't investigate too deeply into how i was utilizing them here so i went with this weird intermediate way that i knew how to make work. but i will try mapping more directly and see if i was missing something in the implementation.

cavis: Hrmm... maybe the yml file doesn't like integer keys?  Weird.  I would have thought something like this would "just work"-

```ruby
[96, 112, 128, 160, 192, 224, 256, 320].map do |val|
  [I18n.t(val), val]
end
```

radical-ube: ok i think i got it. the keys have to be strings in the yml file. i can change this.

app/policies/feed_policy.rb
cavis: FWIW, I just gave the ID staging user `ryancavistest` a readonly role.  So we can masquerade as that, (and click the "reload" button in the Feeder upper-right dropdown), and test some of this readonly stuff.

(I'm actually seeing some issues throughout the app with readonly ... so maybe we'll end up with a ticket to just go through and fix all the pages).

app/policies/feed_policy.rb
```diff
@@ -1,14 +1,18 @@
 class FeedPolicy < ApplicationPolicy
+  def show?
+    PodcastPolicy.new(token, resource.podcast).show?
+  end
+
   def create?
-    token&.authorized?(account_id)
+    PodcastPolicy.new(token, resource.podcast).create?
```
cavis, line 7: Hrm... following the logic in Publish - you don't need to be able to Create/Delete the podcast to CRUD the feeds within the podcast.  So I think create/destroy should still be chained to `update?`.

But the `show?` looks correct to me here.

cavis (changes requested):
Neeeearly there.  Just 1 policy change (and 1 suggestion to DRY up the feeds helper a bit).

test/controllers/api/auth/feeds_controller_test.rb
```diff
@@ -4,8 +4,8 @@
   let(:account_id) { 123 }
   let(:podcast) { create(:podcast, prx_account_uri: "/api/v1/accounts/#{account_id}") }
   let(:feed) { create(:feed, podcast: podcast, slug: "test-slug") }
-  let(:token) { StubToken.new(account_id, ["member"]) }
-  let(:bad_token) { StubToken.new(account_id + 100, ["member"]) }
+  let(:token) { StubToken.new(account_id, ["podcast_edit"]) }
```
radical-ube, line 7: in this and the `FeedPolicyTest`, the scope now has to match how the Podcast scope is determined. in this case, the `FeedPolicy` actions are determined by the `update?` method, which uses the `PodcastPolicy.update?` method, which checks for the `:podcast_edit` scope, not just a `:member` scope.

cavis: Ah, good catch!

cavis (approved):
👍 looks good to me - onwards!

PR #639, opened 2023-05-08
Episode medium

cavis:
I can't remember which of these 3 other PRs this goes after?  But probably get all those merged first.  There will be conflicts!

Trying to fixup the MediaResources schema, so I can add unsegmented audio.

In this chapter:

- [x] Adds an `episode.medium` field, which can currently be audio/video/nil.  Does some validating based on that.
- [x] Deprecate the `Enclosure` subclass of `MediaResource`.  Everything is a `Content` now.
- [x] Migrate existing `Enclosure` to be `Content`, and a bunch of cleanup
- [x] Refactor Episode `media_***` into a concern, for clarity.  And remove the "media can be an Enclosure or Contents" code, which was complicating things.


app/models/concerns/episode_media.rb
```diff
@@ -0,0 +1,118 @@
+require "active_support/concern"
+
+module EpisodeMedia
```
cavis, line 3: These methods pretty much all existed in `Episode`.  Moved/grouped/sorted/standardized them here:

- Try to make them start with `media_` or `complete_media`
- More consistency around checking content status/medium
- Better validation on segment_counts

app/models/content.rb
```diff
@@ -1,32 +1,4 @@
 class Content < MediaResource
-  def self.build_from_content(episode, content)
```
cavis, line 2: These methods weren't actually used, aside from in the tests.  So they're gone!

app/models/episode_entry_handler.rb
```diff
@@ -63,28 +62,12 @@ def update_link
...
     episode.image = overrides[:image_url]
   end
 
   def update_contents
     new_contents = Array(overrides[:contents]).sort_by { |c| c[:position] }
-
-    episode.contents = new_contents.map do |c|
-      Content.build_from_content(episode, c)
-    end
+    episode.media = new_contents.map { |c| c[:url] }
```
cavis, line 71: Invoking the `media=` setter to do the heavy lifting - checking if the `content.original_url` has changed, replacing/deleting/creating contents, etc.

app/models/feed.rb
```diff
@@ -89,7 +89,7 @@ def feed_episodes
     feed_max = display_episodes_count.to_i
 
     filtered_episodes.each do |ep|
-      include_in_feed << ep if ep.media_ready?
+      include_in_feed << ep if ep.include_in_feed?
```
cavis, line 92: Seemed incorrect to have been calling `media_ready?` here.  Was missing some of the logic `include_in_feed?` had.

app/views/podcasts/show.rss.builder
```diff
@@ -116,7 +116,8 @@ xml.rss "xmlns:atom" => "http://www.w3.org/2005/Atom",
         xml.pubDate ep.published_at.utc.rfc2822
         xml.link ep.url || ep.enclosure_url(@feed)
         xml.description { xml.cdata!(ep.description || "") }
-        xml.enclosure(url: ep.enclosure_url(@feed), type: ep.content_type(@feed), length: ep.file_size) if ep.media?
+        # TODO: may not reflect the content_type/file_size of replaced media
```
cavis, line 119: Trying to untangle things - realized these content_type/file_size/duration methods aren't going to be accurate, if you just replaced/changed some files.

But I guess that's an edge case?  So I punted for now.

test/factories/episode_factory.rb
```diff
@@ -11,24 +11,25 @@
...
 
     summary do
-      "<a href='/tina'>Tina</a> McElroy Ansa is a little girl when her father's business goes under"
+      "<a href=\"/tina\">Tina</a> McElroy Ansa is a little girl when her father's business goes under"
     end
 
-    after(:create) do |episode, _evaluator|
-      episode.enclosures << create(:enclosure, episode: episode, status: "complete")
-      episode.images << create(:episode_image, episode: episode, status: "complete")
+    factory :episode_with_media do
```
cavis, line 26: Trying to cleanup factories a bit - it seems like a better idea to have the basic `:episode` factory have 0 contents/images.  And the child `:episode_with_media` creates those.

test/factories/episode_image_factory.rb
```diff
@@ -11,10 +11,6 @@
     format { "png" }
     status { "complete" }
 
-    after(:create) do |episode_image, _evaluator|
-      episode_image.url = episode_image.published_url if episode_image.status == "complete"
```
cavis, line 15: This is handled by the overridden image_file.url method now.  (Same as media_resources).

test/models/concerns/episode_media_test.rb
```diff
@@ -0,0 +1,276 @@
+require "test_helper"
+
+class EpisodeMediaTest < ActiveSupport::TestCase
+  let(:c1) { build_stubbed(:content, status: "complete", position: 1) }
+  let(:c2) { build_stubbed(:content, status: "complete", position: 2) }
+  let(:ep) { build(:episode, segment_count: 2, contents: [c1, c2]) }
+
+  describe "#complete_media_ready?" do
```
cavis, line 8: Again, these were mostly ported from the `episode_test.rb`.  But I filled in the missing coverage, sorted, etc.

app/models/concerns/episode_media.rb
```diff
@@ -0,0 +1,118 @@
...
+  extend ActiveSupport::Concern
+
+  def complete_media
+    contents.complete_or_replaced.group_by(&:position).values.map(&:first)
```
a-maci29, line 7: What is `:position`?

cavis: The audio segments in the episode (the `Content`s) have a `position` field.  To determine what order we the segments of an episode should be sorted/played in.

app/models/media_resource.rb
```diff
@@ -1,4 +1,7 @@
 class MediaResource < ApplicationRecord
+  AUDIO_EXTENSIONS = %w[aac aiff au flac m4a m4b mp2 mp3 ogg wav]
+  VIDEO_EXTENSIONS = %w[avi flv m4v mov mp4 webm wmv]
+
```
a-maci29, line 4: Would it be worthwhile to include an `mp4` as an audio extension? Or is it problematic to always consider mp4s a video?

app/models/episode_import.rb
```diff
@@ -86,7 +86,7 @@ def set_audio_metadata!
   end
 
   def set_file_resources!
-    episode.contents = audio_content_params
+    episode.media = audio_content_params
```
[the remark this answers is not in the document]
cavis: `audio` may be a bit of a misnomer here?  RSS can contain enclosures for other media as well.

Reading down in `def entry_audio_files` ... it looks like it either pulls media urls from the `config[:audio]` (which IIRC is just there in case we want to manually override something).  Or more often, it's the `enclosure_url` - which could be a video url.

db/migrate/20230501194608_convert_enclosures.rb
```diff
@@ -0,0 +1,31 @@
+class ConvertEnclosures < ActiveRecord::Migration[7.0]
+  def up
+    add_column :episodes, :medium, :integer
```
radical-ube, line 3: what makes medium type integer? an enum?

cavis: Yep, it's an enum of `audio` / `video`.  Though I did think of just making a string.  But the other enums in Feeder seem to be integer fields?

I'm still a bit conflicted on this one.  There's also a `media_resources.medium` string field.  But that field is more free-form, since you could theoretically upload any file type.  (We only validate the media_resource medium is audio/video when status == complete).

radical-ube (approved):
+1 reads well and works as expected

Issue #605, opened 2023-03-28
Apple Limit API Endpoint Polling


svevang:
To keep the number of API calls down:

- [ ] Mark the episode with a `apple_synced_at`
- [ ] Guard in the apple publisher for episodes having nil apple_synced_at
- [ ] Consider the presence of `podcast_deliveries` and `podcast_delivery_files` on the container when determining when to sync.

kookster:
also consider batching episodes on import

PR #647, opened 2023-05-11
Limit api endpoint polling

svevang:
closes https://github.com/PRX/feeder.prx.org/issues/605
- ~Fixes a bug where the arranged mp3 expired before it could be uploaded to Apple~
- Resets the source CDN url and metadata with each upload attempt
- Uses SyncLog to store remote state for episodes. All apple models store their remote state counterpart in the SyncLog and local models `have_one` sync log, with validations and uniqueness constraints.
- New `Apple::Episode.poll_episode_state` that mirrors the containers/deliveries/delivery files. Episodes have local state (via sync logs).
- Switches the `Apple::Publisher` to looping in batches of 25 episodes (column to row oriented looping)
- Prevents all API interactions in the case where an episode has "good uploaded audio" marked in local state


app/models/apple/episode.rb
```diff
@@ -336,14 +372,28 @@ def audio_asset_vendor_id
       apple_attributes["appleHostedAudioAssetVendorId"]
     end
 
-    def audio_hosted_audio_asset_container_id
+    def apple_hosted_audio_asset_container_id
```
svevang, line 375: Fixes a typo

app/models/apple/episode.rb
```diff
@@ -114,26 +143,38 @@ def self.publish(api, episodes)
...
 
       results.map do |res|
-        apple_id = res.dig("api_response", "val", "data", "id")
-        guid = res.dig("api_response", "val", "data", "attributes", "guid")
-        ep = episodes_by_guid.fetch(guid)
+        upsert_sync_log(episodes_by_guid[res.dig("request_metadata", "guid")], res)
+      end
+    end
+
+    def self.upsert_sync_log(ep, res)
```
svevang, line 154: This mirrors `Apple::PodcastContainer.upsert_podcast_container`, `Apple::PodcastDelivery.upsert_podcast_delivery`, etc.

But since creating the episode is handled a bit differently, this falls back on a single responsibility of upserting the apple sync log (remote data wrapper).

app/models/apple/podcast_container.rb
```diff
@@ -222,8 +257,27 @@ def missing_podcast_audio?
...
+    def source_url_expires_at
+      return nil if source_url.blank?
+
+      uri = URI.parse(source_url)
+      query_params = CGI.parse(uri.query)
+      timestamp = query_params["exp"].first
+      Time.at(timestamp.to_i).utc
+    end
+
+    def source_url_expired?
```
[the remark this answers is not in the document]
cavis: Wait, how does that happen?  Are you holding onto a `dovetail-cdn` url for multiple days?

Generally, we should always be going back to `dovetail.prxu.org` (DTR) to get a fresh redirect, before doing anything with that file.  `dovetail-cdn` urls aren't meant to be long-lived.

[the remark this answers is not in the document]
svevang: I refactored the source reset routine so that it runs with each re-upload attempt.  It's a bit aggressive in this way, because the delivery files are the sentinels that determine if another upload should be attempted.  But with this approach there is not really an alternative. Resetting the source CDN url for the upload data means that the delivery file is invalidated. So it is destroyed.

app/models/apple/api.rb
```diff
@@ -183,7 +183,7 @@ def unwrap_bridge_response(resp)
 
     def bridge_remote(bridge_resource, bridge_options, batch_size: DEFAULT_BATCH_SIZE)
       url = bridge_url
-      Rails.logger.info("Apple::Api BRIDGE #{bridge_resource} #{url.hostname}:#{url.port}/bridge")
+      Rails.logger.info("Apple::Api BRIDGE #{bridge_resource} #{url.hostname}:#{url.port}/bridge", {param_count: bridge_options.count})
```
cavis, line 186: Digging through the calls to this method a bit ... so `bridge_options` is really the _operations_ we're calling the bridge to perform?  So `bridge_options.count` is a measure of how much work we're calling the bridge with?

db/migrate/20230503192155_sync_log_has_unique_state.rb
```diff
@@ -0,0 +1,14 @@
+class SyncLogHasUniqueState < ActiveRecord::Migration[7.0]
+  def change
+    reversible do |dir|
+      dir.up do
+        SyncLog.where.not(id: SyncLog.latest).delete_all
+      end
+    end
+
+    add_index :sync_logs, [:feeder_type, :feeder_id], unique: true
```
cavis, line 9: Ah, I see.  So there's guaranteed to only be one of these.

Noting that these SyncLogs aren't namespaced to "apple", but are generally treated as if they're unique to apple.  If we want to expand them to handle other external systems, will need to add some other column/scope.  (But that's fine - one thing at a time!)

db/migrate/20230503192155_sync_log_has_unique_state.rb
```diff
@@ -0,0 +1,14 @@
...
+
+    add_column :sync_logs, :api_response, :text
+    rename_column :sync_logs, :sync_completed_at, :updated_at
```
cavis, line 12: There was no `updated_at` until now?  Always good to have the created/updated timestamps!

app/models/apple/publisher.rb
```diff
@@ -2,6 +2,8 @@
 
 module Apple
   class Publisher
+    PUBLISH_CHUNK_LEN = 25
```
cavis, line 5: So this is the number of episodes we're processing at a time, for 1 show?

app/models/apple/publisher.rb
```diff
@@ -32,166 +34,213 @@ def episodes_to_sync
       show.episodes
     end
 
-    def poll!
+    def poll!(eps = episodes_to_sync)
```
cavis, line 37: So what calls this?  (Or... what will eventually call this?)  A cron, or do we trigger from the publish-feed-job or something?

app/models/apple/podcast_container.rb
```diff
@@ -95,23 +123,28 @@ def self.upsert_podcast_container(episode, row)
...
+      pc.reload if action == :updated
+      episode.feeder_episode.reload
 
       pc
     end
 
     def self.get_podcast_containers_via_episodes(api, episodes)
+      # Only query for episodes that don't have a podcast container
+      # The container. Assume that if we have a container record, we don't need to poll.
+      _eps_with_container, eps_without_container = episodes.partition(&:has_container?)
```
cavis, line 138: or `episodes.reject`?

app/models/apple/podcast_container.rb
```diff
@@ -6,17 +6,42 @@ class PodcastContainer < ActiveRecord::Base
...
+        container.podcast_deliveries.destroy_all
+      end.compact
+    end
 
     def self.update_podcast_container_file_metadata(api, episodes)
       containers = episodes.map(&:podcast_container)
       raise "Missing podcast container for episode" if containers.any?(&:nil?)
 
+      # do not update if the source url is expired, see .reset_for_expired_source_urls
+      containers = containers.reject(&:source_url_expired?)
```
cavis, line 43: +1

cavis (approved):
Couple of questions, but all looks good to me!

app/models/apple/api.rb
svevang: That's right.  The api requests are batched as a set of options.

app/models/apple/publisher.rb
cavis: Hrm, maybe I'm looking in the wrong place(s).  But I don't see anything calling `poll!` right now.

svevang: Ahh, ok I thought you were talking about the parameters (which are symmetric with `#publish!`).

It's a workflow method. Called on the rails console to probe remote apple state. `publish!` is the superset method to this one and that is the method called from the job. But this one can be used in a manner that does not affect apple state.

PR #649, opened 2023-05-15
Apple api bulk edit

svevang:
closes #921

Set up some facilities for bulk editing with the intent of:

- Adjusting the publishing state for select episodes from archived to unarchived/drafting
- setting the `apple_hosted_audio_available_start_date` far into the future as a guard for the publishing workflow.

Additionally, set up some memoizing/caching so that we can efficiently poll the back catalog (1000+ episodes).


app/models/apple/episode.rb
```diff
@@ -258,6 +255,30 @@ def episode_create_parameters
       }
     end
 
+    def update_episode_bridge_params
```
svevang, line 258: I think some of this parameter handling code is a bit too nested, (long method call chains with similar names).

I have a few ideas on how to fix this up in the future.  I think we can flatten some of these calls, maybe use some blocks to DRY things up.

cavis (approved):
👍 looks good to me!

PR #653, opened 2023-05-23
Apple sync job

svevang:
Sets up a new Apple sync/publish job as part of a testing strategy. This will kick off apple publishing with rss publishing. Once this has settled down, then we can merge this back into the PublishFeedJob to get the blocking semantics (Apple publishing is blocking RSS publishing).

### DIFF
--- a/app/models/episode.rb
+++ b/app/models/episode.rb
@@ -209,7 +209,7 @@ class Episode < ApplicationRecord
   end
 
   def apple_mark_for_reupload!
-    podcast_deliveries.destroy_all
+    apple_podcast_deliveries.destroy_all
   end
 
   def publish!
--- a/app/models/episode.rb
+++ b/app/models/episode.rb
@@ -208,7 +208,12 @@ class Episode < ApplicationRecord
     images.each { |i| i.copy_media(force) }
   end
 
+  def apple_mark_for_reupload!
+    podcast_deliveries.destroy_all
+  end
+
   def publish!
+    apple_mark_for_reupload!
     podcast&.publish!
   end
 
--- a/config/shoryuken.yml
+++ b/config/shoryuken.yml
@@ -22,4 +22,4 @@ groups:
   single_threaded:
     concurrency: 1
     queues:
-      - ["<%= prefix %>_feeder_publish_artifacts", 1]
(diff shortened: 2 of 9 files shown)


app/models/episode.rb
```diff
@@ -208,6 +208,15 @@ def copy_media(force = false)
     images.each { |i| i.copy_media(force) }
   end
 
+  def apple_mark_for_reupload!
+    apple_podcast_deliveries.destroy_all
+  end
+
+  def publish!
+    apple_mark_for_reupload!
```
cavis, line 216: Just noting that this method will get called _even if_ `episode.published_at.nil?` (the ep isn't published yet).

In that case ... nothing bad would happen, right?  Just a no-op, since it would have no deliveries.

config/shoryuken.yml
```diff
@@ -18,3 +18,8 @@ queues:
   - ["<%= prefix %>_announce_feeder_story_delete", 1]
   - ["<%= prefix %>_announce_feeder_story_publish", 1]
   - ["<%= prefix %>_announce_feeder_story_unpublish", 1]
+groups:
+  single_threaded:
```
cavis, line 22: I've never heard of / tried this!  But I'm guessing it worked locally for you.

I do wonder how this will scale.  How long do these typically take?  And how many shows could we theoretically handle before the lag gets out of control?  (Good problems to solve later - but I'm just curious).

cavis (approved):
Couple of questions, but looks good to me!  (Once the queue exists, that is!)

app/models/episode.rb
svevang: Yep, it would be a no-op. And since we only pull episodes from Feed#feed_episodes we should only get the episodes that are published in the RSS.

config/shoryuken.yml
svevang: I was able to spin up a job queue and run a job.

In the typical case where we are publishing 'one more' episode, the time is roughly a minute or two.  There is a strong assumption in here that we wait for the episode to finish processing completely.  In fact you cannot mark an episode from DRAFTING to PUBLISHED in the api with the media settling down completely.

PR #657, opened 2023-05-26
Feat/crud feed tokens

### LINKED ISSUE #644: Feed Tokens CRUD

radical-ube:
resolves #644

### DIFF
--- a/app/javascript/controllers/toggle_field_controller.js
+++ b/app/javascript/controllers/toggle_field_controller.js
@@ -3,6 +3,12 @@ import { Controller } from "@hotwired/stimulus"
 export default class extends Controller {
   static targets = ["check", "field"]
 
+  connect() {
+    if(this.checkTarget.checked) {
+      this.toggleDisplay()
+    }
+  }
+
   changeCheck() {
     this.checkTarget.parentElement.classList.add("d-none")
     this.fieldTarget.parentElement.classList.remove("d-none")
@@ -28,5 +34,9 @@ export default class extends Controller {
         el.classList.remove("d-none")
       }
     })
+  }
+
+  toggleDisplay() {
+    this.fieldTarget.classList.toggle("d-none")
   }
 }
--- a/app/javascript/controllers/toggle_field_controller.js
+++ b/app/javascript/controllers/toggle_field_controller.js
@@ -4,7 +4,7 @@ export default class extends Controller {
   static targets = ["check", "field"]
 
   connect() {
-    if(this.checkTarget.checked) {
+    if (this.checkTarget.checked) {
       this.toggleDisplay()
     }
   }
--- a/app/models/feed_token.rb
+++ b/app/models/feed_token.rb
@@ -3,6 +3,7 @@ class FeedToken < ApplicationRecord
 
   validates :token, presence: true, uniqueness: {scope: :feed_id}
   validates_format_of :token, with: /\A[0-9a-zA-Z_.-]+\z/
+  validates :label, presence: true
 
   after_initialize :set_defaults
 
--- a/app/views/feeds/_form_auth_token.html.erb
+++ b/app/views/feeds/_form_auth_token.html.erb
@@ -1,15 +1,21 @@
-<div class="row">
-  <div class="col mb-4">
-    <div class="form-floating">
-      <%= form.text_field :label %>
-      <%= form.label :label %>
+<div class="mb-4 prx-field-group">
+  <div class="row">
+    <div class="col-6">
+      <div class="form-floating">
+        <%= form.text_field :label %>
+        <%= form.label :label %>
+      </div>
+    </div>
+
+    <div class="col-6">
+      <div class="form-floating">
+        <%= form.text_field :token %>
+        <%= form.label :token %>
+      </div>
     </div>
   </div>
 
-  <div class="col mb-4">
-    <div class="form-floating">
-      <%= form.text_field :token %>
-      <%= form.label :token %>
-    </div>
-  </div>
+  <%= link_to podcast_feed_path({podcast_id: @podcast, id: feed}, feed_tokens_destroy_params(form)), class: "btn" do %>
+    <span class="material-icons">delete</span>
+  <% end %>
 </div>
--- a/app/views/feeds/_form_auth_tokens.html.erb
+++ b/app/views/feeds/_form_auth_tokens.html.erb
@@ -1,13 +1,19 @@
-<% if feed.tokens.present? %>
-  <%= form.fields_for :feed_tokens do |fields| %>
-    <%= turbo_frame_tag "feed-token-#{fields.object.id}" do %>
-      <% if fields.object.marked_for_destruction? %>
-        <%= fields.hidden_field :_destroy, value: "1" %>
-      <% else %>
-        <%= render "form_auth_token", form: fields, feed: feed %>
-      <% end %>
+<% feed.tokens.build %>
+
+<%= form.fields_for :feed_tokens do |fields| %>
+  <%= turbo_frame_tag "feed-token-#{fields.object.id}" do %>
+    <% if fields.object.marked_for_destruction? %>
+      <%= fields.hidden_field :_destroy, value: "1" %>
+    <% elsif fields.object.id.blank? %>
+      <div data-controller="toggle-field">
+        <div class="d-none" data-toggle-field-target="field">
+          <%= render "form_auth_token", form: fields, feed: feed %>
+        </div>
+        <%= fields.check_box :add_token, checked: false, data: {toggle_field_target: "check", action: "toggle-field#toggleDisplay"} %>
+        <%= fields.label :add_token %>
+      </div>
+    <% else %>
+      <%= render "form_auth_token", form: fields, feed: feed %>
     <% end %>
   <% end %>
 <% end %>
-
-<span>Add token</span>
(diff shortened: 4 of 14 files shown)


app/javascript/controllers/toggle_field_controller.js
```diff
@@ -3,6 +3,12 @@ import { Controller } from "@hotwired/stimulus"
 export default class extends Controller {
   static targets = ["check", "field"]
 
+  connect() {
```
radical-ube, line 6: this was added for the `private` checkbox that toggles the display for Auth Tokens. the default is `d-none` and this checks to see if the checkbox is already checked, and toggles it on if it is.

not sure how i feel about putting this on the `connect()` method though, since i know other uses of this controller don't operate this way (although i don't think it disturbs any of them at the moment). but i wasn't sure if there was a better way to check if a checkbox is checked on render.

app/models/feed_token.rb
```diff
@@ -3,6 +3,7 @@ class FeedToken < ApplicationRecord
 
   validates :token, presence: true, uniqueness: {scope: :feed_id}
   validates_format_of :token, with: /\A[0-9a-zA-Z_.-]+\z/
+  validates :label, presence: true
```
radical-ube, line 6: added this validation since it seems to be a form validation in Publish. (can't save without a label)

app/views/feeds/_form_auth_tokens.html.erb
```diff
@@ -1,7 +1,19 @@
-<% if feed.tokens.present? %>
-  <%= form.fields_for :feed_tokens, feed.tokens do |fields| %>
-    <%= render "form_auth_token", form: fields %>
+<% feed.tokens.build %>
+
+<%= form.fields_for :feed_tokens do |fields| %>
+  <%= turbo_frame_tag "feed-token-#{fields.object.id}" do %>
+    <% if fields.object.marked_for_destruction? %>
+      <%= fields.hidden_field :_destroy, value: "1" %>
+    <% elsif fields.object.id.blank? %>
```
radical-ube, line 7: for the token that is built, i decided to hide it until the Add Token checkbox is checked.

test/controllers/api/auth/feeds_controller_test.rb
```diff
@@ -47,15 +47,15 @@ class << @controller; attr_accessor :prx_auth_token; end
     describe "feed tokens" do
       before do
         feed.tokens.create!(label: "something", token: "tok1")
-        feed.tokens.create!(token: "tok2")
+        feed.tokens.create!(label: "something2", token: "tok2")
```
radical-ube, line 50: this and the rest of the test changes are to pass the new `:label, presence: true` validation by adding labels to all tokens where needed.

i wasn't sure if there was a need for some tokens to be unlabeled though?

app/javascript/controllers/toggle_field_controller.js
cavis: So I know it's not the greatest UX - but I'm going to advocate with consistency with Augury on this one...

Instead of "click the checkbox to add", can you just always have the extra/new token rendered on the page?  (Similar to Series->Placements and Creatives->Pingbacks in Augury).

I do want to come clean this up at some point... but parity with Augury seems like a good point to get feedback.  And really, the "delete button actually deletes" is much better than Augury already.

app/views/feeds/_form_auth_token.html.erb
```diff
@@ -1,15 +1,21 @@
...
     </div>
-  </div>
 
-  <div class="col mb-4">
-    <div class="form-floating">
-      <%= form.text_field :token %>
-      <%= form.label :token %>
+    <div class="col-6">
+      <div class="form-floating">
+        <%= form.text_field :token %>
```
cavis, line 12: Related to an earlier suggestion: if `form.object.new_record?` ... don't render the token field.  Since we auto-generate it anyways, you don't need to provide it.

app/views/feeds/_form_auth_tokens.html.erb
cavis: This does feel a bit awkward to add to the "toggle field controller".  Though I can see how similar it is.

I'd suggest for fast-feedback, you just always show the new token "label" field, and hide the new token "token" field.  Once the user adds a label and saves, they create a token.

Longer term, this should probably be an "add token" button.  That (ideally) allows you to add as many tokens as you want.  But - let's get some quick feedback first.

app/views/feeds/_form_auth_tokens.html.erb
```diff
@@ -1,7 +1,19 @@
-<% if feed.tokens.present? %>
-  <%= form.fields_for :feed_tokens, feed.tokens do |fields| %>
-    <%= render "form_auth_token", form: fields %>
+<% feed.tokens.build %>
+
+<%= form.fields_for :feed_tokens do |fields| %>
+  <%= turbo_frame_tag "feed-token-#{fields.object.id}" do %>
+    <% if fields.object.marked_for_destruction? %>
+      <%= fields.hidden_field :_destroy, value: "1" %>
```
cavis, line 6: So I did get the "unsaved controller" registering a deleted thing as a change, which makes the discard button show up etc.  By adding a `data: {unsaved_target: "changed"}` to the element, which triggers the unsaved `changedTargetConnected` method.

But apparently I broke that doing magic stuff on the audio uploader.  So - I'll open a ticket to fix that!  Then can circle back to this page's UX.

app/views/feeds/_form_custom_fields.html.erb
```diff
@@ -48,24 +48,24 @@
...
 </div>
 
-<div class="col-12">
+<div class="col-12" data-controller="toggle-field">
   <hr>
   <h3><%= t(".headers.private") %></h3>
   <div class="prx-field-group">
     <div>
-      <%= form.check_box :private %>
+      <%= form.check_box :private, data: {toggle_field_target: "check", action: "toggle-field#toggleDisplay"} %>
```
cavis, line 56: +1 this feels like the "standard" use of toggle-field.

test/controllers/api/auth/feeds_controller_test.rb
cavis: Nope - they should always have a label.  Otherwise there's no clue what apps/users might be using it.

app/views/feeds/_form_auth_tokens.html.erb
cavis: TMI: the Unsaved controller determines if the form has "changed" by scanning for [the is-changed class](https://github.com/PRX/feeder.prx.org/blob/master/app/javascript/controllers/unsaved_controller.js#L105).

But - even adding `class: "is-saved"` to this hidden destroy field isn't enough.  Because for some reason, I elected to make the controller [ignore visually hidden elements](https://github.com/PRX/feeder.prx.org/blob/master/app/javascript/controllers/unsaved_controller.js#L113).  So I need to go back and figure out why the heck I did that.

Or maybe you could render a 0x0 non-hidden div right above this destroy field?  May be worth a shot.

```html
<div class="is-changed" data-unsaved-target="changed">
```

app/javascript/controllers/toggle_field_controller.js
radical-ube: maybe worded unclearly, but this check is to display/hide the entire Auth Tokens section based on `private` being checked or not, as in Publish. but I could also just display the section by default and not hide?

cavis: Oh, maybe I misunderstood - I was thinking this was part of the "hide new token fields".

Can you just add the check to the server-side?  `class: "d-none" unless form.object.private?` or something?

app/views/feeds/_form_auth_token.html.erb
```diff
@@ -1,15 +1,25 @@
...
+        <div class="form-floating">
+          <%= form.text_field :token %>
+          <%= form.label :token %>
+        </div>
+      </div>
+    <% end %>
   </div>
+
+  <% unless form.object.new_record? %>
+    <%= link_to podcast_feed_path({podcast_id: @podcast, id: feed}, feed_tokens_destroy_params(form)), class: "btn" do %>
```
cavis, line 21: So this works, but it does generate a very odd looking href:

```
/podcasts/id=680&podcast_id=815/feeds/680?feed%5Bfeed_tokens_attributes%5D%5B0%5D%5B_destroy%5D=1&feed%5Bfeed_tokens_attributes%5D%5B0%5D%5Bid%5D=70
```

I'm actually a bit surprised that works!

But what you actually want is to just pass the 2 models in-order to the path helper, with the query params hash as the last argument:

```ruby
podcast_feed_path(feed.podcast, feed, feed_tokens_destroy_params(form))
```

(Also changed out the `@podcast` there, as it's a _bit_ of a rails convention to only reference those `@` vars in the top-level view).

app/views/feeds/_form_auth_tokens.html.erb
cavis: Hey, looks like this worked!  Neat!

app/javascript/controllers/toggle_field_controller.js
```diff
@@ -29,4 +29,8 @@ export default class extends Controller {
       }
     })
   }
+
+  toggleDisplay() {
```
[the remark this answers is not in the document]
radical-ube: i guess this might not be necessary if i wrapped the Auth Tokens section in a turbo frame? i haven't checked yet.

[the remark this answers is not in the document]
radical-ube: hmm, turbo doesn't quite work because the render is still based on `feed.privacy?`, which doesn't change until submit is clicked. so this method is used to toggle the display without relying on save/re-render.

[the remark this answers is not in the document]
cavis: Ah, missed that.  Yeah, works for me!

cavis (approved):
👍 wooo looks good

PR #658, opened 2023-05-26
Consider drafting episodes as not-synced

svevang:
Fixes a condition where the episode audio can be fully synced with apple and yet the episode in an unpublished state.  This PR makes insures that we include such an episode for the publishing process as well.  It will re-sync with the api, polling everything as usual (except the podcast container), and marked as published in Apple remote state.

### DIFF
--- a/test/models/apple/episode_test.rb
+++ b/test/models/apple/episode_test.rb
@@ -18,31 +18,14 @@ describe Apple::Episode do
       private_feed: private_feed)
   end
   let(:apple_episode) { build(:apple_episode, show: apple_show, feeder_episode: episode) }
+  let(:apple_episode_api_response) { build(:apple_episode_api_response, apple_episode_id: "123") }
+  let(:external_id) { apple_episode_api_response["api_response"]["api_response"]["val"]["data"]["id"] }
 
   before do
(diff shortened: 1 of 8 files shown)


test/models/apple/episode_test.rb
```diff
@@ -156,4 +135,17 @@
       end
     end
   end
+
+  describe "#synced_with_apple?" do
+    let(:apple_episode_api_response) { build(:apple_episode_api_response, publishing_state: "PUBLISH") }
+
+    it "should be false when drafting" do
```
cavis, line 142: So is this the only real change in the tests?  The rest of this looks like factory-shuffling.  (And I only see the one changed line outside the tests).

cavis (approved):
Looks like this is mostly test refactoring - 👍 on the 1 line change!

test/models/apple/episode_test.rb
svevang: The `build(:uploaded_apple_episode)` factory here precipitated a lot of the changes, but yep, mostly trying to get the api responses well represented in the factories to build up that happy path/uploaded episode. And that meant adding a few factories

PR #660, opened 2023-05-30
Fix thread pool logging

svevang:
Fixes the thread pool logging.

Currently, the publishing code tags the logs for easy searching. But since we're spawning threads, those are missing the thread-local tags. This PR makes sure that the thread local logging config has the right tags.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -4,6 +4,7 @@ networks:
     driver: bridge
 services:
   feeder:
+    image: feederprxorg_feeder
     build: .
     volumes:
       - .:/app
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -4,6 +4,7 @@ networks:
     driver: bridge
 services:
   feeder:
+    image: feederprxorg_feeder
```
svevang, line 7: Makes sure that the image names match between worker and here. The default image name no longer follows this pattern/format so pinning it here makes sure they all match.

cavis (approved):
👍

PR #663, opened 2023-06-02
Feat/feed images card

### LINKED ISSUE #643: Feed Images Card

radical-ube:
resolves #643 

mainly just following the existing pattern from Episode image upload, reusing code where possible.

### DIFF
--- a/app/views/feeds/_form_image.html.erb
+++ b/app/views/feeds/_form_image.html.erb
@@ -1,0 +1,8 @@
+<% feed.feed_images.build if feed.feed_images.empty? %>
+
+<%= form.fields_for :feed_images, include_id: false do |fields| %>
+  <% dpath = feed_destroy_image_path(feed, fields) %>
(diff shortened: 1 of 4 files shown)


app/views/feeds/_form_image.html.erb
```diff
@@ -0,0 +1,29 @@
+<%= turbo_frame_tag "itunes-images", data: {controller: "morphdom"} do %>
```
radical-ube, line 1: hmm, i copied these turbo frames over from `episode/_form_image` but i'm not sure what they are for? i tried with testing with them commented out and i couldn't discern the difference.

cavis: I think if you remove these, there's no turbo-frame wrapping the individual images at any level.

So when you click the delete button (or upload a new image) ... the entire page navigates, and you end up with a bunch of query params in your browser url.

With the turbo frame in there, just that frame navigates/refreshes.

cavis (approved):
👍 looks good to me!

PR #666, opened 2023-06-06
Feat/feed crud pt2

radical-ube:
adds create and delete actions for Feeds.


app/assets/stylesheets/shared/button.scss
```diff
@@ -16,6 +16,19 @@
   }
 }
 
+.btn-secondary-link {
```
radical-ube, line 19: the shared folder seems to be getting out of sync with Augury?

cavis: Yeah, that's probably true.  Though I'm a bit surprised to see something in Augury and not in Feeder?  (I thought generally, Feeder CSS was way ahead of Augury now - and the plan was to backport Feeder's code to Augury...)

app/views/feeds/new.html.erb
```diff
@@ -1,9 +1,5 @@
-<h1>New feed</h1>
+<% content_for :title, "New feed" %>
```
cavis, line 1: Maybe include the podcast title in this?

app/controllers/feeds_controller.rb
```diff
@@ -46,10 +53,15 @@ def update
...
   def destroy
-    @feed.destroy
-
     respond_to do |format|
-      format.html { redirect_to feeds_url, notice: "Feed was successfully destroyed." }
+      if @feed.destroy
+        format.html { redirect_to podcast_feed_path(@podcast, @podcast.default_feed), notice: (t ".success", model: "Feed") }
+      else
+        format.html do
+          flash.alert = t ".failure", model: "Feed"
```
cavis, line 61: I think these alerts in here need to be the `.now` flash style - otherwise you get 2 alerts.  Once after you click save, and a second one if you click away.

Generally, if you're doing a `render`, you need to use `flash.now`.  If you're redirecting, you use the regular `flash` (or append the `notice:` to the `redirect_to` itself... which amounts to the same thing).

app/views/feeds/_form.html.erb
```diff
@@ -1,4 +1,8 @@
-<%= form_with(url: podcast_feed_path(podcast, feed), model: feed, html: {autocomplete: "off"}, data: {controller: "unsaved disable", unsaved_changed_value: feed.changed?}, unsaved_confirm_value: t(".confirm")) do |form| %>
+<% model = feed.new_record? ? [feed.podcast, feed] : feed %>
+<% url = feed.new_record? ? podcast_feeds_path(feed.podcast) : podcast_feed_path(feed.podcast, feed) %>
+<% method = feed.new_record? ? :post : :put %>
+
+<%= form_with(url: url, model: model, method: method, html: {autocomplete: "off"}, data: {controller: "unsaved", unsaved_changed_value: feed.changed?}, unsaved_confirm_value: t(".confirm")) do |form| %>
```
cavis, line 5: You're missing the `disable` controller on this form?  The image uploads are complaining about it.  I might also pull all of this data up into a `<% data = {...} %>` above, to shorten the wrapping.

(I think I added the `disable` controller in a recent PR, to disable the save button while uploads are still in-progress).

test/controllers/feeds_controller_test.rb
```diff
@@ -5,21 +5,35 @@ class FeedsControllerTest < ActionDispatch::IntegrationTest
...
   let(:update_params) { {url: "/a_public_url", display_episodes_count: 5} }
+  let(:create_params) { {podcast: podcast, slug: "new_feed", title: "new title", private: false} }
 
   setup_current_user { build(:user, account_id: 123) }
 
-  # test "should get new" do
-  #   get new_feed_url
-  #   assert_response :success
-  # end
+  test "should get new" do
```
cavis, line 12: I'd probably also add a validation test to this - "authorizes new feed" version of this.

app/controllers/feeds_controller.rb
```diff
@@ -46,10 +53,15 @@ def update
 
   # DELETE /feeds/1
   def destroy
-    @feed.destroy
-
     respond_to do |format|
-      format.html { redirect_to feeds_url, notice: "Feed was successfully destroyed." }
+      if @feed.destroy
```
cavis, line 57: I think it appeared in a different PR ... but noting this "destroy" UI does deviate from the episode.

(The episode status card has a little carat-dropdown thing on the save button, which triggers the delete.  And then you get a bootstrap-modal to confirm rather than the browser built-in).

Should probably open a ticket to convert this one to match?  Or tackle here if you'd rather.

app/views/feeds/_form.html.erb
cavis: Hmm... there's also something odd going on with the "unsaved" controller here?  If I make a change, the field gets marked with the orange outline, but it lets me navigate away without prompting me "you have unsaved changes".

Maybe adding the "disable" controller will help?

app/controllers/feeds_controller.rb
radical-ube: added! see changes in `form_status` partial.

app/views/feeds/_form_status.html.erb
```diff
@@ -5,11 +5,50 @@
...
-      <%= form.submit t(".update"), class: "btn btn-primary" %>
+
+      <% if policy(feed).update? %>
+        <%= link_to t(".discard"), podcast_feed_path(podcast, feed), class: "btn btn-discard-changed invisible me-2", data: {action: "unsaved#discard", unsaved_target: "discard"} %>
+
+        <div class="btn-group">
+          <% data = {disable_with: t(".updating"), upload_with: t(".uploading")} %>
+          <%= form.submit t(".update"), class: "btn btn-primary", data: data %>
+
+          <% if policy(feed).destroy? && !feed.default? && !feed.new_record? %>
```
radical-ube, line 16: destroy button only allowed on persisted feeds that aren't the default. is this true? would the user ever need to be able to delete the default feed?

cavis: You got it!  Deleting the default feed would probably cause "bad things" to happen.

app/views/feeds/_form.html.erb
```diff
@@ -11,10 +16,13 @@
           <%= render "form_status", podcast: podcast, feed: feed, form: form %>
         </div>
       </div>
-      <% unless feed.default? %>
-        <%= render "form_image", podcast: podcast, feed: feed, form: form %>
+      <% unless feed.default? && !feed.new_record? %>
```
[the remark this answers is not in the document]
cavis: (update: see my controller comment)

cavis (approved):
👍 looks/works good!

PR #667, opened 2023-06-09
Fix replaced media coming from CMS

cavis:
Okay, so... I had written a [mark_for_replacement](https://github.com/PRX/feeder.prx.org/blob/main/app/models/media_resource.rb#L170) method to mimic how `mark_for_destruction` works.  But it would set both `deleted_at` and `replaced_at`.

This gets called from the [media=](https://github.com/PRX/feeder.prx.org/blob/main/app/models/concerns/episode_media.rb#L41) setter, which is called for API and Announced CMS updates.

But - apparently when you `.mark_for_destruction` something and then save its parent... it _ignores_ any other pending changes to that record.  So it was just setting `deleted_at` and ignoring the changed `replaced_at` field.

~Took a look at hacking/extending the `paranoia` gem to make it also set this field.  But that was complicated.  This takes an additional SQL statement, but it's fairly straightforward.   So... here we go!~

Just learned about `paranoia_destroy_attributes` ... and updated to that!  Should have done a better job reading the docs.

### DIFF
--- a/test/models/media_resource_test.rb
+++ b/test/models/media_resource_test.rb
@@ -128,16 +128,40 @@ describe MediaResource do
 
   it "marks completed resources for replacement" do
     mr = build_stubbed(:media_resource, status: "started")
+    refute mr.marked_for_destruction?
     refute mr.marked_for_replacement?
 
     mr.mark_for_replacement
+    assert mr.marked_for_destruction?
     refute mr.marked_for_replacement?
 
     mr.status = "complete"
     mr.mark_for_replacement
+    assert mr.marked_for_destruction?
     assert mr.marked_for_replacement?
+  end
 
-    mr.replaced_at = nil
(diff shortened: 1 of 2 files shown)


test/models/media_resource_test.rb
```diff
@@ -128,16 +128,40 @@
...
 
     mr.status = "complete"
     mr.mark_for_replacement
+    assert mr.marked_for_destruction?
     assert mr.marked_for_replacement?
+  end
 
-    mr.replaced_at = nil
-    refute mr.marked_for_replacement?
+  it "sets both deleted_at and replaced_at on save" do
```
cavis, line 144: Wrote some actual tests that create/delete stuff in the DB.  The stubbed tests weren't really covering this bug.

svevang (approved):

PR #668, opened 2023-06-09
Feat/feed helpers

### LINKED ISSUE #535: Edit Feed

radical-ube:
resolves #535 

this should be last thing needed after the crud PR to bring edit feeds to parity. this PR adds getter/setter methods to address the `audio_format` and `include_zones` fields.


app/javascript/controllers/toggle_field_controller.js
```diff
@@ -22,10 +22,11 @@ export default class extends Controller {
     const field = event.target.value
 
     this.fieldTargets.forEach((el) => {
-      if (field !== el.getAttribute("field")) {
-        el.classList.add("d-none")
-      } else {
+      const types = el.getAttribute("field").split(" ")
```
radical-ube, line 25: this small change is to account for being able to handle display toggling for multiple types. in this case, `mp3`, `wav`, or `flac`.

app/views/feeds/_form_main.html.erb
```diff
@@ -78,37 +78,44 @@
...
       <div class="form-floating">
-        <%= form.select :audio_format, audio_format_options %>
-        <%= form.label :audio_format %>
+        <%= form.select :audio_type, audio_format_options, {include_blank: true}, data: {action: "toggle-field#displayOnlyThisField"} %>
+        <%= form.label :audio_type %>
       </div>
     </div>
 
-    <div class="col-3">
+    <div class="col-3 <%= display_bitrate(feed) %>" data-toggle-field-target="field" field="mp3">
```
radical-ube, line 94: for this and the below fields, 2 things are defined on this div line: initial display class, and display toggle.

the initial display is in `feeds_helper` and is based on `audio_format[:f]` or `audio_format.blank?` depending on the field.

the toggle is triggered by the change in value of `audio_type` (aka `audio_format[:f]`). so this one displays only with `mp3`, bitdepth with either `wav` or `flac`, and the other two with any of the three types. and `d-none` for blank.

app/views/feeds/_form_custom_fields.html.erb
```diff
@@ -73,16 +73,16 @@
...
+  <%= form.check_box :house %>
   <%= form.label :house, class: "me-2" %>
 
-  <%= form.check_box :paid, value: 0 %>
+  <%= form.check_box :paid %>
   <%= form.label :paid, class: "me-2" %>
 
-  <%= form.check_box :sonic_id, value: 0 %>
+  <%= form.check_box :sonic_id %>
   <%= form.label :sonic_id, class: "me-2" %>
```
cavis, line 86: Where did this `System (Sonic IDs)` label come from?  Is that in the designs?  I've never heard a sonic_id zone referred to as "system" before... can't decide how confusing I find it yet.  (That term is overridden by Campaign Type in Augury).

app/controllers/feeds_controller.rb
```diff
@@ -85,6 +85,15 @@ def feed_params
...
+      :audio_type,
+      :audio_bitrate,
+      :audio_bitdepth,
+      :audio_channel,
+      :audio_sample,
+      :billboard,
+      :house,
+      :paid,
+      :sonic_id,
       feed_tokens_attributes: %i[id label token _destroy],
```
cavis, line 97: Unrelated to this PR request: can you sort the Feeds tabs by something?  Maybe just `created_at: :asc`, so they show in the order you created them?  Just a bit jarring to compare with Publish when they're in a different order.

app/models/concerns/feed_ad_zone.rb
```diff
@@ -0,0 +1,77 @@
...
+  end
+
+  def billboard
+    return unless include_zones.present?
+
+    include_zones.include?("billboard") ? 1 : 0
+  end
+
+  def billboard=(val)
+    self.include_zones ||= []
```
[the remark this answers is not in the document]
cavis: And for my 3rd comment in this thread...

Maybe it's better if we deviate this UI from Publish?  Publish doesn't make a ton of sense anyways.  So for the allowed values of `include_zones`:

- When `nil` ... show all 4 boxes checked.  And when the user saves with all 4 checked, set to nil.
- When `[]` ... show all 4 boxes unchecked.
- When `["<name>"]` ... check just those 1, 2, or 3 zones.

That make more sense?  It maintains the existing data as well (it can't be an array with 4 items in it ... it becomes nil instead).

cavis (changes requested):
Okay, generally works good, a few suggestions/comments, and 1 gotcha that needs to be fixed.

I think after these 2 PRs merge, we should do a bit of QA and figure out where this is on parity with Publish.

app/views/feeds/_form_custom_fields.html.erb
radical-ube: yes, i think i pulled verbiage from the design.

app/models/concerns/feed_audio_format.rb
```diff
@@ -0,0 +1,56 @@
...
+
+  def audio_sample
+    audio_format.try(:[], :s)
+  end
+
+  def audio_sample=(sample)
+    return unless audio_format.present?
+
+    audio_format[:s] = sample.to_i
+  end
```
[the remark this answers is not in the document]
radical-ube: tests added below!

app/views/feeds/_form_custom_fields.html.erb
radical-ube: i can change the labels back to parity with Publish and we can bring it up it later.

test/models/concerns/feed_ad_zone_test.rb
```diff
@@ -0,0 +1,118 @@
...
+        feed.include_zones << "sonic_id"
+        assert_equal feed.sonic_id, "1"
+      end
+    end
+  end
+
+  describe "setter methods" do
+    let(:all_zones) { ["billboard", "house", "ad", "sonic_id"] }
+
+    describe "#add_zone" do
```
radical-ube, line 56: this suite and `#remove_zone` are possibly misnomers? since i don't test either of those methods directly (they're both private), but rather through calling each of the `method=` setters. so i'm not sure if i should have gone the more redundant route and just had each of those be their own suites and repeating the same tests in each? thinking about long-term readability if this is the kind of thing that would be confusing to understand a year from now.

cavis: Ah, I see what you're saying.

I think either way is fine - though it is a bit odd to reference a private method here, even as a comment.  Stylistically, I might prefer the more verbose style?  Or write one mega-test, and just loop over all 4 zone names, running the same checks?

Your call.  Definitely multiple ways.

test/models/concerns/feed_ad_zone_test.rb
```diff
@@ -0,0 +1,118 @@
...
+      it "ignores nil" do
+        refute_equal feed.include_zones, all_zones
+        feed.billboard = ("1")
+        refute_equal feed.include_zones, all_zones
+      end
+
+      it "adds to include_zones if zone is checked" do
+        feed.include_zones = []
+        feed.billboard = ("1")
```
cavis, line 65: Why the parenths around all these?

cavis (approved):
👍 looks good to me.

test/models/concerns/feed_ad_zone_test.rb
radical-ube: ended up rewriting the verbose way and i think it reads better.

PR #669, opened 2023-06-13
Convert episode form to tabbed layout

cavis:
Hey look, I had already committed just the tabbed-layout change!


radical-ube (approved):
looks and works as expected

Issue #640, opened 2023-05-09
Episode Categories field autocomplete


cavis:
Make the episode categories picker JS thing an autocomplete!

- [ ] Can we do this with an async request (ideal)?  May depend on how that boostrap-tag-library works
- [ ] Otherwise, I guess pull in the tags ahead of time (slower).  With some sane limit.
- [ ] Cleanup the UI/UX on this field while you're at it.

PR #676, opened 2023-06-21
UX passthrough

brandonhundt:
- Changes how popovers are handled (`<a>` instead of `<button>`) Will now close when you click off of it.
- Changes how help text handled (using `.input-group`) which improves general layout and spacing
- Slight layout spacing (so there's more left/right margin on some of the form pages). The spacing helps the eye focus.


app/views/episodes/_form_tags.html.erb
```diff
@@ -5,15 +5,10 @@
...
-        <div class="form-floating">
-          <%= form.tag_select :categories, episode.categories, {}, data: {allow_new: true, allow_clear: true} %>
-          <%= form.label :categories %>
-        </div>
-
-        <button type="button" class="prx-btn-help" data-bs-toggle="popover" data-bs-content="<%= t(".help.categories") %>">
-          <span class="material-icons">help</span>
-        </button>
+      <div class="form-floating input-group">
+        <%= form.tag_select :categories, episode.categories, {}, data: {allow_new: true, allow_clear: true} %>
```
cavis, line 9: This field still needs some work - bit buggy.  But... pre-existing condition.

cavis (approved):
👍 looks good to me!

app/views/episodes/_form_tags.html.erb
brandonhundt: Write a ticket!

cavis: I think #640 covers this?  Though I disagree with that "low" priority.

cavis:
@brandonhundt i think this needs some conflicts resolved?  But then it's good to go.

PR #678, opened 2023-06-23
Fix case where the source metadata is not reset

svevang:
Fixes a bug where an old (finished) upload was polling for metadata on an expired CDN url.   This sets up a guard, and will only reset and refresh "source" file metadata if needed.


cavis (approved):
👍 looks good to me

PR #679, opened 2023-06-26
Trix editor for wysiwyg editing

kookster:
Copying what was done in id.prx.org:
https://github.com/PRX/id.prx.org/pull/345
https://github.com/PRX/id.prx.org/pull/348


brandonhundt:
The editor works great!

Is it in the scope of this PR to apply to all the fields where its needed?

app/views/episodes/_form_main.html.erb
```diff
@@ -26,11 +26,12 @@
...
 
 <div class="col-12 mb-4 prx-field-group">
-  <div class="form-floating textarea-help">
-    <%= form.text_area :description %>
-    <%= form.label :description %>
+  <div class="form-floating">
+    <div>
+      <%= form.label :description %>
+      <%= form.trix_editor :description %>
+    </div>
```
brandonhundt, line 33: This is going to create a conflict with #676, but I can resolve it in 676.

brandonhundt:
> I think it is - I was going to apply it to podcast descriptions in this PR at least - what other fields require it?

The itunes_summary fields on episode and podcast forms. It would need to be link only.

kookster:
ok, I added this to the desc and itunes:summary fields.
I want to say, I don't think Apple is using that summary any more, maybe we should just get rid of it!
https://help.apple.com/itc/podcasts_connect/#/itcb54353390

cavis (approved):
I can't remember - was there a clean way to importmap this library?  Using sprockets is inconsistent with the other 3rd party stuff in here.  But I suppose I can live with that.

This misses the Feed description/teaser fields.  But can clean that up after, to finish out the ticket.

cavis:
Changed to importmap, and removed itunes:summary entirely.  (The show/hide checkbox didn't play well with Trix, and didn't seem worth sorting out right now).

Issue #589, opened 2023-03-01
Validate episode `description` is present on podcast RSS


svevang:
The `description` field is required for Apple API episode create. Non-blank and Non-null.

Make sure that the RSS has a default value for each episode:

- [ ] helper to validate the presence of the description field
- [ ] provide a default description value for the RSS description field

kookster:
I think we more generally need to look at bringing over validations from CMS -> Feeder models

cavis:
Just noting from our last discussion: Publish doesn't require description at the moment.  And (I think) the CMS RSS-Importer also doesn't require it.  So we either need to add validations/data-migrations there first, or wait until CMS/Publish are deprecated.  Then we can add whatever we want to Feeder validations.

cavis:
We have fallbacks for description now - so calling this good.

PR #681, opened 2023-06-26
Set default description for publishing

svevang:
Partially solves https://github.com/PRX/feeder.prx.org/issues/589 by setting up a default.

Sets up a helper method which wraps the subtitle and title as fallback defaults. This is so that the description is not left blank in the RSS feed.  The motivation here is that Apple validates the presence of the description via interactions to their API. So if the feed is consumed by the RSS poller with a blank description, then unrelated API operations can fail with validation errors.

### DIFF
--- a/app/models/episode.rb
+++ b/app/models/episode.rb
@@ -281,6 +281,10 @@ class Episode < ApplicationRecord
     self.keywords = keywords.map { |kw| sanitize_keyword(kw, kw.length) }
   end
 
+  def description_with_default
+    description || subtitle || title || ""
+  end
+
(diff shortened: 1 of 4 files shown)


app/models/episode.rb
```diff
@@ -281,6 +281,10 @@ def sanitize_text
     self.keywords = keywords.map { |kw| sanitize_keyword(kw, kw.length) }
   end
 
+  def description_with_default
+    description || subtitle || title || ""
```
kookster, line 285: all the fallbacks!

kookster (approved):
lgtm!

PR #682, opened 2023-06-27
Filter out video episodes for Apple episode sync

svevang:
Fixes a case in the subscription sync where a.video episode is in the podcast's feed. Since Apple Podcasts Subscriptions only supports audio content, we need to filter out video content.

### DIFF
--- a/app/models/apple/publisher.rb
+++ b/app/models/apple/publisher.rb
@@ -31,7 +31,15 @@ module Apple
     end
 
     def episodes_to_sync
-      show.episodes
+      filter_episodes(show.episodes)
+    end
+
+    def filter_episodes(eps)
+      # Reject episodes if the audio is marked as uploaded/complete
+      # or if the episode is a video
+      eps
+        .reject(&:synced_with_apple?)
+        .reject(&:video_content_type?)
     end
 
     def poll_all_episodes!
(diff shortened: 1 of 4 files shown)


app/models/apple/publisher.rb
```diff
@@ -31,7 +31,15 @@ def podcast
...
-      show.episodes
+      filter_episodes(show.episodes)
+    end
+
+    def filter_episodes(eps)
+      # Reject episodes if the audio is marked as uploaded/complete
+      # or if the episode is a video
+      eps
+        .reject(&:synced_with_apple?)
+        .reject(&:video_content_type?)
```
kookster, line 42: gotcha
This is filtered out of the Apple specific episode list
so we would still have the video in the public rss feed (in the case of ad free on apple sub)
makes sense

app/models/apple/publisher.rb
```diff
@@ -47,9 +55,6 @@ def poll!(eps = episodes_to_sync)
       end
 
       Rails.logger.tagged("Apple::Publisher#poll!") do
-        # Reject episodes if the audio is marked as uploaded/complete
-        eps = eps.reject(&:synced_with_apple?)
```
kookster, line 51: moved to the `filter_episodes`, makes sense

kookster (approved):
lgtm!

app/models/apple/publisher.rb
cavis: Not sure it matters, as this already merged - but it might be simpler to check the `episode.medium` enum now?  

```ruby
.reject(&:medium_video?)
```

Saves the trouble of iterating over the contents.

PR #686, opened 2023-06-29
Pass in the private feed to get the enclosure filename

svevang:
Fixes a bug where the enclosure filename was not being properly parameterized by the feed.  Handles the case where the file extensions of the main feed don't match the private feed in Apple publishing.


cavis (approved):
👍 good catch!

PR #689, opened 2023-07-06
Consistent card groupings

brandonhundt:
- Uses card headers with no border and a shadow for base cards
- `card-shadow-light` to replace the black (I though the black was too stark), light blue still available for other card headers too.
- Adds cards to the episode planner
- fixes a bug where the `invalid-feedback` message was added before the help text.

In review:
- [ ] Save a blank episode and ensure error state messaging is inline.

### DIFF
--- a/app/views/episodes/_form_tags.html.erb
+++ b/app/views/episodes/_form_tags.html.erb
@@ -1,6 +1,6 @@
 <div class="col-12 mb-4">
   <div class="card shadow border-0">
-    <div class="card-header card-header-secondary d-flex justify-content-between">
+    <div class="card-header card-header-light d-flex justify-content-between">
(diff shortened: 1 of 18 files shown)


app/views/episodes/_form_tags.html.erb
```diff
@@ -1,6 +1,6 @@
 <div class="col-12 mb-4">
   <div class="card shadow border-0">
-    <div class="card-header card-header-secondary d-flex justify-content-between">
+    <div class="card-header card-header-light d-flex justify-content-between">
```
cavis, line 3: On my (admittedly not-great) external monitor, I can barely distinguish this from the background.  Not a huge deal - but I suspect for accessiblility we might need to darken the `$light` a bit.

cavis (approved):
👍 looks good to me!

PR #691, opened 2023-07-06
Feat/podcast imports

### LINKED ISSUE #531: Edit Podcast Imports
After #505 and #536.  Build the Podcast -> Settings -> [Import](https://www.figma.com/proto/Yewv83Q8btHcaOBWCtGgKs/PRX-Dovetail-with-Navigation?page-id=80%3A2947&node-id=2461%3A26796&viewport=798%2C-3032%2C0.25&scaling=min-zoom&starting-point-node-id=80%3A3420&hide-ui=1) tab.

NOTE: the designs for this don't actually represent the work to be done.  It should be more sourced/improved from Publish, but allowing > 1 import on the same podcast, and different types of imports in the future (just RSS to start).

image

- [ ] Setup the routes/tab
  - [x] Use `podcast_imports#index` and `/podcasts/123/imports` to list them
  - [ ] And then full resource CRUD new/show/etc, since there can be multiple
- [ ] Use a bootstrap grid and other bootstrap-y things for the layout.
- [ ] A lot of this is TBD, depending on the schema.
- [x] Use turbo to track progress and refresh a piece of the page.

radical-ube:
resolves #531 

This PR doesn't accomplish everything stated in the ticket, however it should be enough for feature parity with Publish, while allowing for improvements to be iterated upon.

The biggest change in this feature from Publish is that a `PodcastImport` is no longer a top-level feature, nor does it create a podcast on import. Instead, it is a Podcast-level feature, and multiple imports can be made on a single podcast. Episode imports should get assigned to their respective Episodes (if present), and create/update accordingly.


app/controllers/imports_controller.rb
```diff
@@ -1,71 +1,45 @@
...
-  end
-
-  # GET /imports/new
-  def new
-    # @import = Import.new
-  end
-
-  # GET /imports/1/edit
-  def edit
+    @imports = PodcastImport.where(podcast_id: @podcast).order(created_at: :desc)
```
radical-ube, line 6: decided on utilizing the `index` action for both "show" and "new", where a user can both see all of the previous imports, as well as start a new one.

app/controllers/imports_controller.rb
```diff
@@ -1,71 +1,45 @@
...
-
-  # DELETE /imports/1
-  def destroy
-    # @import.destroy
-
-    # respond_to do |format|
-    #   format.html { redirect_to imports_url, notice: "Import was successfully destroyed." }
-    # end
+    # validate RSS
+    # validate URL
```
radical-ube, line 14: i think i was supposed add validations for the RSS feed and URL to make sure they were legitimate? probably should go in the model. but also i wasn't sure what the specifications were for a valid RSS and URL?

app/policies/podcast_import_policy.rb
```diff
@@ -1,4 +1,12 @@
 class PodcastImportPolicy < PodcastPolicy
+  def index?
+    PodcastPolicy.new(token, resource.podcast).new?
```
radical-ube, line 3: `PodcastPolicy` doesn't have an `index?` method, but i figured `new?` was the scope that i wanted here.

app/models/episode_import.rb
```diff
@@ -49,6 +49,8 @@ def import
     set_audio_metadata!
     update!(status: PodcastImport::AUDIO_SAVED)
 
+    lookup_and_assign_episode!
```
radical-ube, line 52: new method to assign already existing episodes when attempting to create another import that has the same/existing episodes on it.

app/models/podcast_import.rb
```diff
@@ -6,7 +6,7 @@ class PodcastImport < ApplicationRecord
 
   serialize :config, HashSerializer
 
-  attr_accessor :feed, :feed_raw_doc, :templates
+  attr_accessor :feed, :feed_raw_doc, :templates, :feed_source
```
radical-ube, line 9: technically unused, but setting up the select dropdown for later when different sources will be implemented.

app/views/imports/_form_main.html.erb
```diff
@@ -0,0 +1,43 @@
+<div class="overflow-auto text-nowrap">
```
radical-ube, line 1: so this differs slightly from the design, but i tried to balance the tasks of the ticket with what we needed for parity with Publish. so i used a table to display imports, and moved the importer over to the `form_status` partial, to keep things all on one view.

app/views/imports/_form_main.html.erb
```diff
@@ -0,0 +1,43 @@
...
+          <th>Feed Url</th>
+          <th>Status</th>
+          <th>Episode Count</th>
+          <th>Import Date</th>
+        </tr>
+      </thead>
+      <tbody>
+        <% if imports.present? %>
+          <% imports.each do |import| %>
+            <% if %w[complete failed].include?(import.status) %>
```
radical-ube, line 15: should this conditional be extracted into a helper? basically the conditional is `PodcastImport.finished?`, to determine whether to keep polling or not (happens in the else branch)

also, i found out that this conditional can be buggy; if i somehow bork the podcast import but somehow the podcast import status doesn't update to `failed`, then a user can theoretically get stuck in the else branch with the poller. (this has happened once in development, though i don't know how to replicate it.)

app/views/imports/_form_status.html.erb
```diff
@@ -0,0 +1,35 @@
+<div class="card shadow border-0">
+  <div class="card-header-success">
+    <h5 class="card-title">Import</h5>
+  </div>
+
+  <div class="card-body row">
+    <div class="col-12 mb-4 prx-field-group">
+      <div class="form-floating">
+        <%= form.select :feed_source, ["RSS"] %>
+        <%= form.label :feed_source, "Feed source" %>
```
radical-ube, line 10: isn't functional yet, but here to hold space for what will eventually be added as a source selector.

app/views/imports/_form_status.html.erb
```diff
@@ -0,0 +1,35 @@
...
+      <button type="button" class="prx-btn-help" data-bs-toggle="popover" data-bs-content="<%= "help" %>"><span class="material-icons">help</span></button>
+    </div>
+
+  </div>
+
+  <div class="card-footer d-flex-align-items-center justify-content-between">
+
+    <div class="col-12 mb-4">
+      <%= form.hidden_field :account_id, value: @podcast.account_id %>
+      <%= form.hidden_field :podcast_id, value: @podcast.id %>
```
radical-ube, line 31: these `hidden_field`s i think were defined in the `index` action because i think they are necessary for a valid `PodcastImport`? is there a better way to handle this?

app/views/imports/_form_main.html.erb
cavis: I think this view makes a lot of sense.  But - I also think we're going to need `podcast_imports#show`, to track the breakdown of episode statuses, show errors, etc.

Doesn't need to be anything real fance... just a turbo-frame with some version of this, and the polling-controller to refresh it until complete.



This is also where you show "duplicate guids" in the RSS feed and any errors (here's [the publish code](https://github.com/PRX/publish.prx.org/blob/main/src/app/series-import-status-card/series-import-status-card.component.html#L34) fwiw)

app/views/imports/_form_main.html.erb
cavis: I think you should add a few enums/helpers to the podcast/episode import models.  Enums are the rails way!

```rb
# these have to be string-style enums, since the db has strings not ints
enum status: {created: "created", started: "started"...}, prefix: true

def finalized?
  status_complete? || status_failed?
end
```

Hrmm... per getting stuck in a non-final status, should probably keep an eye out for that.  May need to rescue something, so we can set the status to failed.
```

File: app/views/imports/_form_status.html.erb (hunk 9 above)
Comment (cavis) [reply app/views/imports/_form_status.html.erb:31]: So I think you should drop `account_id` ... it's just not needed.

And `:podcast_id` shouldn't be a form field.  Once you added a `has_many :podcast_imports` in `podcast.rb`... in your controller you can just do a more standard:

```
@import = @podcast.podcast_imports.new(import_params)
```

File: app/views/imports/_form_main.html.erb (hunk 6 above)
Comment (cavis) [reply app/views/imports/_form_main.html.erb:1]: Alternatively, you could add these to the `#index` table you've got.  But it might look kind of crowded?  Feel free to try it out though.  The things I'm looking for are:

- The entries in RSS feed
- Successfully imported count
- In progress count
- Remaining count
- Errors count
- Duplicate guids count
- And some sort of visual-spinner/progressbar indicating things are still happening

**Review by cavis — CHANGES_REQUESTED** [review]
Looking/working pretty good.  A few things to fix/add, and i18n-ify.  We can talk async about the episode_import statuses/counts, if those are still being weird.

File: app/models/episode_import.rb [hunk 10]
```diff
@@ -183,4 +185,9 @@ def connection(u = uri)
   def account
     podcast_import.try(:account)
   end
+
+  def lookup_and_assign_episode!
+    lookup = Episode.find_by_item_guid(guid)
```
cavis, line 190: Oh... erm... this should be scoped to this single Podcast.  Episode `item_guid` is only unique _within a single Podcast_.  You could have some other podcast with the same episode item_guid.

app/controllers/imports_controller.rb
```diff
@@ -1,71 +1,51 @@
...
-  end
-
-  # GET /imports/new
-  def new
-    # @import = Import.new
-  end
-
-  # GET /imports/1/edit
-  def edit
+    authorize @import
```
radical-ube, line 13: `show` action and view is now implemented.

app/controllers/imports_controller.rb
```diff
@@ -1,71 +1,51 @@
...
-  # DELETE /imports/1
-  def destroy
-    # @import.destroy
-
-    # respond_to do |format|
-    #   format.html { redirect_to imports_url, notice: "Import was successfully destroyed." }
-    # end
+    @import = @podcast.podcast_imports.new(import_params)
+    authorize @import
+    @import.get_feed
```
radical-ube, line 20: added this line as per the last comment re: validating the RSS url, which i realized already [happens here](https://github.com/PRX/feeder.prx.org/blob/05d5462625e84778fc7a610a36acfb80b1b0e82a/app/models/podcast_import.rb#L179), so i'm not sure if this call is actually just redundant.

app/controllers/imports_controller.rb
```diff
@@ -1,71 +1,51 @@
...
+
+    respond_to do |format|
+      if @import.save
+        @import.import_later
+        format.html { redirect_to podcast_import_path(@podcast, @import), notice: "Beginning import." }
```
radical-ube, line 25: now redirects to the specific import view instead of index.

app/helpers/imports_helper.rb
```diff
@@ -0,0 +1,9 @@
+module ImportsHelper
+  def progress(podcast_import)
```
radical-ube, line 2: progress bar logic. took a long time to figure out that i needed to convert to floats to get accurate percentages.

app/models/concerns/import_utils.rb
```diff
@@ -4,6 +4,16 @@
 module ImportUtils
   extend ActiveSupport::Concern
 
+  AUDIO_SAVED = "audio saved".freeze
```
radical-ube, line 7: moved status definitions into this concern, which is shared by both `PodcastImport` and `EpisodeImport`

app/models/episode_import.rb
```diff
@@ -18,25 +18,37 @@ class EpisodeImport < ApplicationRecord
   scope :having_duplicate_guids, -> do
     unscope(where: :has_duplicate_guid).where(has_duplicate_guid: true)
   end
-  scope :complete, -> { where(status: PodcastImport::COMPLETE) }
+  scope :complete, -> { where(status: COMPLETE) }
+  scope :finished, -> { where(status: [COMPLETE, FAILED]) }
+  scope :in_progress, -> { where.not(status: [COMPLETE, FAILED, CREATED]) }
+  scope :failed, -> { where(status: FAILED) }
```
radical-ube, line 24: different scopes needed for various UI conditionals. not sure if there was a better way to go about this but it feels cleaner here than in the views.

app/models/episode_import.rb
```diff
@@ -18,25 +18,37 @@ class EpisodeImport < ApplicationRecord
...
 
   before_validation :set_defaults, on: :create
 
   validates :entry, :guid, presence: true
 
+  enum :status, {
```
radical-ube, line 30: new enums for statuses.

app/models/episode_import.rb
```diff
@@ -18,25 +18,37 @@ class EpisodeImport < ApplicationRecord
...
+
   def unlock_podcast
     if podcast_import.finished?
       podcast_import.podcast.update(locked: false)
     end
   end
 
   def retry!
-    update(status: PodcastImport::RETRYING)
+    update(status: RETRYING)
```
radical-ube, line 46: i'm not sure why these are not the bang version of `update`, so i tried to leave as is in case that was a specific choice.

app/models/episode_import.rb
```diff
@@ -47,24 +59,22 @@ def import_later
 
   def import
     set_audio_metadata!
-    update!(status: PodcastImport::AUDIO_SAVED)
+    status_audio_saved!
```
radical-ube, line 62: anything else that was `update!` rewritten to use enum magic methods.

app/models/episode_import.rb
```diff
@@ -103,7 +113,13 @@ def entry_audio_files(entry)
   end
 
   def create_or_update_episode!
-    self.episode ||= Episode.new(podcast: podcast)
+    lookup = Episode.where(podcast_id: podcast.id).find_by_item_guid(guid)
```
radical-ube, line 116: didn't need this line to split into a different method so moved it here as recommended. now scoped to the specific podcast as well.

app/models/podcast_import.rb
```diff
@@ -6,24 +6,26 @@ class PodcastImport < ApplicationRecord
...
 
   belongs_to :podcast, -> { with_deleted }, touch: true, optional: true, autosave: true
   has_many :episode_imports, -> { where(has_duplicate_guid: false).includes(:podcast_import) }, dependent: :destroy
 
   before_validation :set_defaults, on: :create
 
-  validates :account_id, :url, presence: true
+  validates :url, presence: true
```
radical-ube, line 16: removed the validation on `account_id`, in preparation for its removal (to be done at a different time)

app/models/podcast_import.rb
```diff
@@ -47,13 +49,21 @@ def episode_importing_count
...
+      feed_episode_count - episode_imports.finished.count
+    end
+  end
+
   def status
     return super unless episode_imports.count > 0
     return super if episode_importing_count > episode_imports.count
 
     if complete?
+      status_complete!
```
radical-ube, line 63: ok, so this is related to [this thread](https://prx.slack.com/archives/G0FKLK2Q5/p1689705702150919) where nothing sets the import status as complete/failed at the end of an import. I tried to incorporate an `update_status` method into a hook, but i couldn't figure out what the proper end trigger was. `after_save` `if: :status_changed?` didn't seem to trigger the status to update, also tried `if: :finished?` but that wasn't correct either. so in the meantime i put the status update here, which is admittedly still not the correct place for it.

would placing the status update [after this](https://github.com/PRX/feeder.prx.org/blob/05d5462625e84778fc7a610a36acfb80b1b0e82a/app/models/podcast_import.rb#L115) be the right way to go? but i'm not sure if that method waits for the worker to finish the jobs before moving on? want to avoid the status reading `complete` but the imports are still happening.

test/controllers/imports_controller_test.rb
```diff
@@ -1,47 +1,67 @@
...
-  #     delete import_url(@import)
-  #   end
+  test "authorizes creating imports" do
+    @podcast.update(prx_account_uri: "/api/v1/accounts/456")
+    post podcast_imports_url(@podcast), params: {podcast_import: params}
+    assert_response :forbidden
+  end
 
-  #   assert_redirected_to imports_url
+  # test "validates creating imports" do
```
radical-ube, line 51: for some reason i can't seem to figure out how to hit the branch in the controller where an `import.save` fails and we get the expected response. but the test is written and commented out for now.

app/models/episode_import.rb
cavis: Yeah, that is odd.  I'd be fine with `status_retrying!` here.  It's surprising to have the non-`!` version happening inside a method with a `!` on it.

We'll catch more of these inconsistencies after some in-depth QA, once we hit feature parity! (I hope).

app/models/episode_import.rb
cavis: +7

app/models/podcast_import.rb
cavis: Huh... well maybe let's just open a ticket to sort this out?  Seems a bit complex to iron out entirely here.

The `import_episodes!` method won't actually work for this.  It does create all the `EpisodeImport`s, but it doesn't process them.  It kicks off another async job for each of them in `enqueue_episode_import_jobs`.

So as each of the N `EpisodeImport`s change status (async), we need to re-calculate the `PodcastImport` status.

But yeah, let's open a ticket.

test/controllers/imports_controller_test.rb
cavis: The only validation I see on `PodcastImport` is `validates :url, presence: true`.  And you can't even get to that, because that `get_feed` method throws an error before we even try to save.

So - I think to make this test work, you need to:

- Rescue the `.get_feed` error(s) that you get from bad uris, non-rss, etc.  (May have to trial and error to see what exactly it throws for each of those bad inputs).
- In the rescue, render the flash error and return `:unprocessable_entity` (same as a model validation would have)
- Then you should be able to run this test

cavis (changes requested):
Ooookay... the UI/UX on this is looking good.  (I didn't look for i18n stuff... but sounds like you're on that anyways).

I had a few nitpicks, but the main blockers on merging this are:

1. Handle readonly users in the policy/controller/view (they should be able to index/show PodcastImports, but not create them).  This may change a few tests.
2. Rescue errors thrown by `.get_feed` in the controller, and return a flash error.  Then you can uncomment one of the controller tests.

app/controllers/imports_controller.rb
```diff
@@ -1,71 +1,58 @@
...
-    # end
+    @imports = @podcast.podcast_imports
+    @import = @podcast.podcast_imports.new(import_params)
+    authorize @import
+    begin
+      @import.get_feed
+    rescue URI::InvalidURIError
+      flash.now[:notice] = t(".failure_get_feed")
+      render :index, status: :unprocessable_entity
+      return
```
radical-ube, line 26: `rescue` branch added here. also had to `return` out of this to avoid double rendering error.

cavis (commented):
:+4: nice!  Looks/works good to me.  Onwards!

cavis (approved):

PR #693, opened 2023-07-10
Episode distributions card

### LINKED ISSUE #633: Episode Distribution Card
Implement the Episode distribution card.

image

- [ ] Audio file link should be the dovetail enclosure url
- [ ] Episode url starts out blank, but can magically prefill with the play.prx.org landing page?
- [ ] Embeddable player link same as Publish has today
- [ ] Editable guid

Realizing there's some nuance around the player link when the episode is unpublished.  May need to research how Publish does that currently, and come up with a solution that fits in this smaller UI space.

cavis:
Fixes #633.  Fills in the Episode distribution card.

State of this form varies a bit based on episode state (new, is media uploaded, etc).

This _does not_ do the magic for "previewing" an embed-player before the episode publishes.  But I think that's the right call for this component.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -20,6 +20,10 @@ class ApplicationController < ActionController::Base
     redirect_to PrxAuth::Rails::Engine.routes.url_helpers.new_sessions_path
   end
 
+  def nilify(p)
+    p.transform_values { |v| v.present? ? v : nil }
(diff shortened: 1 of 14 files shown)


brandonhundt (approved):
I adjusted a few minor styles and pushed it up. This is good. Lets Ship it! :shipit:

PR #695, opened 2023-07-10
Episode embed player

### LINKED ISSUE #670: Episode Embeddable Player

cavis:
Fixes #670. Adds the embed player page on the episode.

image

### DIFF
--- a/app/helpers/embed_player_helper.rb
+++ b/app/helpers/embed_player_helper.rb
@@ -4,13 +4,41 @@ module EmbedPlayerHelper
   EMBED_PLAYER_PATH = "/e"
   EMBED_PLAYER_FEED = "uf"
   EMBED_PLAYER_GUID = "ge"
+  EMBED_PLAYER_CARD = "ca"
 
-  def embed_player_episode_url(ep, preview = false)
+  def embed_player_episode_url(ep, type = nil, preview = false)
+    params = {}
+
     if preview && !episode.published?
       # TODO: private token auth url
+      {}
     else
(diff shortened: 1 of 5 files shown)


app/helpers/embed_player_helper.rb
```diff
@@ -4,18 +4,62 @@ module EmbedPlayerHelper
...
 
   private
 
   def embed_params(params)
     "#{play_root}#{EMBED_PLAYER_PATH}?#{params.to_query}"
   end
+
+  def enclosure_with_token(ep)
+    sep = ep.enclosure_url.include?("?") ? "&" : "?"
+    ep.enclosure_url + sep + {DOVETAIL_TOKEN => prx_jwt}.to_query
```
cavis, line 63: Here's the magic that appends a `?_t=<giant-auth-token-string>` to the Dovetail MP3 url.  So you can listen to episodes that are not published yet.

app/views/episode_player/_form.html.erb
```diff
@@ -0,0 +1,64 @@
+<%= turbo_frame_tag "embed-player", data: {controller: "morphdom"} do %>
+
+  <div class="col-12 mb-4">
+    <div class="form-floating input-group" data-controller="dynamic-form">
+      <%= select_tag :embed_player_type, embed_player_type_options(embed_player_type), class: "form-control", data: {action: "dynamic-form#change"} %>
```
cavis, line 5: Because I'm lazy, I used plain form tags instead of the formbuilder here.  But maybe I should have made this a slim-select as well...

cavis:
I've got slack approval of this from @brandonhundt, and would be nice to have merged for demoing.  Onwards!

Issue #685, opened 2023-06-29
Retry Apple uploads


svevang:
The commit https://github.com/PRX/feeder.prx.org/commit/93efce734a4572377f19a1486826f825fed720db was intended handles the case where there is podcast audio uploaded (marked in the container and episode) but the deliveries and delivery files were missing. This was to workaround the fact that if we were using the podcast deliveries and delivery files as sentinels to mark successful upload.

Later, https://github.com/PRX/feeder.prx.org/commit/4678171ebfe775426a910fb031d9f2b6f7d5df1d landed, which effectively marks the episode as "done" so that once audio is successfully uploaded to apple, the [episode never enters into the publishing flow](https://github.com/PRX/feeder.prx.org/blob/230e8527ddabd2a9547aef5567bf94665ae149cc/app/models/apple/publisher.rb#L41).

With https://github.com/PRX/feeder.prx.org/issues/680 it's now clear that using the podcast deliveries as a sentinel at the outset of the publishing flow does not make sense. Instead, we can rely on the [`#synced_with_apple?`](https://github.com/PRX/feeder.prx.org/blob/712f93f0f849d0fc9bde4f4b8e4c60ecdb940f6f/app/models/apple/publisher.rb#L41)  guard to leverage our local state.

If an episode is in the publishing flow, it will be re-processed. So no need to [guard if the delivery is present](https://github.com/PRX/feeder.prx.org/blob/712f93f0f849d0fc9bde4f4b8e4c60ecdb940f6f/app/models/apple/podcast_container.rb#L25).

TODO:
- [x] Set a guard on the Apple::Episode inline with `#synced_with_apple?` but capturing error semantics on the delivery. 
- [x] Filter out episodes with uploaded and processed delivery errors in Apple::Publisher#filter_episodes
- [x] Apple::PodcastContainer resets file source metadata and local delivery state at the outset of the file upload flow, re-uploads audio.

svevang:
If there are episodes with validation errors, this would filter them out of the publishing attempt.  But if there are episodes with transient problems, like an interrupted upload etc, then these could be retried.

Currently we don’t retry anything.  The proposed change here would be a first step in that direction and deal with situations that, currently, have to be cleaned up by hand.

Issue #680, opened 2023-06-26
Handle gone podcast delivery


svevang:
Looking back on finished episodes (`synced_with_apple? == true`), all the episodes over a month in age have missing podcast deliveries. It appears that these resources expire with time and can't be expected to be present

Handle the case where there is a partially uploaded podcast episode and the local podcast delivery references an apple resource that's returning a 404. 

When we lookup deliveries via the podcast container, do a set difference and see what is returned from Apple versus what is persisted locally.  If there are local deliveries that are not persisted in Apple's system, then:

- adapt the api code to handle 404
- log a warning
- remove the local state (delivery / delivery files)

This means that sending episodes that are marked as `Apple::Episode#synced_with_apple?` into the publishing flow could cause them to have their audio re-uploaded (if they're missing their delivery data) .  There is [a guard against this currently](https://github.com/PRX/feeder.prx.org/blob/712f93f0f849d0fc9bde4f4b8e4c60ecdb940f6f/app/models/apple/publisher.rb#L41).

PR #696, opened 2023-07-11
Fix destroy through relations

svevang:
Fixes a bug where the call to `delete_all` via a `has_many through: ..` throws an ORM exception.

This unifies the code paths in the episode publish method with the metadata reset in the Apple::PodcastContainer class method.

Touches on https://github.com/PRX/feeder.prx.org/issues/685 where the assumption is that if we have _missing_ local copies of the podcast delivery apple data, we intend to re-upload.  The presence of the local podcast deliveries metadata indicates that _we_ have records of doing the uploads, even if [Apple discards them](https://github.com/PRX/feeder.prx.org/issues/680)

### DIFF
--- a/app/models/apple/podcast_container.rb
+++ b/app/models/apple/podcast_container.rb
@@ -34,9 +34,7 @@ module Apple
 
         # Back to DTR to pick up fresh arrangements:
         container.reset_source_metadata!(episode)
-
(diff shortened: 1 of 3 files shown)


cavis (approved):
👍 looks good to me!

PR #698, opened 2023-07-11
Field change confirmations

cavis:
Confirms when a bunch of fields are changed, as Publish does.

- Podcast Settings -> Owner
- Podcast Settings -> Complete (only when changing to "complete")
- Feed -> File Name
- Feed -> Slug (non default feeds)
- Feed -> Public Feed URL (different text depending if you're creating/deleting/changing)
- Feed -> New Feed URL (different text depending if you're creating/deleting/changing)
- Episode -> GUID
- Episode -> New (when you have 0 future-drafts planned)

### DIFF
--- a/app/assets/stylesheets/shared/form.scss
+++ b/app/assets/stylesheets/shared/form.scss
@@ -157,6 +157,25 @@
   }
 }
 
+// slightly different layout for checkboxes with help text
+.form-check {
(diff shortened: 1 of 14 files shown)


brandonhundt (approved):
This is good. Nice! Onwards.

PR #700, opened 2023-07-13
Backfill validations

cavis:
Adds a bunch of validations that existed in Publish/CMS.  I fixed all prod data that didn't adhere to these, so shouldn't cause any issues with CMS -> Feeder data syncing.

See [the parity check spreadsheet](https://docs.google.com/spreadsheets/d/1ZidmwH3JUM4ag2eb1i5YWkbMqx2PsUrSAQ8Te6LdoPM/edit#gid=0) for a list of things I've added here.

### DIFF
--- a/app/controllers/feeds_controller.rb
+++ b/app/controllers/feeds_controller.rb
@@ -89,7 +89,7 @@ class FeedsController < ApplicationController
 
   # Only allow a list of trusted parameters through.
   def feed_params
-    params.fetch(:feed, {}).permit(
+    nilify params.fetch(:feed, {}).permit(
       :file_name,
       :slug,
       :title,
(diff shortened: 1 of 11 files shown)


app/controllers/feeds_controller.rb
```diff
@@ -89,7 +89,7 @@ def set_feed
 
   # Only allow a list of trusted parameters through.
   def feed_params
-    params.fetch(:feed, {}).permit(
+    nilify params.fetch(:feed, {}).permit(
```
cavis, line 92: I wrote this helper awhile back, to make sure we store "nil" in the database when you leave a field blank in the UI.  Added to all forms.

config/locales/en.yml
```diff
@@ -38,15 +38,22 @@ en:
...
       models:
         episode:
           attributes:
             segment_count:
               greater_than: Can't be negative
               less_than_or_equal_to: Enter a number less than %{count}
+        podcast:
+          attributes:
+            payment_pointer:
+              invalid: Payment pointers must have the format $prx.wallet.example/abcd1234
```
cavis, line 56: I haven't translated all error messages - but I think they're looking okay?  Maybe we'll identify more to add here, during QA.

brandonhundt (approved):
LGTM Thanks @cavis

PR #704, opened 2023-07-14
Episode categories

cavis:
Fix styling on the Episode -> Categories field (which uses [bootstrap5-tags](https://github.com/lekoala/bootstrap5-tags)).

Unlike slimselect, this library doesn't watch the underlying/hidden `<select>` for changes.  So I had to manually add a bunch of js to get the tag-select `<div>` in sync with the blank/changed/etc state.

Also - with this bootstrap5-tags update, the magical `<div>` you see is rendered _after_ the `<select>` in the DOM.  So I couldn't use any CSS `~` next sibling selectors.  (And previous-sibling hacks via `:has` aren't supported in Firefox yet).


brandonhundt (commented):
This is looking much better. Is there a way to create a new tag when a comma is used?

If not, we should specifically add clear help text (in the card, and not a popover) that says "press enter to create a new category"

Otherwise we may get this:
Screenshot 2023-07-17 at 9 28 46 AM

brandonhundt (approved):
@cavis after looking at Publish, this is parity. :shipit: APPROVED.

PR #707, opened 2023-07-17
Apple retry uploads

### LINKED ISSUE #685: Retry Apple uploads

svevang:
closes #685 

This sets up a series of changes to help with upload retries:

- shifts some of the attention from Delivery + DeliveryFile over to the PodcastContainer. The PodcastContainer's `files` attributes is a proxy to the validation and processing status on the Delivery/DeliveryFile models. It appears that [Apple retains the overall container state and discards the delivery state](https://github.com/PRX/feeder.prx.org/issues/680), so likely building transitions in a state machine around the PodcastContainer will be better in the long run.
- Fixes a case where the `duplicate` status in the delivery was treated as an error and not properly covered by the success/error conditionals
- expands a `skip_delivery?` method and related helpers on the container to clarify retry logic.


app/models/apple/episode.rb
```diff
@@ -439,11 +441,7 @@ def synced_with_apple?
     end
 
     def waiting_for_asset_state?
-      (podcast_delivery_files.length > 0 &&
-        podcast_delivery_files.all?(&:delivered?) &&
-        podcast_delivery_files.all?(&:processed?) &&
-        !podcast_delivery_files.all?(&:processed_errors?) &&
-        !audio_asset_state_finished?)
+      podcast_container.delivery_settled? && !audio_asset_state_finished?
```
svevang, line 444: so now only looking the episode's `audio_asset_state` and at the podcast container (which manages / coordinates the deliveries).

app/models/apple/podcast_container.rb
```diff
@@ -133,18 +131,14 @@ def self.upsert_podcast_container(episode, row)
     end
 
     def self.get_podcast_containers_via_episodes(api, episodes)
-      # Only query for episodes that don't have a podcast container
-      # The container. Assume that if we have a container record, we don't need to poll.
-      eps_without_container = episodes.filter { |ep| ep.apple_persisted? && ep.missing_container? }
-
```
svevang, line 139: Trying to move special guards and state checks into more top level concerns (like `PodcastContainer#delivery_settled?`)

app/models/apple/publisher.rb
```diff
@@ -114,6 +114,10 @@ def wait_for_upload_processing(eps)
         pdfs = eps.map(&:podcast_delivery_files).flatten
 
         Apple::PodcastDeliveryFile.wait_for_delivery_files(api, pdfs)
+
+        # Get the latest state of the podcast containers
+        # which should include synced files
+        Apple::PodcastContainer.poll_podcast_container_state(api, eps)
```
svevang, line 120: Again, missing local state, this time in the `files` Apple attribute -- which gives a readout of the successfully processed/delivered files.

app/models/apple/podcast_delivery_file.rb
```diff
@@ -327,14 +327,14 @@ def delivered?
     def processed_errors?
       return false unless asset_processing_state.present?
 
-      processed_validation_failed? || processed_duplicate?
```
svevang, line 330: Uploading a duplicate file is not really an error.

app/models/apple/episode.rb
```diff
@@ -116,7 +116,7 @@ def self.update_audio_container_reference(api, episodes)
       episode_bridge_results
     end
 
-    def self.remove_audio_container_reference(api, show, episodes)
+    def self.remove_audio_container_reference(api, show, episodes, apple_mark_for_reupload: true)
```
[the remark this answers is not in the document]
cavis: +1 i can see that being useful from the ol rails console.

app/models/apple/episode.rb
```diff
@@ -382,9 +386,7 @@ def drafting?
     end
 
     def apple_upload_complete?
-      pdfs = feeder_episode.apple_podcast_delivery_files
-
-      pdfs.present? && pdfs.to_a.flatten.all?(&:apple_complete?)
+      feeder_episode.apple_podcast_container.skip_delivery?
```
cavis, line 389: Bit odd conceptually (the upload is complete when we're skipping the delivery?)  Though I guess it's all the same - we skip the delivery when the upload is complete.  Same check.

app/models/apple/podcast_delivery_file.rb
cavis: +1

cavis (approved):
👍 looks good to me!

PR #710, opened 2023-07-18
Wait for timeouts raise exception

svevang:
Bug/TODO fix.  This makes sure that if the remote apple state polling _times out_, then an exception is raised and the publishing flow halts.


cavis (approved):
👍

PR #711, opened 2023-07-18
Handle castle not found error

cavis:
For new podcasts (or in local dev, all new podcasts you create) ... Castle will for some time have no data.

Handle this, and display a nicer message:

image


brandonhundt (approved):
This is great. Thanks @cavis

PR #713, opened 2023-07-19
Extract the apple key out from the config

### LINKED ISSUE #672: Database dumps sanitize Apple::Config credentials
- [ ] re-extract the apple credentials from the apple config
- [ ] skip downloading the apple credentials data with each database dump

svevang:
Closes #672 

This PR:
- extracts the key from the Apple::Config as an optional relation
- prevents the keys from being downloaded via the db dumping script.

### DIFF
--- a/app/models/apple/api.rb
+++ b/app/models/apple/api.rb
@@ -18,16 +18,16 @@ module Apple
     end
 
     def self.from_apple_config(apple_config)
-      if apple_config.no_apple_credentials?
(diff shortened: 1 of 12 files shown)


cavis (approved):
👍 looks good to me!

I wish we had an easier way to obfuscate columns from database dumps.  This works, but it won't be feasible in every case to extract the field into a separate table.  I guess we'll tackle that later (and it's mostly a MySQL db problem for us).

PR #715, opened 2023-07-20
Thoughts?? Sticky sidebars on forms

brandonhundt:
This is a proposal. Merging is accepting. Feedback welcomed.

- I've gotten tired of scrolling back up to save. Makes right rail sticky for desktop screen widths. Moved any image boxes to the main form since they were really big (consistent with episode form). Distribution box stays put (its well sized)
- Sticky form tabs, it will be easier to go from the audio file selection to the audio segmentor tab. 
- Sticky to bottom "add new feed button"


app/views/feeds/_tabs.html.erb
```diff
@@ -1,5 +1,5 @@
 <% content_for :tabs do %>
-  <div class="list-group d-flex col">
+  <div class="list-group d-flex col prx-fixed-btn-container">
```
brandonhundt, line 2: really all this does is add a very large margin to the bottom so you can still read all the tabs if the Add feed button is overlaying it.

app/views/episodes/_form.html.erb
```diff
@@ -15,8 +15,10 @@
       </div>
     </div>
     <div class="col-lg-4 d-grid align-content-start gap-3">
+      <div class="sticky-lg-top">
       <%= render "form_status", episode: episode, form: form %>
       <%= render "form_distribution", episode: episode, form: form %>
```
cavis, line 20: Needs indent?  (I wish our linter would do that automatically...)

PR #718, opened 2023-07-25
Non sticky modals

cavis:


brandonhundt (approved):
LGTM thanks for the assist! Much cleaner.

PR #719, opened 2023-07-26
Podcast planner cards

cavis:
Rewire the podcast planner a bit to:

- Simplify the interface (reduce the number of fields, using some JS and controller trickery)
- Reorganize into 3 cards/steps
- Add episode medium and "ad breaks" fields (instead of "num segments")

image


app/controllers/podcast_planner_controller.rb
```diff
@@ -34,21 +34,51 @@ def set_beginning_of_week
     Date.beginning_of_week = :sunday
   end
 
-  # Only allow a list of trusted parameters through.
+  # Make params match PodcastPlanner interface
```
cavis, line 37: Here's the first bit of magic.  I swapped several fields out, and instead "infer" them here.  Translating into the existing PodcastPlanner interface.

app/controllers/podcast_planner_controller.rb
```diff
@@ -34,21 +34,51 @@ def set_beginning_of_week
...
+      :medium,
+      :ad_breaks,
       selected_days: [],
-      monthly_weeks: [],
+      selected_weeks: [],
       selected_dates: []
     )
   end
+
+  def monthly_week_options(selected_weeks)
```
cavis, line 73: So I'm keeping the unique keys from the helper (`:first` / `:every_two` / etc) as the values in the selects.  And here, I translate them back into integers, as the planner model expects.

app/helpers/podcast_planner_helper.rb
```diff
@@ -1,32 +1,20 @@
 # frozen_string_literal: true
 
 module PodcastPlannerHelper
-  PERIODIC_WEEKS = I18n.t([:every_one, :every_two, :every_three, :every_four], scope: [:podcast_planner, :helper, :period_options])
-  MONTHLY_WEEKS = I18n.t([:first, :second, :third, :fourth, :fifth], scope: [:podcast_planner, :helper, :monthly_options])
```
cavis, line 5: Translating these in the helper wasn't working in staging for some reason?  They always showed up as translation-missing.  So converted to doing them just-in-time here.

app/javascript/controllers/date_controller.js
```diff
@@ -1,22 +1,6 @@
 import { Controller } from "@hotwired/stimulus"
 
 export default class extends Controller {
-  static targets = ["preselected"]
-
-  connect() {
-    if (!this.isDisabled(this.preselectedTarget)) {
-      if (this.isDraft(this.preselectedTarget)) {
-        this.element.classList.add("bg-danger", "text-light")
```
cavis, line 9: This logic moved into the helper `calendar_day_tag` method, so it gets rendered server-side now.

app/javascript/controllers/planner_controller.js
```diff
@@ -0,0 +1,50 @@
+import { Controller } from "@hotwired/stimulus"
+
+export default class extends Controller {
```
cavis, line 3: I broke this out from the `count_controller.js`... because there was just too much custom stuff going on here.

This handles both updating the "create N drafts" button text, and enabling/disabling it based on if you've selected at least 1 date and a num_ad_breaks.

app/javascript/controllers/slim_select_controller.js
```diff
@@ -23,6 +23,28 @@ export default class extends Controller {
           }
           this.element.dispatchEvent(new Event("blur"))
         },
+        beforeChange: (newOpts, oldOpts) => {
```
cavis, line 26: Here's the UX magic.  Since I combined all the "weeks of the month" options into 1 select ... some of those are exclusive.

So when you select "Every two weeks", this deselects anything else.

app/javascript/controllers/toggle_blank_controller.js
```diff
@@ -0,0 +1,21 @@
+import { Controller } from "@hotwired/stimulus"
+
+export default class extends Controller {
+  static targets = ["toggle"]
+
+  toggleTargetConnected(el) {
```
cavis, line 6: A light controller to clear the "Number of Episodes" / "End Date" fields, when you enter a non-blank value in the other one.

app/views/podcast_planner/_form_draft.html.erb
```diff
@@ -0,0 +1,68 @@
+<% data = {controller: "planner", planner_dates_count_value: @planner.dates&.count.to_i} %>
+<%= form_with(url: podcast_planner_path(@podcast), method: :post, data: data) do |form| %>
```
cavis, line 2: Ported / modified from `_form_side.html.erb`.

app/views/podcast_planner/_form_calculator_dates.html.erb
```diff
@@ -0,0 +1,43 @@
+<div class="col-6 mb-4">
```
cavis, line 1: Ported/modified from `_form_date_calculator.html.erb`.

app/views/podcast_planner/_form_calculator.html.erb
```diff
@@ -0,0 +1,20 @@
+<%= form_with(url: podcast_planner_path(@podcast), method: :get, data: {controller: "click"}) do |form| %>
```
cavis, line 1: Ported/modified from `_form_main.html.erb`.  Trying to clarify the purpose of this form a bit (to "calculate" the draft form).

config/locales/en.yml
```diff
@@ -498,55 +500,56 @@ en:
...
+        month: Month
+      title:
+        calendar: Production Calendar
+        settings: Draft Settings
+    form_draft_settings:
       help:
-        preview: Select more options to get a schedule preview.
+        ad_breaks: Do we need this?
+        medium: Do we need this?
+        publish_time: Do we need this?
```
cavis, line 536: I ported a bunch of the text from the previous version into help icons/popovers.  But probably need to audit what's here, and what's useful.

app/views/podcast_planner/_form_draft.html.erb
```diff
@@ -0,0 +1,68 @@
...
+
+  <div class="col-12 mb-4">
+    <div class="card shadow border-0">
+      <div class="card-header card-header-info d-flex justify-content-between">
+        <h5 class="card-title flex-grow-1"><%= t(".title.settings") %></h5>
+        <%= help_text t(".help.settings") %>
+      </div>
+
+      <div class="card-body pt-4 pb-0" data-morph="false">
```
radical-ube, line 11: i'm assuming this `data-morph="false"` is to cancel out the morphdom controller that is above this partial, in order to utilize its own morphdom controller in the turbo frame below? does this mean you can't have a morphdom controller within a morphdom controller (like the parent still overrides the child or something)? or does it mean that we want to control the elements that get changed/updated within this element independently from changes made outside of this element? i still need to read up on morphdom as i still struggle to understand how to use it.

config/locales/en.yml
```diff
@@ -171,6 +171,8 @@ en:
           true: Serial
         subtitle: Teaser
         title: Title
+      podcast_planner:
+        selected_days: asanoteuh
```
radical-ube, line 175: where does this show up?

app/javascript/controllers/slim_select_controller.js
radical-ube: is the returned boolean what deselects/clears the existing values? so, `beforeChange` (before the value changes) we want to set the new value and clear whatever has been selected? I'm not familiar with the inner workings of the SS controller.

radical-ube (changes requested):
just a few curiosity questions, but otherwise i think this still works as expected. just need the verbiage on those help labels.

config/locales/en.yml
radical-ube: verbiage for these labels? @brandonhundt

app/javascript/controllers/slim_select_controller.js
cavis: Yep - [returning false from beforeChange](https://slimselectjs.com/events#beforeChange) prevents the change from happening.  Which is okay in those 2 special cases, since I'm manually changing `setSelected(...)` myself.

app/views/podcast_planner/_form_draft.html.erb
cavis: Yes - this prevents the parent `turbo_frame_tag "podcast-planner"` (all the way up in `show.html.erb`) from blowing away any changes you've made to the selection.  Without this, every time you made changes in the calculator, it would reset these format/ad-breaks/time-of-day fields.

That `data-morph` thing is something I added in `morphdom_controller.js`.  As morphdom is traversing the DOM from top to bottom, if it hits any element with `data-morph=false`, it stops making updates for that element and its children.

config/locales/en.yml
cavis: Whoops!  Leftover from some testing.  I think I was trying to get rails to do its magical translation of model fieldnames.  But I never did get it working.

radical-ube (approved):
looks good works good!

PR #721, opened 2023-07-26
Podcast.publish! on change

cavis:
Oversight: forgot to call `podcast.publish!` to write the updated RSS feed.

This logic was in several other places:

1. All the API controllers [already did this](https://github.com/PRX/feeder.prx.org/blob/main/app/controllers/api/podcasts_controller.rb#L9), calling `publish!` on the podcast in an after_action hook
    - TODO: should the episode APIs be publishing the episode, instead of the podcast???
2. Feed images will [publish! the podcast](https://github.com/PRX/feeder.prx.org/blob/master/app/models/concerns/feed_image_file.rb#L24) when the status becomes complete after processing.
3. Episode contents (audio/video) will [publish! the episode](https://github.com/PRX/feeder.prx.org/blob/master/app/models/content.rb#L45) when the status becomes complete after processing.
4. Episode images will [publish! the episode](https://github.com/PRX/feeder.prx.org/blob/master/app/models/episode_image.rb#L22) when the status becomes complete after processing.
5. The [release episodes cron job](https://github.com/PRX/feeder.prx.org/blob/main/app/jobs/release_episodes_job.rb#L5) will look for any scheduled episodes with [published_at < updated_at](https://github.com/PRX/feeder.prx.org/blob/main/app/models/episode.rb#L82) and call `.publish` on the podcast.
    - TODO: that updated_at logic makes my head hurt

### DIFF
--- a/app/controllers/feeds_controller.rb
+++ b/app/controllers/feeds_controller.rb
@@ -26,6 +26,7 @@ class FeedsController < ApplicationController
     respond_to do |format|
       if @feed.save
         @feed.copy_media
+        @feed.podcast.publish!
         format.html { redirect_to podcast_feed_path(@podcast, @feed), notice: t(".success", model: "Feed") }
       else
         format.html do
@@ -44,6 +45,7 @@ class FeedsController < ApplicationController
     respond_to do |format|
       if @feed.save
         @feed.copy_media
+        @feed.podcast.publish!
         format.html { redirect_to podcast_feed_path(@podcast, @feed), notice: t(".success", model: "Feed") }
       else
         format.html do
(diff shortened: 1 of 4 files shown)


app/controllers/feeds_controller.rb
```diff
@@ -58,6 +60,7 @@ def update
   def destroy
     respond_to do |format|
       if @feed.destroy
+        @feed.podcast&.publish!
```
svevang, line 63: This leaves the RSS file in S3, right?  Does DTR handle that by looking at the Feeder API and returning 404s?

svevang (approved):

app/controllers/feeds_controller.rb
cavis: Yeah, currently this doesn't do anything.

But I was thinking in the future, we may decide to "do things".  Delete the RSS from S3, or redirect, or something?  This at least sets up the hook, and brings these to parity with the API controllers (which have after-destroy hooks).

PR #722, opened 2023-07-26
Gone podcast delivery metadata

svevang:
closes https://github.com/PRX/feeder.prx.org/issues/680

Sets up a way to filter responses that have a 404 status.

In the case of polling, we just want to make sure we have the most up-to-date copy of the remote API state. If the delivery or delivery files are missing on the remote state, we can still use our local copies as sentinels e.g. in `Apple::PodcastContainer#delivery_settled?` and `Episode#apple_mark_for_reupload`

### DIFF
--- a/app/models/apple/api.rb
+++ b/app/models/apple/api.rb
@@ -3,7 +3,7 @@
 module Apple
   class Api
     ERROR_RETRIES = 3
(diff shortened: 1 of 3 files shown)


cavis (approved):
👍 makes good sense

PR #724, opened 2023-07-27
Sets up Apple::Api inspect

svevang:
- Don't print out Apple::Api instance variables by default
- More better exception messaging for missing keys


cavis (approved):
👍

PR #726, opened 2023-07-27
Feed distribution card and other fields

cavis:
Ticking off some Publish parity items on the Feeds page:

1. Adds a Podcast "Distributions" box
    - Easily copyable ID and "public/published URL".  Plus ability to set the Homepage (landing page default still TODO)
2. Reorder episode "Distributions" box to match podcast (guid comes first)
3. Adds a Feed "Distributions" box, and move the most distribute-y fields here
4. Fix some hiding/showing of fields based on the public/private checkbox
5. Ability to add/remove multiple tokens at once
6. Adds missing "Include episode categories" and "engagement settings checkboxes"

image


brandonhundt (commented):
Noting: The "open a new" link for the public feed URL always opens a page in beta regardless of the value.

brandonhundt:
@cavis I pushed up two very minor display tweaks.

- Removes the underline on the token btn-link
- Adjusts styling on tags-badge so they are a little bit bigger. They were hard to read so small.

brandonhundt (approved):
@cavis LGTM onwards!

PR #728, opened 2023-07-31
Retryable media

cavis:
There was a bug in "retrying" audio/image processing.  Where it would (A) poll forever, and (B) submit dozens of processing jobs to retry instead of just one.

This fixes that - you can test by uploading with no worker running.  After ~100 seconds, you get a retry button that will fire off another Porter job.

### DIFF
--- a/app/javascript/controllers/polling_controller.js
+++ b/app/javascript/controllers/polling_controller.js
@@ -1,15 +1,15 @@
 import { Controller } from "@hotwired/stimulus"
 
 const DEFAULT_DEBOUNCE = 1000
+const DEFAULT_MAX = 120
 
 export default class extends Controller {
-  static values = { debounce: Number }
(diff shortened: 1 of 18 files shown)


app/javascript/controllers/polling_controller.js
```diff
@@ -1,15 +1,15 @@
 import { Controller } from "@hotwired/stimulus"
 
 const DEFAULT_DEBOUNCE = 1000
+const DEFAULT_MAX = 120
```
cavis, line 4: Added a new setting to the polling controller, so it'll stop after 2 minutes.

Combined with the `upload_stalled?` helper method, we can mark these unresponsive files as "errored", and let you retry processing them.

brandonhundt (approved):
The spinner is strange, but otherwise this worked as described.

PR #729, opened 2023-07-31
Generate default links

cavis:
Generates default links for podcasts/episodes.

So if you clear the Podcast `Homepage Link` or the episode `Episode URL` ... it will change it back to a play.prx.org `/listen` page.  (NOTE: those `/listen` links won't work for unpublished episodes).


brandonhundt (approved):
@cavis LGTM!!

PR #734, opened 2023-08-02
A11y pass #1

brandonhundt:
This fixes base issues. Still some more to get, but we should get them later as these issues persist in Augury

- [ ] Lighthouse doesn't like slim-select dropdowns
- [ ] Missing label in Trix Editor


cavis (approved):
👍 looks good to me!

Issue #735, opened 2023-08-02
Partition SyncLog and the Container + Delivery models on the apple_config_id


svevang:
Currently the data model supports a has_many apple_configs. However the underlying delegated delivery models are not currently partitioned on the apple_config id.

- [ ] add the apple_config_id to sync_logs, containers, deliveries, and delivery_files
- [ ] partition `has_many` lookups via a method on the association. Bisect the application logic to make sure nothing is calling the unadorned `has_many` -- make sure everything is useing the lookups.
- [ ] Update the `create_or_update` to patch in the apple_config_id -- likely as a top level parameter throughout the apple_publisher
- [ ] update the unique constraints to allow for the partitioning apple_config_id

PR #736, opened 2023-08-02
Apple config tracks apple data

svevang:
Sets up a convenience method to get at all the apple data in a single tree.  Related to https://github.com/PRX/feeder.prx.org/issues/735
This allows for a handle on the data as a means of resetting the delegated delivery state for a fresh sync.


cavis (approved):
👍

PR #737, opened 2023-08-02
Increase audio asset state wait timeout

svevang:
Seeing some timeouts on the audio asset processing.  For the small sample size we're pretty consistently bumping into the timeout.

This PR:

- Increases the timeouts to help more publishing runs complete without error
- Adds logging around the waiting/polling routine, so that we can characterise the distribution of the wait times.

### DIFF
--- a/app/models/apple/episode.rb
+++ b/app/models/apple/episode.rb
@@ -12,11 +12,14 @@ module Apple
     AUDIO_ASSET_FAILURE = "FAILURE"
     AUDIO_ASSET_SUCCESS = "SUCCESS"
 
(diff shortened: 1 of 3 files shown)


app/models/apple/episode.rb
```diff
@@ -12,11 +12,14 @@ class Episode
     AUDIO_ASSET_FAILURE = "FAILURE"
     AUDIO_ASSET_SUCCESS = "SUCCESS"
 
+    EPISODE_ASSET_WAIT_TIMEOUT = 8.minutes.freeze
```
cavis, line 15: 5-8 minutes to process files?  Yeesh, thanks apple.

cavis (commented):
👍 looks good to me!

PR #740, opened 2023-08-02
Show skips updates via API

svevang:
Since the API has a more stringent set of requirements on the required fields,
and we don't specifically need to do updates via the API,
we can skip the updates from the Apple::Publisher publish routine.

For now lets skip any updates to the show and figure out if we really need to do that via https://github.com/PRX/feeder.prx.org/issues/739


cavis (approved):
Makes sense for now.  And maybe forever.

PR #742, opened 2023-08-03
Fix guard for podcast container

svevang:
Fixes two issues:

- When polling remote data, an episode with missing delivery models necessarily has a settled delivery and falls back on the container to check if the file is present.
- Guards against a nil podcast container for episodes


cavis (approved):
👍

PR #747, opened 2023-08-07
Filter out archived episodes

svevang:
It's not possible to update some attributes of the episode when it's in an archived state so filter out the archived episodes before entering into the publishing flow.

### DIFF
--- a/app/models/apple/episode.rb
+++ b/app/models/apple/episode.rb
@@ -388,6 +388,10 @@ module Apple
       apple_json&.dig("attributes", "publishingState") == "DRAFTING"
     end
 
+    def archived?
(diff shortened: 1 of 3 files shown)


app/models/apple/episode.rb
```diff
@@ -388,6 +388,10 @@ def drafting?
       apple_json&.dig("attributes", "publishingState") == "DRAFTING"
     end
 
+    def archived?
+      apple_json&.dig("attributes", "publishingState") == "ARCHIVED"
```
cavis, line 392: What does "archived" mean?  Is that some action someone took manually in apple connect?  Or an episode that fell out of the public RSS?  Or something else?

cavis (approved):
👍

PR #750, opened 2023-08-08
Retry publishing attempts

### LINKED ISSUE #714: Retry Publishing Pipelines
Based on the work in https://github.com/PRX/feeder.prx.org/issues/664

We need a retry mechanism to ensure that if a publishing attempt falls into error, then we can retry and complete the publishing of the episode(s). Currently we're relying on [redundant writes of the RSS to S3  ](https://github.com/PRX/feeder.prx.org/pull/716).

The idea is to hook into the publishing 'tick' -- the same spot as the dead pipeline reaper, so that we can retry a publishing error

- Use the tick-based system and set up the retries adjacent to the reaper logic.
- Log errors upon initiating a retry.
- Pass in the _intended_ queue item id.
- Log errors if the intended queue item does not match the current publishing pipeline's queue item.  Return early without doing any publishing work in that case.

svevang:
closes #714 

Sets up a retry mechanism that tests if the latest (most current) publishing pipeline has failed. If so, a new publishing publishing item is created and a new publishing pipeline is attempted.

### DIFF
--- a/app/jobs/publish_feed_job.rb
+++ b/app/jobs/publish_feed_job.rb
@@ -7,12 +7,17 @@ class PublishFeedJob < ApplicationJob
 
   attr_accessor :podcast, :episodes, :rss, :put_object, :copy_object
 
-  def perform(podcast, publishing_queue_item)
-    # Since we don't current have a way to retry failed attempts,
-    # and until somthing akin to https://github.com/PRX/feeder.prx.org/issues/714 lands
-    # the RSS publishing is extracted from the publishing pipeline semantics.
-    # TODO: recombine the publishing invocations (RSS, Apple) once we have some retry guarantees
-    podcast.feeds.each { |feed| save_file(podcast, feed) }
+  def perform(podcast, pub_item)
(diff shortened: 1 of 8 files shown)


app/jobs/publish_feed_job.rb
```diff
@@ -7,12 +7,10 @@ class PublishFeedJob < ApplicationJob
 
   attr_accessor :podcast, :episodes, :rss, :put_object, :copy_object
 
-  def perform(podcast)
-    # Since we don't current have a way to retry failed attempts,
-    # and until somthing akin to https://github.com/PRX/feeder.prx.org/issues/714 lands
-    # the RSS publishing is extracted from the publishing pipeline semantics.
-    # TODO: recombine the publishing invocations (RSS, Apple) once we have some retry guarantees
-    podcast.feeds.each { |feed| save_file(podcast, feed) }
```
svevang, line 15: Returns the RSS publishing to it's previous place in the partitioned concurrency semantics. So now Apple publishing and delivery will block RSS, for those podcasts with that configured.

cavis (approved):
👍 looks good to me

PR #752, opened 2023-08-09
Disable AR query caching when grabbing the publishing lock 

svevang:
This PR fixes a situation where we can get a stale read via AR query caching when we are probing of the current publishing item and state.

The uniqueness constraints here would prevent a duplicate state transition, but until https://github.com/PRX/feeder.prx.org/pull/752 the publishing attempts could land in an 'error' status because of those constraints.


cavis (approved):
👍

PR #754, opened 2023-08-09
Required fields

cavis:
Adds required asterisk indicator to fields.

Also fixed a minor bug of sort-ordering on the podcast#show "Next Scheduled or Draft" card.


brandonhundt (approved):
@cavis this LGTM. Thanks!

PR #759, opened 2023-08-11
Guard for no current or scheduled publishing items

svevang:
Bug fix for the cases where:

- the job is started but there is no current publishing item to compare (which would itself be an error condition)
- the in-flight jobs via https://github.com/PRX/feeder.prx.org/pull/750 may not have had a pub item supplied as top level params.

### DIFF
--- a/app/models/publishing_pipeline_state.rb
+++ b/app/models/publishing_pipeline_state.rb
@@ -127,6 +127,8 @@ class PublishingPipelineState < ApplicationRecord
         else
           PublishFeedJob.perform_now(podcast, latest_unfinished_item)
         end
+
+        latest_unfinished_item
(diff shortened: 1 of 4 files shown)


app/models/publishing_pipeline_state.rb
```diff
@@ -127,6 +127,8 @@ def self.attempt!(podcast, perform_later: true)
         else
           PublishFeedJob.perform_now(podcast, latest_unfinished_item)
         end
+
+        latest_unfinished_item
```
svevang, line 131: Better to return a handle on the `publishing item -> pipeline` pair.

cavis (approved):
👍 looks good to me!

PR #760, opened 2023-08-11
Handle unpublish (update) from deleted story

svevang:
Seeing some 404s from CMS staging. The story unpublish (update) worker action is invoked after the story (CMS) and episode (Feeder) are deleted.

This PR guards against a gone story in the story update worker, logging a warning in that case.


cavis (approved):
👍 good catch!  Kinda surprised it was already guarding for this, but just not in the update.

PR #763, opened 2023-08-14
Feat/podcast language

### LINKED ISSUE #635: Podcast languages dropdown
I skipped the "Language" dropdown on the Podcast form the first time around.

I'm actually not entirely sure what are valid values in there?  (What standard is it using, anyways?)  Publish does something I'm not confident is correct.

radical-ube:
resolves #635 

So the task was to include all the current codes, but only allow selection based on what was RSS Language Code compliant. So, I took the codes from the `langmap` module that Publish uses and made those as options, but took the RSS Language Code codes and used them to determine which codes were non-compliant; the non-compliant codes still render if they have previously been chosen, but are otherwise disabled for selection.


app/helpers/podcasts_helper.rb
```diff
@@ -75,4 +81,8 @@ def podcast_retry_image_path(podcast, feed_form, image_form)
       edit_podcast_path(podcast, uploads_retry_params(feed_form, image_form))
     end
   end
+
+  def rss_language_options
```
cavis, line 85: So I'd say just leave `NON_COMPLIANT_CODES` out of the dropdown entirely.  Avoid any confusion from the user as to why they're there.

And if you pass the podcast into this method, you can check:

```ruby
if RSS_LANGUAGE_CODES.include?(podcast.language)
  # return just the language_codes
else
  # return the language_codes.concat podcast.language
end
```

cavis (changes requested):

cavis (approved):
👍 looks good to me!

PR #764, opened 2023-08-15
Set up a helper method to onboard podcasts

svevang:


app/models/apple/config.rb
```diff
@@ -2,6 +2,10 @@
 
 module Apple
   class Config < ApplicationRecord
+    DEFAULT_FEED_SLUG = "apple-delegated-delivery-subscriptions"
+    DEFAULT_TITLE = "Apple Delegated Delivery Subscriptions"
```
cavis, line 6: +1 on the scary-enough-that-the-user-probably-won't-edit-this name.

app/models/apple/config.rb
```diff
@@ -27,6 +31,41 @@ class Config < ApplicationRecord
                                          message: "can only have one credential per public and private feed"}
     validates :public_feed, exclusion: {in: ->(apple_credential) { [apple_credential.private_feed] }}
 
+    def self.build_private_feed(podcast)
```
cavis, line 34: Any case for erroring out if one already exists?  Or maybe... updating its attributes if it already exists?  Kind of a `build_or_update_private_feed`?

For the moment at least, we probably want just 1 of these per podcast.

cavis (approved):
Looks good to me.  One thought on a possible idempotent-ify change.

app/models/apple/config.rb
svevang: For sure!  Also, I added some loud warnings so that we can be aware, at least until https://github.com/PRX/feeder.prx.org/issues/793 lands, which should stem the danger in having multiple apple_config rows per podcast.

PR #786, opened 2023-08-16
Add option to exclude podcast metadata during import

### LINKED ISSUE #778: Option to exclude podcast metadata from import
Add an "include podcast metadata" checkbox to the podcast import page.

Default to checked.  But if they uncheck it - only import episodes, without altering podcast metadata.

radical-ube:
resolves #778

### DIFF
--- a/app/controllers/imports_controller.rb
+++ b/app/controllers/imports_controller.rb
@@ -28,7 +28,7 @@ class ImportsController < ApplicationController
 
     respond_to do |format|
       if @import.save
-        @import.import_later(ActiveRecord::Type::Boolean.new.cast(params[:podcast_import][:import_metadata]))
+        @import.import_later(import_metadata)
         format.html { redirect_to podcast_import_path(@podcast, @import), notice: t(".success") }
       else
         format.html do
@@ -54,5 +54,9 @@ class ImportsController < ApplicationController
     params.fetch(:podcast_import, {}).permit(
       :url
     )
+  end
+
+  def import_metadata
+    ActiveRecord::Type::Boolean.new.cast(params[:podcast_import][:import_metadata])
   end
 end
--- a/app/controllers/imports_controller.rb
+++ b/app/controllers/imports_controller.rb
@@ -28,7 +28,7 @@ class ImportsController < ApplicationController
 
     respond_to do |format|
       if @import.save
-        @import.import_later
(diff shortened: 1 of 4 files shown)


app/controllers/imports_controller.rb
```diff
@@ -28,7 +28,7 @@ def create
 
     respond_to do |format|
       if @import.save
-        @import.import_later
+        @import.import_later(ActiveRecord::Type::Boolean.new.cast(params[:podcast_import][:import_metadata]))
```
radical-ube, line 31: passing in a boolean here determines whether or not `import_podcast!` will be called.

app/models/podcast_import.rb
```diff
@@ -114,10 +114,6 @@ def import_episodes!
     # Request the RSS feed again
     get_feed
 
-    # Update podcast attributes
-    create_or_update_podcast!
```
radical-ube, line 118: this method needed to be removed here since it defeats the purpose of the above-mentioned boolean for `import_podcast!`; it's already called in that method.

cavis: Huh, weird.  It does seem a bit redundant/out-of-place.  Good call!

app/models/podcast_import.rb
```diff
@@ -6,7 +6,7 @@ class PodcastImport < ApplicationRecord
 
   serialize :config, HashSerializer
 
-  attr_accessor :feed, :feed_raw_doc, :templates, :feed_source
+  attr_accessor :feed, :feed_raw_doc, :templates, :feed_source, :import_metadata
```
cavis, line 9: Curious - why do you need the attr_accessor here?  I don't see it being accessed anywhere.

app/controllers/imports_controller.rb
cavis: Minor nitpick - I might move that params parsing to another line / a private method, for readability.

cavis (approved):
A question and a minor nitpick, but looks good to me!

app/models/podcast_import.rb
radical-ube: maybe a habit from a different ticket. for some reason i thought it was necessary, but i suppose not!

cavis (approved):
👍 looks good to me!

PR #788, opened 2023-08-16
Extract the call to podcast publish to a job

svevang:
In the case where we have a transaction block calling `Podcast#publish!`, we can have dirty reads from other threads _outside_ of the transaction. This PR wraps the manipulation of the partitioned concurrency publishing queues in another async job invocation. This way, the database sessions outside of the authoring transactions can see the new state of the publishing queues.


cavis (approved):
👍

PR #789, opened 2023-08-17
Fix: no delivery files are expected if there is no delivery

svevang:
Fixes a case where an episode was uploaded to the apple system, but the wait for linked asset state timed out.

Upon manual retry, the episode was marked for re-upload by deleting the delivery/delivery_files on the episode (`PodcastContainer#reset_source_file_metadata`). But since the _creation_  of delivery files changed with 520f2c9 (now based on Episode#needs_delivery?), the guard in the Apple::PodcastDeliveryFile class was blowing up -- expecting a delivery to be inplace.

This PR removes the assumption that a Apple::PodcastDelivery has to be in place in the delivery file sync routine. If the file is in the container, there is no need to redeliver -- in this case the episode simply needed to be marked as published.

### DIFF
--- a/app/models/apple/podcast_delivery_file.rb
+++ b/app/models/apple/podcast_delivery_file.rb
@@ -98,17 +98,14 @@ module Apple
       episode_bridge_results
     end
 
+    def self.select_podcast_deliveries(episodes)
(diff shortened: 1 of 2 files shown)


cavis (approved):
👍 looks good to me!

PR #792, opened 2023-08-17
Fix z-index and other style bugs

brandonhundt:
- Fixes Z-index issue in the ID App Switcher dropdown menu on Episode pages by adjusting custom z-index values to [bootstrap z-index variables](https://getbootstrap.com/docs/5.2/layout/z-index/)

EPISODE Landing page:
- Adds a container div to the episodes index page, so that when teasers get long, the episode columns don't get wildly wide
- Audio status badge should now always be present in the episode-card

### DIFF
--- a/app/views/episodes/index.html.erb
+++ b/app/views/episodes/index.html.erb
@@ -34,6 +34,7 @@
     </div>
   </div>
 
+<div class="container">
   <%= turbo_frame_tag "episodes", target: "_top" do %>
     <div class="row mb-4">
       <div class="col-md-6">
@@ -61,4 +62,5 @@
       </div>
     </div>
   <% end %>
+</div>
 </div>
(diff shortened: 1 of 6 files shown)


app/views/episodes/index.html.erb
```diff
@@ -34,31 +34,33 @@
...
   </div>
 
-  <%= turbo_frame_tag "episodes", target: "_top" do %>
-    <div class="row mb-4">
-      <div class="col-md-6">
-        <h2 class="fw-bold"><%= t(".draft_or_scheduled") %></h2>
-        <% @scheduled_episodes.each do |episode| %>
-          <%= render episode %>
-        <% end %>
+  <div class="container">
```
brandonhundt, line 37: This adds a wrapper + indents, no changes made within the wrapper

cavis (approved):
👍

Issue #768, opened 2023-08-15
Style drag-n-drop uploaders


cavis:
Do some simple CSS styling when dragging files onto the app.  At least highlighting the droppable area.

PR #804, opened 2023-08-21
Style drag n drop

brandonhundt:
Closes: #768 

- This is a NO JS approach. It does most things except hover on the drop (which I think is okay.) If folks feel that is important we can add.

### DIFF
--- a/app/assets/stylesheets/shared/form.scss
+++ b/app/assets/stylesheets/shared/form.scss
@@ -112,6 +112,10 @@ label[required="true"] {
       border-color: $red;
     }
 
+    &.prx_dragndrop {
+      border: 2px dashed $gray-300;
+    }
+
     /*&:focus {
       box-shadow: none;
       border-color: transparent;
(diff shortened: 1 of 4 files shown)


app/assets/stylesheets/shared/form.scss
```diff
@@ -112,6 +112,10 @@ label[required="true"] {
       border-color: $red;
     }
 
+    &.prx_dragndrop {
+      border: 2px dashed $gray-300;
+    }
```
brandonhundt, line 117: Adds a custom class to apply to file fields - I had to do this to overwrite the form-control border

config/locales/en.yml
```diff
@@ -545,7 +545,7 @@ en:
         sample_rate: Frequency
         status: Status
       new:
-        hint: Upload a file
+        hint: Upload a file or drag it here
```
brandonhundt, line 548: Figure some language to indicate drag n drop is available might be enough in and of itself.

cavis (approved):
👍

PR #807, opened 2023-08-24
Fix/readonly access

### LINKED ISSUE #767: Readonly user testing/fixing
Try "all the things" as a readonly user.  Figure out what's wrong, document it here, then fix it.

(The file uploaders probably have some issues).

KNOWN ISSUES:

- [ ] Create new podcast link (the floating one on podcasts#index) should not show up if you can't create one.

radical-ube:
resolves #767 

attempts to address the `readonly` UI/UX. authorization seemed to be working for most things; a few extra authorizations were added, and for the most part the rest was hiding/disabling UI.

A simplified list of changes:

- podcasts#show: hide Create Episode and Plan Episodes buttons
- podcasts#edit (metadata): disable Subcategory field
- image cards (podcast metadata, episodes#show, feeds non-default): hide delete icon (trash can), hide new upload UI altogether
- podcasts#engagement: hide save button
- podcasts#planner: authorization for route, hide side tab nav UI
- episodes#index: hide Create Episode and Plan Episodes buttons
- episode media files: hide Save button, hide trash can, hide new upload UI altogether

it may seem like many lines have been changed, however this is mostly due to wrapping certain `div`s or partials with a policy conditional to hide them.


app/views/episode_media/media/_new.html.erb
```diff
@@ -1,93 +1,95 @@
...
-    <%# fake display field; real text area is opacity-0 on top of it %>
-    <% invalid_msg = upload_invalid_messages(media) %>
-    <% fake_cls = "form-control d-flex align-items-center prx_dragndrop" %>
-    <% fake_cls += " is-changed" if media.marked_for_destruction? %>
-    <% fake_cls += " is-invalid" if invalid_msg %>
-    <div class="<%= fake_cls %>">
-      <span class="material-icons text-primary">upload</span>
-      <div class="text-secondary mx-2 flex-grow-1 overflow-hidden text-truncate"><%= t(".hint") %></div>
-    </div>
+<% if policy(episode).update? %>
```
cavis, line 1: To minimize the diff - I might move this into `media/_media.html.erb`.  And also use the `policy(media).create?` (assuming that works... can't remember if all the associations get set on there ahead of time).

app/views/episode_media/media/_complete_audio.html.erb
```diff
@@ -25,8 +25,10 @@
 
   <label><%= episode_media_label(episode, media) %></label>
 
-  <%= link_to episode_media_path(media.episode, uploads_destroy_params(form)), class: "input-group-text prx-input-group-text" do %>
-    <span class="material-icons text-danger">delete</span>
+  <% if policy(episode).update? %>
```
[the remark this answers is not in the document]
cavis: Noting there are a few less-common delete buttons that also need to be wrapped for readonly users - 

In `media/_error.html.erb` and `media/_processing.html.erb`.

app/views/images/_complete.html.erb
```diff
@@ -8,8 +8,10 @@
 
     <%= help_text t("images.help.#{image.model_name.singular}") %>
 
-    <%= link_to delete_path, class: "prx-btn-help" do %>
-      <span class="material-icons">delete</span>
+    <% if policy(@podcast).update? %>
```
cavis, line 11: I think there are Image policies, for every flavor - so can probably just `policy(image).destroy?`.

And the error/processing partials also need this.

app/views/images/_new.html.erb
```diff
@@ -1,90 +1,92 @@
...
-
-    <%# delete buttons in header, after upload completes %>
-    <button type="button" class="prx-btn-help d-none" data-action="upload#cancelUpload" data-upload-target="success">
-      <span class="material-icons">delete</span>
-    </button>
-    <button type="button" class="prx-btn-help d-none" data-action="upload#cancelUpload" data-upload-target="error">
-      <span class="material-icons">delete</span>
-    </button>
-  </div>
+<% if policy(@podcast).update? %>
```
cavis, line 1: I might also move this up into `images/_form.html.erb` to minimize the diff.

app/views/podcasts/_form_main.html.erb
```diff
@@ -67,7 +67,7 @@
     <div class="form-floating">
       <%= form.fields_for :default_feed do |fields| %>
         <% opts = itunes_subcategory_options(podcast.itunes_category) %>
-        <%= fields.select :itunes_subcategory, opts, {include_blank: true}, disabled: opts.blank? %>
+        <%= fields.select :itunes_subcategory, opts, {include_blank: true}, disabled: (opts.blank? || !policy(podcast).update?) %>
```
cavis, line 70: Ah, good catch.  It had its own disable-the-field logic.

app/views/episodes/index.html.erb
```diff
@@ -5,6 +5,7 @@
     <h1 class="text-black fw-bold d-flex flex-fill my-4"><%= t(".episodes") %></h1>
     <div class="d-flex gap-2 align-items-center">
       <% if @podcast.present? %>
+        <% if policy(@podcast).update? %>
```
cavis, line 8: This could be `policy(@podcast.episodes.new).create?` ... in case we someday want to give some users edit permission but not create.

cavis (changes requested):
A couple minor image/media states that still have a delete button, but otherwise looking good.

Just in terms of future-proofing - I would recommend always checking the most specific policy we have.  Trying to delete a media file?  Check the `policy(media).destroy?`, instead of shortcutting to the `policy(podcast).update?`.  Just in case we someday change the Policy logic, or invent ID roles that are different than today.

cavis (approved):
👍 looks good to me!

PR #808, opened 2023-08-25
Add conditional on filter to default to base query

### LINKED ISSUE #798: Mismatched podcast total count
For some users (farski) - the podcast switcher dropdown total count doesn't match the podcast index total count.

> The global podcast selector for me says I have 306 podcasts, but the page that link takes me to says "Displaying podcasts 1 - 10 of 303 in total"

Try to replicate that, and figure out what's going on.

radical-ube:
resolves #798 

masqueraded as Farski and was able to reproduce the mismatched podcast bug. it was an issue of the search filter used on each; the switcher does not use a text filter if nothing is input, whereas `podcasts#index` did not have that conditional. as a result, `podcasts#index` seemed to be filtering some podcasts even with an empty string, resulting in the mismatched podcast numbers.


cavis (approved):
👍 nice catch!

PR #811, opened 2023-08-30
Fix/episode categories presets

### LINKED ISSUE #779: Episode categories adfree preset
Add some preset mini-badge-button-things to the Episode categories.  Similar to Publish.

We only had `adfree` in Publish.  But include `paidadfree` and `houseadfree` as well.

radical-ube:
resolves #779


cavis (changes requested):
One minor request on colors/toggles.  But otherwise this is working great.

app/views/episodes/_form_tags.html.erb
```diff
@@ -1,15 +1,21 @@
...
       <div class="form-floating input-group">
-        <%= form.tag_select :categories, episode.categories, {}, data: {allow_new: true, allow_clear: true} %>
+        <%= form.tag_select :categories, episode.categories, {}, data: {allow_new: true, allow_clear: true, category_target: "field"} %>
         <%= form.label :categories %>
         <%= field_help_text t(".help.categories") %>
       </div>
     </div>
+
+    <div class="card-footer" data-controller="tooltip">
+      <button type="button" class="btn btn-sm btn-light" data-bs-toggle="tooltip" data-bs-title="<%= t(".tooltip.adfree") %>" data-bs-placement="bottom" value="adfree" data-action="click->category#addCategory"><%= t(".label.adfree") %></button>
```
radical-ube, line 16: so, the `badge` class on a button element is a little weird in terms of the text vs background (ends up being light on light, very unreadable) so instead i just turned it into `btn-sm` to at least make them small buttons and for the css to layer properly.

cavis: Oh, I see.  That is a bit odd... you could toggle the `text-dark` class as well?  But even then, I can't see the difference between the badge color and the background on my monitor.  So this works!

app/javascript/controllers/category_controller.js
```diff
@@ -0,0 +1,20 @@
...
+import Tags from "bootstrap5-tags"
+
+export default class extends Controller {
+  static targets = ["field"]
+
+  addCategory(event) {
+    const vals = Tags.getInstance(this.fieldTarget).getSelectedValues()
+
+    if (!vals.includes(event.target.value)) {
+      Tags.getInstance(this.fieldTarget).addItem(event.target.value)
```
cavis, line 11: Noticed one small bug - if the tags are initially empty and you add one, the field label doesn't float successfully.

I think you either need to fire a `changed` event on it... or remove the `form-control-blank` class?

cavis (changes requested):
Okay okay, 2 more minor things I didn't catch the first time.

app/views/episodes/_form_tags.html.erb
```diff
@@ -1,15 +1,21 @@
...
+      <button type="button" class="btn btn-sm btn-light" data-bs-toggle="tooltip" data-bs-title="<%= t(".tooltip.houseadfree") %>" data-bs-placement="bottom" value="houseadfree" data-action="click->category#addCategory"><%= t(".label.houseadfree") %></button>
+      <button type="button" class="btn btn-sm btn-light" data-bs-toggle="tooltip" data-bs-title="<%= t(".tooltip.paidadfree") %>" data-bs-placement="bottom" value="paidadfree" data-action="click->category#addCategory"><%= t(".label.paidadfree") %></button>
```
cavis, line 18: Oh, one more thing ... these need to check in ruby if the episode has the tags or not.  So on initial page render ... they should be blue, if `adfree` etc is already saved to the episode.

cavis (approved):
Needs some linting, but otherwise :+5: looks good to me!

Issue #775, opened 2023-08-15
Prevent deleting uncut audio for published episodes


cavis:
If you upload a single segment, then publish... you're allowed to delete that single segment without replacing it.

You should be allow to replace + re-segment it.  But not delete it entirely.

Or come up with some other better UI to indicate that they're replacing, not deleting.

PR #816, opened 2023-09-05
Media file bugfixes

### LINKED ISSUE #770: Fix episode media tab with nil segment_count
Sometimes episodes may come in (via the API, or just real old data) with a nil segment count.  Need to handle this in the UI.

cavis:
Fixes #770 and #775.  And some other bugs I found along the way (did a lot of QA switching media formats, resaving, publishing, etc).

In alphabetical order of files changed:

- [x] Incomplete media for published episodes now has a red `Incomplete` badge, instead of the orange warning color.  To really indicate this needs attention.
- [x] Always set a non-nil `segment_count` for episodes (that may have come in via the API), so the UI doesn't break.
- [x] When changing from "upload single file" to "upload multiple segments" ... keep the existing segments around if they exist.
- [x] Adds `media_versions` to episodes, grouping together completed media files that should serve together.  This will eventually replace `media_resources.replaced_at` - but I haven't removed that field yet.  Just in case!
- [x] Fix a bug where you couldn't replace+resegment a single file.  You now can (though you'll get the angry red `Incomplete` badge until you go re-mark it).
- [x] Fixed some other misc changed/invalid indicators, and made error messages more friendly when your files aren't ready.


app/models/concerns/episode_media.rb
```diff
@@ -3,19 +3,32 @@
...
+    # backfill media_versions for newly completed media
+    if media_ready? && latest_ids != media_ids
+      new_version = media_versions.build
+      media.each { |m| new_version.media_version_resources.build(media_resource: m) }
+      new_version.save!
+      new_version
+    else
+      latest_version
+    end
+  end
```
cavis, line 20: Okay, so creating media versions...

I landed on _not_ init-ing these on save.  They're only created once all media files complete processing, they're the correct mime type, and they match the segment count.  This is a bit more SQL to run, but it seemed safest to fill them in as we publish the version of the media out.

The complete flow is:

1. You upload/save `episode.contents` (or we create them by segmenting the `episode.uncut`).
2. Async processing kicks off - eventually all the files complete.
3. If/when the publish jobs (apple or RSS) run, they eventually generate the eps in the feed, and ask [ep.is_include_in_feed?](https://github.com/PRX/feeder.prx.org/blob/main/app/models/feed.rb#L125)
4. If the episode purposefully has no media, it's included in the feed.  And there's no media_version to create.
5. If it has media, we call `complete_media?` (below) - which in turn calls this method.
6. This method checks if we need to create a new version or not.

And that's it.  Fairly simple?  Maybe a bit magical how we get here?  But I've thought/tested/QA'd through a bunch of edge cases, where you're swapping files before we publish the prior version.  And this holds up.  Since we're always (soft-) deleting the old files when we replace them.  And all of publishing is gated on `complete_media?`.

Anyways, any thoughts on that?  Ways to make it clearer what's happening?

app/models/concerns/episode_media.rb
```diff
@@ -3,19 +3,32 @@
...
+
+  def media_version_id
+    cut_media_version!&.id
```
cavis, line 23: Unused yet - but I figured this would be the place to track which media version we've sent to Apple.  Add a column to one of the apple tables, to track what the version was the last time we ran.

svevang: Do we need this if we're cutting a new version?

1) If the set of media files have new version, we mark the episode as ready for re-upload
2) A new version means we call `Episode#publish!`
3) The the delegated delivery step in publishing picks up the new enclosure url and re-uploads.

app/models/concerns/episode_media.rb
```diff
@@ -3,19 +3,32 @@
 module EpisodeMedia
   extend ActiveSupport::Concern
 
+  def cut_media_version!
+    latest_version = media_versions.first
+    latest_media = latest_version&.media_resources || []
+    latest_ids = latest_media.map(&:id)
+
+    # backfill media_versions for newly completed media
+    if media_ready? && latest_ids != media_ids
```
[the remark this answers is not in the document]
cavis: Yes - both the `media_version.resources` and `episode.contents` are sorted by position ASC.  So these IDs are directly comparable.

Initially I had a `.sort` on these - but that's not _quite_ right.  If we someday allowed re-ordering existing media segments (for some reason).

svevang (approved):

PR #821, opened 2023-09-11
Add div wrapping to metadata and episode description fields

### LINKED ISSUE #802: Changed-field indicator problems
Some minor issues remaining with highlighting changed fields in orange, and displaying the "discard" button.

- [ ] WYSIWYG don't have the is-changed style
- [ ] For some fileds (e.g., Episode Type) if you change the value, and then change it back to what it was, the field isn't marked as dirty. For other (e.g., Subtitle) the field remains marked even if the value returns to the existing value
    - This may be a `nil` vs `""` issue.  Because I'm not seeing those.

radical-ube:
resolves #802 

the `div` wrappers solve the `is-changed` part of the bug. and it seems that in staging the other part of the bug no longer shows up, so it seems resolved for now.

### DIFF
--- a/app/views/episodes/_form_main.html.erb
+++ b/app/views/episodes/_form_main.html.erb
@@ -29,7 +29,9 @@
       <%= help_text t(".help.description") %>
     </div>
     <div class="card-body">
-      <%= form.trix_editor :description %>
+      <div class="form-floating">
+        <%= form.trix_editor :description %>
+      </div>
     </div>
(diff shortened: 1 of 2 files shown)


cavis (approved):
👍 easy enough!  And I'm not seeing the other bug in the ticket either.  So I suspect it was a `nil` vs `""` thing (which I think I fixed in the formbuilder at some point).

PR #822, opened 2023-09-11
Fixup episode feed-ready logic

cavis:
Fixes 2 recent prod issues:

1. An episode publishing to the feed before all of its files had processed (`episode.include_in_feed?` was true ... because `complete_media?` was also true)
2. CMS announced updates to a published episode briefly resulted in `complete_media?` returning false (it deleted the old media files without setting `replaced_at`).  Dovetail 500'd.

### DIFF
--- a/app/models/concerns/episode_media.rb
+++ b/app/models/concerns/episode_media.rb
@@ -7,8 +7,11 @@ module EpisodeMedia
     contents.complete_or_replaced.group_by(&:position).values.map(&:first)
   end
 
+  # TODO: not ideal, but we need to ensure this stays true forever, after the
+  # first time the episode is written to any feed. for now, just assume that
+  # takes about an hour.
   def complete_media?
-    if published?
+    if published? && published_at < 1.hour.ago
       complete_media.any?
     else
       media_ready?
--- a/app/models/episode_story_handler.rb
+++ b/app/models/episode_story_handler.rb
@@ -80,8 +80,12 @@ class EpisodeStoryHandler
       []
     end
 
-    episode.media = audio.sort_by(&:position).map do |a|
-      a.links["prx:storage"].href
+    # possible race condition i couldn't quite track down, when replacing files on a
+    # published episode - so let's just not allow deleting it for now
+    unless episode.published? && audio.empty?
+      episode.media = audio.sort_by(&:position).map do |a|
+        a.links["prx:storage"].href
+      end
     end
   end
 
--- a/app/models/episode_story_handler.rb
+++ b/app/models/episode_story_handler.rb
@@ -105,22 +105,6 @@ class EpisodeStoryHandler
     end
   end
 
-  def build_content(audio)
-    update_content(Content.new, audio)
-  end
-
-  def update_content(content, audio)
-    content.tap do |c|
-      c.position = audio.attributes["position"]
-      c.href = audio.links["prx:storage"].href
-      c.mime_type = audio.links["prx:storage"].type || audio.attributes["content_type"]
-      c.file_size = audio.attributes["size"]
-      c.duration = audio.attributes["duration"]
-      c.bit_rate = audio.attributes["bit_rate"]
-      c.sample_rate = audio.attributes["frequency"] * 1000 if audio.attributes["frequency"]
-    end
-  end
-
   def get_story(account = nil)
     return nil unless episode.prx_uri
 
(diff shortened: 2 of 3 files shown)


app/models/concerns/episode_media.rb
```diff
@@ -7,8 +7,11 @@ def complete_media
     contents.complete_or_replaced.group_by(&:position).values.map(&:first)
   end
 
+  # TODO: not ideal, but we need to ensure this stays true forever, after the
+  # first time the episode is written to any feed. for now, just assume that
+  # takes about an hour.
```
cavis, line 12: Long term plan here: add a `media_resource_versions` table to track episode `Content` that can be grouped together.  (And maybe only create these records when all contents in that version are "complete"?).

Basically, we want to track any past "version" of the audio we served.  Because the user may change the episode segment_count / medium / re-upload-files / re-cut-files, making the current files invalid.  And we need something to fall back on.

app/models/concerns/episode_media.rb
```diff
@@ -7,8 +7,11 @@ def complete_media
...
   def complete_media?
-    if published?
+    if published? && published_at < 1.hour.ago
```
svevang, line 14: So this is to guard against media changing "quickly"?

Is there some logging we can do to track the races here? Like "here is the old version, and here is the new version"?

Maybe we can sample frequently enough that we can bisect the internal state of the race.

app/models/concerns/episode_media.rb
svevang: That makes sense. Versioning the media with each mutation would be a great way to tackle these race conditions and start picking apart the details of the concurrency model.  Though mostly the details are handled already!

svevang: How will adding versions subsequent to this patch fit with the changes here?  Where can I find a sketch of what that work consists of?

app/models/episode_story_handler.rb
```diff
@@ -80,8 +80,12 @@ def update_audio
...
     end
 
-    episode.media = audio.sort_by(&:position).map do |a|
-      a.links["prx:storage"].href
+    # possible race condition i couldn't quite track down, when replacing files on a
+    # published episode - so let's just not allow deleting it for now
+    unless episode.published? && audio.empty?
+      episode.media = audio.sort_by(&:position).map do |a|
+        a.links["prx:storage"].href
+      end
```
svevang, line 88: Would this allow setting an episode's media via publish UI only one time?  Or another way, the episode should be unpublished first, then update the audio, and then publish the episode again?

app/models/concerns/episode_media.rb
cavis: This guards against a brand new episode publishing with only a subset of its contents completed.

- For the first hour, we check that `episode.segment_count` matches `episode.contents.status_complete.count`.  (We need a completed file for every position).
- After that, allow what files we serve to deviate (in case the user changes the segment_count or something).  Preserving existing functionality for now.

This is intended to a temporary patch - #816 introduces the actual "episode media versions" tracking.  (But I think that whole branch needs more QA).

app/models/concerns/episode_media.rb
cavis: Started working on that in [#816](https://github.com/PRX/feeder.prx.org/pull/816/files#diff-bda765f89a866286ad1f1b6a4cf2ed1eed4bb2b6859d7e354da995bfaf139f32R11).  Basically, whenever we have a set of completed files for every position in the segment_count ... we cut a new version.

(I'm still mulling over that branch ... was going for minimal changes on the first pass)

app/models/episode_story_handler.rb
cavis: This only blocks deleting all the files of a published episode.  (I think because of how Publish serializes API calls to CMS, we were briefly announcing an empty `prx:storage` ... causing Feeder to remove the files entirely, instead of replacing them as the user intended).

svevang (approved):
It looks like this works around some of the media update races coming from CMS+publish. #816 is the direction for a deeper solution.

If [#816](https://github.com/PRX/feeder.prx.org/pull/816/files#diff-bda765f89a866286ad1f1b6a4cf2ed1eed4bb2b6859d7e354da995bfaf139f32R11) is where we'll eventually land it makes sense to move forward on this patch!

PR #823, opened 2023-09-11
820: Embed player forms

### LINKED ISSUE #820: Update player forms to use play preview route

rpeterman-gp:
Closes #820 

- Reworks embed player forms to work with Play's new `/preview` route.
- Configuration changes are now sent to preview embed iframe via postMessage events, allowing the Play app to render changes without reloading iframe.
- Updated embed URL and HTML, along with height and styles, are sent from the preview iframe via postMessages for the feeder form to display and update preview iframe wrapper markup.
- Updates dyanamic-form controller to update select field help text based on selected values without needing to reload the form state (no phantom clicking a hidden button.)
- Simplifies embed style selection to "Standard" and "Card". Player embed markup is always responsive, but the added _Player Width_ option allows a max width to be set.
- Adds theme selection
- Adds accent color selection
- Reworks podcast embed player playlist options too not require an extra select field for "All" or "Number". _Number of Episodes_ field has been reworked to treat empty value as all, and 1 episode as latest episode only.
- Adds some skeleton text around preview to help demonstrate in-use behavior.
- Adds resizable wrapper around preview so user can easily test how their changes will responsively behave.
- Updates labels and help text to reflect changes, and for more clarity.

## To Review

- Get your Play local dev updated and running with PR PRX/Play-Next.js#123
- Make sure puma dev is configured to host `play.prx.test` from port 4300
- Checkout this branch in your feeder dev
- Add `PLAY_HOST=play.prx.test` to you `.env`
- Start server
- Go to the Embed Player of your favorite podcast
- Ensure embed player loads with default setting
- Ensure changing Player Type updates the preview, and the URL and HTML in copy fields
- Ensure setting a Player Width updates the preview, and the URL and HTML in copy fields
- Ensure changing Player Theme updates the preview, and the URL and HTML in copy fields
- Ensure changing the Accent Color updates the preview, and the URL and HTML in copy fields
- Ensure incrementing the Number Of Episodes updates the preview, and the URL and HTML in copy fields
  - Empty should show all episodes
  - 0 or 1 should not show a playlist
  - 2-5 should show playlist without scrollbars
  - > 5 should show playlist with scrollbars
  - Resizing preview area should result in no more than 5 and a half episode rows being shown
- Ensure changing the Session updates the preview playlist, and the URL and HTML in copy fields
- Ensure changing Category updates the preview playlist, and the URL and HTML in copy fields
- Ensure copy fields for Embed Player Link and Embed Player IFrame copy field content to clipboard (Should no ask for permission or not show button if user settings deny clipboard access.)
- Ensure _Restore Default Options_ button resets form
- Go to the Embed Player for a published episode
- Ensure preview uses RSS feed to render player
- Go to the Embed Player for an unpublished episode
- Ensure preview uses override parameters to render player

### DIFF
--- a/app/helpers/embed_player_helper.rb
+++ b/app/helpers/embed_player_helper.rb
@@ -3,6 +3,7 @@ module EmbedPlayerHelper
 
   EMBED_PLAYER_LANDING_PATH = "/listen"
   EMBED_PLAYER_PATH = "/e"
+  EMBED_PLAYER_PREVIEW_PATH = "/preview"
   EMBED_PLAYER_FEED = "uf"
   EMBED_PLAYER_GUID = "ge"
   EMBED_PLAYER_CARD = "ca"
@@ -48,48 +49,73 @@ module EmbedPlayerHelper
     params = {}
     params[EMBED_PLAYER_FEED] = podcast.published_url
 
-    params[EMBED_PLAYER_PLAYLIST] = (options[:all_episodes] == "all") ? options[:all_episodes] : options[:episode_number]
-    params[EMBED_PLAYER_SEASON] = options[:season]
-    params[EMBED_PLAYER_CATEGORY] = options[:category]
+    if options[:all_episodes] === "all"
+      params[EMBED_PLAYER_PLAYLIST] = "all"
+    end
 
-    # TODO: the height styling doesn't really work here in the way Play specifies how it needs to be for a playlist, so leaving this as a TODO for now.
-    # if options[:embed_player_type] == "card" || options[:embed_player_type] == "fixed_card"
-    #   params[EMBED_PLAYER_CARD] = "1"
-    # end
+    if options[:all_episodes] == "number" && options[:episode_number].to_i > 1
+      params[EMBED_PLAYER_PLAYLIST] = options[:episode_number]
+    end
+
+    if !options[:season].to_s.strip.empty? && options[:season].to_i > 0
+      params[EMBED_PLAYER_SEASON] = options[:season]
+    end
+
+    if !options[:category].to_s.strip.empty?
+      params[EMBED_PLAYER_CATEGORY] = options[:category]
+    end
+
+    if options[:embed_player_type] == "card" || options[:embed_player_type] == "fixed_card"
+      params[EMBED_PLAYER_CARD] = "1"
+    end
 
     embed_params(params)
   end
 
   def embed_player_episode_iframe(ep, type = nil, preview = false)
     src = embed_player_episode_url(ep, type, preview)
-    allow = "monetization"
+    allow = "monetization origin"
+    data = { :embed_preview_target => 'embedIframe' }
 
     if type == "card"
-      # TODO: this is NOW working, but I'm not sure how helpful this is to a producer that wishes to embed it.
-      tag.iframe src: src, allow: allow, width: "100%", height: "700", style: "--aspect-ratio: 2/3; width: 100%;"
+      # Responsive cards require a wrapper div around iframe. The wrapper uses padding to determine the size of the area the iframe should fill.
+      tag.div style: "position: relative; height: 0; width: 100%; padding-top: calc(100% + 200px);" do
+        tag.iframe src: src, allow: allow, width: "100%", height: "100%", style: "position: absolute; inset: 0;", data: data
+      end
     elsif type == "fixed_card"
-      tag.iframe src: src, allow: allow, width: "500", height: "700"
+      # Fixxed card heights should always be 200px larger than the width.
+      tag.iframe src: src, allow: allow, width: "500", height: "700", data: data
     else
-      tag.iframe src: src, allow: allow, width: "100%", height: "200"
+      tag.iframe src: src, allow: allow, width: "100%", height: "200", data: data
     end
   end
 
-  def embed_player_podcast_iframe(podcast, options, preview = false)
-    src = embed_player_podcast_url(podcast, options, preview)
+  def embed_player_iframe(options)
+    src = ''
     allow = "monetization"
+    iframeHeight = '200'
+    iframeStyle = ''
+    wrapperStyle = 'line-height: 0;'
 
     if options[:embed_player_type] == "card"
-      # TODO: this is NOW working, but I'm not sure how helpful this is to a producer that wishes to embed it.
-      tag.iframe src: src, allow: allow, width: "100%", height: "700", style: "--aspect-ratio: 2/3; width: 100%;"
-    elsif options[:embed_player_type] == "fixed_card"
-      tag.iframe src: src, allow: allow, width: "500", height: "700"
-    else
-      tag.iframe src: src, allow: allow, width: "100%", height: "200"
+      wrapperStyle = "position: relative; height: 0; width: 100%; padding-top: calc(100% + #{iframeHeight}px); line-height: 0;"
+      iframeHeight = '100%'
+      iframeStyle = "position: absolute; inset: 0;"
+    end
+
+
+    tag.div style: wrapperStyle, data: { :embed_preview_target => 'embedIframeWrapper' } do
+      tag.iframe class: 'bg-light', src: src, allow: allow, width: "100%", height: iframeHeight, style: iframeStyle, data: { :embed_preview_target => 'embedIframe' }
     end
   end
 
   def embed_player_type_options(selected)
-    opts = %w[standard card fixed_card].map { |v| [t("helpers.label.episode.embed_player_types.#{v}"), v] }
+    opts = %w[standard card].map { |v| [t("helpers.label.episode.embed_player_types.#{v}"), v] }
+    options_for_select(opts, selected)
+  end
+
+  def embed_player_theme_options(selected)
+    opts = %w[dark light auto].map { |v| [t("helpers.label.episode.embed_player_themes.#{v}"), v] }
     options_for_select(opts, selected)
   end
 
@@ -106,7 +132,7 @@ module EmbedPlayerHelper
   private
 
   def embed_params(params)
-    "#{play_root}#{EMBED_PLAYER_PATH}?#{params.to_query}"
+    "https://#{ENV["PLAY_HOST"]}#{EMBED_PLAYER_PREVIEW_PATH}?#{params.to_query}"
   end
 
   def enclosure_with_token(ep)
--- a/app/helpers/embed_player_helper.rb
+++ b/app/helpers/embed_player_helper.rb
@@ -24,10 +24,10 @@ module EmbedPlayerHelper
     "#{play_root}#{EMBED_PLAYER_LANDING_PATH}?#{params.to_query}"
   end
 
-  def embed_player_episode_url(ep, type = nil, preview = false)
+  def embed_player_episode_url(ep, options = nil, preview = false)
     params = {}
 
-    if preview && !ep.published?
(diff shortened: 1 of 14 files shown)


rpeterman-gp:
Probably will want to merge and deploy the Play PR **BEFORE** merging this PR.

cavis:
Just FYI ... I'm still waiting on the Play changes to deploy before reviewing this one.

app/views/episode_player/_form.html.erb
```diff
@@ -1,64 +1,123 @@
-<%= turbo_frame_tag "embed-player", data: {controller: "morphdom"} do %>
+<div class="row my-4 mx-2" data-controller="embed-preview"
```
cavis, line 1: Bit aesthetic ... but I might break out some of the components into sub-partials?  `_form_player.html.erb` `_form_copy.html.erb` `_form_sidebar.html.erb` etc (naming is hard).

Just because this file is getting a bit lengthy + indented.

app/helpers/embed_player_helper.rb
```diff
@@ -37,59 +38,65 @@ def embed_player_episode_url(ep, type = nil, preview = false)
...
-    params[EMBED_PLAYER_CATEGORY] = options[:category]
+    if options[:all_episodes] === "all"
+      params[EMBED_PLAYER_PLAYLIST] = "all"
+    end
 
-    # TODO: the height styling doesn't really work here in the way Play specifies how it needs to be for a playlist, so leaving this as a TODO for now.
-    # if options[:embed_player_type] == "card" || options[:embed_player_type] == "fixed_card"
-    #   params[EMBED_PLAYER_CARD] = "1"
-    # end
+    if options[:all_episodes] == "number" && options[:episode_number].to_i > 1
```
cavis, line 56: `> 1` not `>=` 1?

app/helpers/embed_player_helper.rb
```diff
@@ -37,59 +38,65 @@ def embed_player_episode_url(ep, type = nil, preview = false)
...
-      tag.iframe src: src, allow: allow, width: "500", height: "700"
-    else
-      tag.iframe src: src, allow: allow, width: "100%", height: "200"
+      wrapper_styles = "position: relative; height: 0; width: 100%; padding-top: calc(100% + #{iframe_height}px); line-height: 0;"
+      iframe_height = "100%"
+      iframe_styles = "position: absolute; inset: 0;"
+    end
+
+    tag.div style: wrapper_styles, data: {embed_preview_target: "embedIframeWrapper"} do
+      tag.iframe class: "bg-light", src: src, allow: allow, width: "100%", height: iframe_height, style: iframe_styles, data: {embed_preview_target: "embedIframe"}
```
cavis, line 89: So I'm curious - why the big diff in this method?

It looks like you've structurally moved the generation of the iframe code to the frontend / stimulus.  Rather than generating the changes in ruby and reloading via turbo.  So I'm not entirely sure when most of these conditional branches in this method would get called.

app/javascript/controllers/clipboard_controller.js
```diff
@@ -3,7 +3,18 @@ import { Controller } from "@hotwired/stimulus"
 export default class extends Controller {
   static values = { copy: String, tooltip: String }
 
+  originalStyleDisplay
```
cavis, line 6: What is this?  I guess I don't usually see stimulus controllers defining instance variables outside of the `connect()`.

app/javascript/controllers/clipboard_controller.js
```diff
@@ -3,7 +3,18 @@ import { Controller } from "@hotwired/stimulus"
...
+
   connect() {
+    this.originalStyleDisplay = this.element.style.display
+
+    navigator.permissions.query({ name: "clipboard-write" }).then((result) => {
+      this.updatePermissionState(result.state)
+      result.addEventListener("change", () => {
```
cavis, line 13: So this is just checking if we have permission to write to the clipboard, and hiding the copy button entirely if we don't?  And how would it change, while the user is sitting on this page?

Does this event listener need to be disconnected at all?

app/javascript/controllers/dynamic_form_controller.js
```diff
@@ -1,7 +1,17 @@
...
 
 export default class extends Controller {
-  static targets = ["link"]
+  static targets = ["link", "help"]
+
+  static values = {
+    help: Object,
+  }
+
+  updateHelp(event) {
```
cavis, line 10: This feels like it belongs in a different controller.  The small/single purpose of this one is to watch for a field to change and click an anchor.  Simulating a non-js flow where the user changes a select or something, then clicks "update form".

app/javascript/controllers/embed_preview_controller.js
```diff
@@ -0,0 +1,140 @@
...
+      "message",
+      (e) => {
+        that.handlePostMessage(e)
+      },
+      false
+    )
+  }
+
+  handlePostMessage(e) {
+    if (!/^https?:\/\/play(?:\.staging)?\.prx\.(?:org|tech|test)$/.test(e.origin)) return
```
cavis, line 38: I'm not sure I follow what this code is doing - can you clarify?  (Either here, or in the code itself)

app/views/podcast_player/_form.html.erb
```diff
@@ -1,94 +1,149 @@
-<%= turbo_frame_tag "embed-player", data: {controller: "morphdom"} do %>
-  <div class="col-12 mb-4">
-    <div class="form-floating input-group" data-controller="dynamic-form">
-      <%= select_tag :embed_player_type, embed_player_type_options(player_options[:embed_player_type]), class: "form-control", data: {action: "dynamic-form#change"} %>
-      <%= label_tag :embed_player_type, t("helpers.label.podcast_player.embed_player_type") %>
-      <a hidden href="<%= request.fullpath %>" data-dynamic-form-target="link"></a>
+<div class="row my-4 mx-2" data-controller="embed-preview"
```
cavis, line 1: I'd probably also split this out into partials.

app/helpers/embed_player_helper.rb
```diff
@@ -3,6 +3,7 @@ module EmbedPlayerHelper
 
   EMBED_PLAYER_LANDING_PATH = "/listen"
   EMBED_PLAYER_PATH = "/e"
+  EMBED_PLAYER_PREVIEW_PATH = "/preview"
```
cavis, line 6: Not exactly sure where this is happening - but this doesn't appear to fix the "user copies a url with their JWT in it" bug.  In fact, even the not-in-iframe fields at the bottom of the page have a JWT in them.

We want to prevent user's from every copying/using a url that looks different from after the episode publishes.

If I go to an unpublished episode, i see it loading `/preview` as expected.  But:

- In the iframe, "Download Audio" has a JWT.  This isn't great - I'd rather hide or give them a url that 404s with dovetail until the episode publishes.
- In the iframe "Copy Embed Code" has a JWT.  This is a problem.
- Outside the iframe, the "Embeddable Player Link" field has a JWT.  It should not, and neither should the copy link. And we should hide the "open in new tab" link until the episode has published.
- Outside the iframe, the "Embeddable Player Iframe" field has the JWT (same problems as above)
- Outside the iframe, the "Media Link" seems correct (you get a 404 from DT until it publishes).

cavis (changes requested):
Couple bugs found, and some questions.  Functionally, a big improvement though.

Generally, this seems to change the embed pages from turbo-reloaded rails views to a stimulus js mini-app.  Which given the complexity/options of the new UI, I'm okay with.  But just stating it is a structural change (and I'm not entirely sure what the ruby helper methods are doing now?)

Couple suggestions for refactoring views.  And for stimulus - I'd prefer keeping controllers very small/generic and single-purpose when possible.  Until they become more of a "mini app" flavor of controller.

app/helpers/embed_player_helper.rb
rpeterman-gp: 1 episode a playlist doesn't make. Playlist of 1 episode is just the player.

cavis: Oh, I see.  That makes sense.

app/helpers/embed_player_helper.rb
rpeterman-gp: Yes. The embed URL and iframe HTML is generated by play now. This logic is basically used once to initialize the iframe and wrapper based on configured defaults.

app/javascript/controllers/dynamic_form_controller.js
rpeterman-gp: Ok. I can move it to its own controller. I interpreted it as a dynamic form feature, which is why I put it here.

app/helpers/embed_player_helper.rb
rpeterman-gp: Since Play is now sending back the embed URL and iframe markup, we'll have to add some logic to remove the JWT from the audio URL when generating those, and only use it for audio playback. Unpublished audio shouldn't be getting the share buttons to copy ULR or embed code in the iframe. I'll check to see where the disconnect is that is still showing those buttons.

app/javascript/controllers/clipboard_controller.js
rpeterman-gp: Needed a class property that could be referenced and updated in multiple class methods overtime.

app/helpers/embed_player_helper.rb
rpeterman-gp: Kind of wish the Play PR hadn't been merged till this PR was reviewed and approved. The two PR's were tightly coupled. :(

app/javascript/controllers/clipboard_controller.js
rpeterman-gp: Actually, I do't think this needs cleaned up. This even handler is being attached to the permission query result, which happens once when navigating to that screen. When the user returns to this screen, a new query is done and the event is attached to the new result. Shouldn't have to worry about it firing multiple times.

app/views/episode_player/_form.html.erb
rpeterman-gp: Broke the elements shared by episode and podcast player forms into sub-partials. Podcast player form update to use episode player forms sub-partials.

app/views/podcast_player/_form.html.erb
rpeterman-gp: Updated to use episode player forms sub-partials.

app/javascript/controllers/dynamic_form_controller.js
```diff
@@ -1,8 +1,6 @@
 import { Controller } from "@hotwired/stimulus"
 
 export default class extends Controller {
-  static targets = ["link"]
```
cavis, line 4: I think maybe you didn't mean to remove this?  We still need it.

app/views/episode_player/_form.html.erb
cavis: +1 nice!

app/views/podcast_player/_form.html.erb
cavis: +1

cavis (approved):
Looks and works good!  Just lost 1 param on the `dynamic_form_controller` - add that back and this is good to go!

app/javascript/controllers/dynamic_form_controller.js
rpeterman-gp: Yep. My mistake. Fixed.

PR #829, opened 2023-09-14
Publicfeeds for new feeds

cavis:
Fixes PRX/publish.prx.org#771.  And a couple other bugfixes along the way.

You'll need to add `PUBLIC_FEEDS_PREFIX=https://publicfeeds.net/f` to your `.env` to see this in action.

(And it won't really work locally, since we have no publicfeeds url pointing to the `f.development.prxu.org` domain).


app/models/feed.rb
```diff
@@ -75,6 +77,16 @@ def sanitize_text
     self.title = sanitize_text_only(title) if title_changed?
   end
 
+  def set_public_feeds_url
+    if public? && url.blank? && ENV["PUBLIC_FEEDS_PREFIX"].present?
+      self.url = "#{ENV["PUBLIC_FEEDS_PREFIX"]}/#{podcast.path}/#{published_path}"
+    end
+  end
```
cavis, line 84: Note this is only set if the feed is public _and_ you left the `Public Feed URL` field blank.

And on deploy, this won't do anything until we set that ENV.

app/models/feed.rb
```diff
@@ -75,6 +77,16 @@ def sanitize_text
...
+
+  def remove_url
+    self.url = nil
+  end
```
cavis, line 88: Fixes a bug where changing a Feed from public -> private would retain the `url`.  Which caused a bad `atom:link` in the RSS.  (The `atom:link` needs to be that `p.prxu.org` for things to work).

app/models/feed.rb
```diff
@@ -93,19 +105,27 @@ def default_runtime_settings?
...
-      elsif include_token.nil?
-        "#{private_path}{?auth}"
-      else
-        private_path
-      end
+      published_private_url(include_token)
+    else
+      published_public_url
+    end
+  end
```
cavis, line 112: Abstacted the logic into 2 separate methods - so in the view, I can render them both and `d-none` one of them.

This allows us to show the correct `Private Feed URL` as you check/uncheck the "Private" checkbox.

app/views/feeds/_form_distribution.html.erb
```diff
@@ -31,13 +31,23 @@
         </div>
       </div>
 
-      <div class="col-12 mb-4">
-        <div class="form-floating input-group">
-          <%= form.text_field :published_url, value: feed.published_url(false), disabled: true %>
-          <%= form.label :published_url %>
-          <%= field_link feed.published_url(true) %>
+      <% if feed.persisted? %>
```
cavis, line 34: It doesn't make sense to show these until you create the Feed.  Since the path requires a `feed.slug`.

env-example
```diff
@@ -49,6 +49,7 @@ FEEDER_CDN_PRIVATE_HOST=p.staging.prxu.org
 ID_HOST=id.staging.prx.tech
 PLAY_HOST=play.staging.prx.tech
 PRX_HOST=beta.staging.prx.tech
+PUBLIC_FEEDS_PREFIX=https://publicfeeds.net/f
```
cavis, line 52: Opted to make this a "prefix" rather than "host".  Since it also needs that `/f` path.  But I could be convinced otherwise.

farski: Okay with prefix (though in other places seems like we may use the term _base_ for this sort of thing), but may consider `PUBLIC_FEEDS_URL_PREFIX` to make it less ambiguous

cavis: I was modeling on the feeder `feed.enclosure_prefix` column - which we fill in with this like `http://www.podtrac.com/pts/redirect.mp3`.

Yeah, will add `_URL` to that.

farski (approved):
Didn't try this locally, but looks fine

PR #831, opened 2023-09-14
830: Audio Segmenter Ad Breaks Change

### LINKED ISSUE #830: Audio Segmenter handle changes to number of ad breaks

rpeterman-gp:
Closes #830 

- rename segmentCount value to adBreaks
- add state management to handle changes to adBreaks value
- ad breakpoint start times can now be 0
- add break point fields should now indicate being changed
- update mark ad breaks card title
- updated to playhead buttons should function after ad breaks change


app/javascript/controllers/audio_breakpoint_controller.js
```diff
@@ -30,10 +31,29 @@ export default class extends Controller {
   }
 
   startTimeValueChanged() {
+    if (this.hasInitialMarkerValue) {
+      const isChanged = this.startTimeValue !== this.initialMarkerValue.startTime
+
+      this.startTimeTarget.parentNode.classList.toggle("is-changed", isChanged)
```
cavis, line 37: Hrmm... for me, I'm getting all orange highlighted fields all the time.  Even after clicking save or refreshing the page.

app/javascript/controllers/audio_breakpoints_controller.js
```diff
@@ -6,39 +6,61 @@ export default class extends Controller {
...
     this.initMarkers()
   }
 
   /**
    * Initial render of ad markers.
    */
   initMarkers() {
-    this.breakpointMarkers = [...Array(this.segmentCountValue - 1).keys()]
-      .map((key) => this.markersValue?.[key] || [])
+    if (!this.hasMarkersValue || !this.hasAdBreaksValue) return
```
cavis, line 36: May want to also do some light validation on this?  I'm noting I can (by typing, not clicking up/down) input `-3` etc.  And if i just add a few zeroes to it, the UI lags quite awhile trying to render 100s of these.

Or alternatively, pass in a value that is `ad-breaks-valid-value="<%= episode.validate(:ad_breaks) %>"`.

cavis (changes requested):
Okay, the fix for ad-breaks changing works as advertised.

Logically, you can't have midrolls at 0 seconds.  But it sounds like you'll tackle that in a subsequent branch, so this is fine for now.

I did uncover a couple other bugs, noted in the comments though.

.editorconfig
```diff
@@ -0,0 +1,9 @@
+# Editor configuration, see https://editorconfig.org
+root = true
+
+[*]
+charset = utf-8
+indent_style = space
+indent_size = 2
+insert_final_newline = true
+trim_trailing_whitespace = true
```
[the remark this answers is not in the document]
rpeterman-gp: Is that a thing for config files? I thought final newlines where for files that would be combined/included into other files.

app/javascript/controllers/audio_breakpoint_controller.js
rpeterman-gp: Do you have a specific episode this is happening on?

app/javascript/controllers/audio_breakpoints_controller.js
rpeterman-gp: Did know there was a validate method on the model. Added that to the existing controller value attribute.

app/javascript/controllers/audio_breakpoint_controller.js
rpeterman-gp: Found what was causing this. Fixed.

app/javascript/controllers/audio_breakpoints_controller.js
rpeterman-gp: Added `adBreaksValid` value and use it to prevent initializing/updating UI with invalid number of ad breaks.

cavis (commented):
Okay, after much QA, the only bug I could find is:

When you initially have a `nil` `segmentation` field in the db, this form always shows as "changed" (with the discard button visible).

You can simulate this by altering an episode from the console:

```ruby
Episode.find_by_guid(<guid>).uncut.update(segmentation: nil)
```

cavis (approved):
Okay, sounds like the nil/changed issue will be addressed later on, since this needs to change from "ad breaks" to "segmentation" anyways.

👍 onwards!

PR #845, opened 2023-09-19
Set empty link after save

### LINKED ISSUE #836: New podcasts get a bad default Homepage Link
When you create a new Podcast in feeder, and leave the Homepage Link blank, you end up with something like:

```
https://play.staging.prx.tech/listen?uf=https%3A%2F%2Ff-staging.prxu.org%2F%2Ffeed-rss.xml
```

That f-staging feed url is incorrect.  I'm guessing because this is being set before saving, there is no podcast id to put in there.

Fixitfixitfixit

radical-ube:
resolves #836

### DIFF
--- a/app/models/podcast.rb
+++ b/app/models/podcast.rb
@@ -47,6 +47,7 @@ class Podcast < ApplicationRecord
   validates :explicit, inclusion: {in: VALID_EXPLICITS}, allow_nil: false
 
   before_validation :set_defaults, :sanitize_text
+  after_save :set_empty_link
 
   scope :filter_by_title, ->(text) { where("podcasts.title ILIKE ?", "%#{text}%") }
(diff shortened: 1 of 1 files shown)


app/models/podcast.rb
```diff
@@ -66,7 +67,12 @@ def self.release!(options = {})
 
   def set_defaults
     self.explicit ||= "false"
-    self.link ||= embed_player_landing_url(self)
+  end
+
+  def set_empty_link
+    if link.nil?
```
radical-ube, line 73: this should only update `link` if it's still blank after saving, whether it's a new podcast or updating an existing one.

cavis: Minor nitpick: I might use `.blank?` here, to catch empty strings `""` as well?  Just in case the web forms submitted a blank string.

cavis (approved):
One minor nitpick, but this feels like the right solution.

PR #847, opened 2023-09-20
default to .mp3 for any non-wav non-flac media

### LINKED ISSUE #837: Enclosure extension mismatch when using default mp3 audio format
If you create a Feed with no audio format, and upload a non mp3/wav/flac file ... you end up with an enclosure ending in, say, `.ogg`.  But Dovetail CDN is going to serve you the "default audio format" (an mp3).

So to fix the [enclosure_template_default](https://github.com/PRX/feeder.prx.org/blob/main/app/models/feed.rb#L65)... we'd want to [check here](https://github.com/PRX/feeder.prx.org/blob/main/app/models/enclosure_url_builder.rb#L66) if the `ex` extension was to one of mp3/wav/flac.  AND MAKE SURE this is actually an audio episode, not video or something else.  If so, change the extension to the default `mp3`.

radical-ube:
resolves #837

### DIFF
--- a/app/models/enclosure_url_builder.rb
+++ b/app/models/enclosure_url_builder.rb
@@ -52,6 +52,11 @@ class EnclosureUrlBuilder
     fn = File.basename(base[:path].to_s)
     ex = File.extname(base[:path].to_s)
 
+    # extensions only matter for audio, so default to .mp3 otherwise
+    unless media.try(:medium) == "audio" && %w[.wav .flac].include?(ex)
+      ex = ".mp3"
+    end
+
     {
       original_filename: orig_fn,
       original_extension: orig_ex,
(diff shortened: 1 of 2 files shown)


app/models/enclosure_url_builder.rb
```diff
@@ -52,6 +52,11 @@ def podcast_episode_expansions(podcast, episode, feed)
     fn = File.basename(base[:path].to_s)
     ex = File.extname(base[:path].to_s)
 
+    # extensions only matter for audio, so default to .mp3 otherwise
+    unless media.try(:medium) == "audio" && %w[.wav .flac].include?(ex)
```
cavis, line 56: It seems like this would change a `.mp4` video to a `.mp3` extension - which is not what we want.  I think we _only_ want to change when medium `== "audio"`.

Can you also add a test to confirm this?  (That it leaves extensions alone for `medium == "anything-else"` ... but alters it for `"audio"`).

cavis (changes requested):

cavis (approved):
👍 looks good to me!

PR #848, opened 2023-09-20
Fix form updated_at cascading

### LINKED ISSUE #776: Cascading updated_at from UI
For some reason, rails associations with `touch: true` are not cascading up to the parent correctly.  When updated via an `accepts_nested_attributes_for` form.

I did some research, and it seemed like this was a known bug in rails?  May need to dig that up again.

First noticed when updating fields on the default_feed, via the Podcast form.

But we need this if we ever hope to cache things.

cavis:
Fixes #776.

Bit of a weird one - some of our `touch: true` associations weren't bubbling up to the parent records.  But only from UI updates - API and console updates worked fine.

Doing a bunch of QA, the places I saw it was:

- Updated a podcast -> settings description (a field on the `default_feed`) didn't touch the podcast
- Updating image fields (caption/credit) didn't touch parent feed+podcast or episode+podcast
- Sometimes, changing an image/media would update the podcast but _not_ the episode?
- Adding/removing images/media didn't always bubble up

Turns out, it was related to `accepts_nested_attributes_for`.  There was an issue in rails core, but that was solved several major versions ago.  The best suggestion I could find was [this stackoverflow](https://stackoverflow.com/questions/46448140/belongs-to-touch-true-not-fired-on-association-build-and-nested-attribute-assig) ... which recommended marking these nested attributes relations with an `inverse_of`.

I _think_ these problems were only occurring where the assocation name didn't match the model name?  But I couldn't nail that down, so I added the `inverse_of` to every assocation that accepts nested attributes.  Doesn't hurt anything.

### DIFF
--- a/app/models/episode.rb
+++ b/app/models/episode.rb
@@ -23,10 +23,10 @@ class Episode < ApplicationRecord
   serialize :overrides, HashSerializer
 
   belongs_to :podcast, -> { with_deleted }, touch: true
-  has_many :contents, -> { order("position ASC, created_at DESC") }, autosave: true, dependent: :destroy
+  has_many :contents, -> { order("position ASC, created_at DESC") }, autosave: true, dependent: :destroy, inverse_of: :episode
   has_many :media_versions, -> { order("created_at DESC") }, dependent: :destroy
-  has_many :images, -> { order("created_at DESC") }, class_name: "EpisodeImage", autosave: true, dependent: :destroy
-  has_one :uncut, -> { order("created_at DESC") }, autosave: true, dependent: :destroy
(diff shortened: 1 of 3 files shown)


app/models/episode.rb
```diff
@@ -23,10 +23,10 @@ class Episode < ApplicationRecord
   serialize :overrides, HashSerializer
 
   belongs_to :podcast, -> { with_deleted }, touch: true
-  has_many :contents, -> { order("position ASC, created_at DESC") }, autosave: true, dependent: :destroy
+  has_many :contents, -> { order("position ASC, created_at DESC") }, autosave: true, dependent: :destroy, inverse_of: :episode
```
kookster, line 26: Yeah, the advice I see is that the Rails built in ability to figure out these relationships is not great, if you set any other options, and there seems to be no harm in setting explicitly.

PR #853, opened 2023-09-21
Delegated delivery reupload

### LINKED ISSUE #749: Replace media without unlinking the container
https://github.com/PRX/feeder.prx.org/pull/745 set up a new sentinel on linked audio asset container because we can no longer rely on the delivery and delivery file instances being in place. But that left the [re-upload logic ](https://github.com/PRX/feeder.prx.org/blob/205204651f20bbeea19fabb0f91db5dd60f2da7c/app/models/episode.rb#L225-L229)stranded.  The current setup works, but does lack the ability to mark the episode for re-upload upon swapping media.

It's possible to set up a "replace" flow: instead of unlinking the audio asset container from the episode, we can upload the media and let the file be replaced on the container.  This should allow the subscriber media to sit in place seamlessly until the replacement is processed and ready.

To finish up the patch from #755
- [x] Set up a new boolean column on the episode, like `#needs_apple_delivery?`. By default it should be true
- [x] use the new `#needs_apple_delivery?` in `Apple::Episode#needs_delivery?` in lieu of references to `apple_hosted_audio_asset_container_id` 
- [x] Fixup the [guard here](https://github.com/PRX/feeder.prx.org/blob/205204651f20bbeea19fabb0f91db5dd60f2da7c/app/models/apple/podcast_container.rb#L25) to make sure we're probing the top-level `Apple::Episode#needs_delivery?` (not the container)
- [x] When the upload processing is successful, mark the episode `needs_apple_delivery = false`
- [x] Integrate `Episode#needs_apple_delivery?` into `Episode#synced_with_apple?` so that we can re-initiate the delivery with that column set.
- [x] in the case where we are re-uploading, [set `needs_apple_delivery` to true](https://github.com/PRX/feeder.prx.org/blob/205204651f20bbeea19fabb0f91db5dd60f2da7c/app/models/episode.rb#L225-L229): 
- [x] increment the file name. The delivery file processing is somehow keyed on the filename in the container, So make sure that the filename is incremented somehow, with a counter like `"somefilename-#{current_attempt}.mp3"`

svevang:
closes #749

This adds a column to the episode to enable more explicit state managements around episode + media delivery.  This will allow the episode media to be reuploaded without disconnecting the container (as in https://github.com/PRX/feeder.prx.org/pull/745 ).


db/migrate/20230921203839_add_fetch_count_to_podcast_container.rb
```diff
@@ -0,0 +1,5 @@
+class AddFetchCountToPodcastContainer < ActiveRecord::Migration[7.0]
+  def change
+    add_column :apple_podcast_containers, :source_fetch_count, :integer, default: 0, null: false
```
cavis, line 3: +1 this seems useful to keep track of.

db/migrate/20230920153421_add_needs_apple_delivery_to_episode.rb
```diff
@@ -0,0 +1,29 @@
+class AddNeedsAppleDeliveryToEpisode < ActiveRecord::Migration[7.0]
+  def change
+    add_column :episodes, :needs_apple_delivery, :boolean, default: true
```
cavis, line 3: Just curious - why track on the `Episode` vs the `Apple::Episode`?  Or I guess that's not a table ... so the `Apple::PodcastContainer`?  It seems like you've already got some logic in there to check if it needs re-delivery.  Do you need to set this flag when that container doesn't exist yet or something?

(I'm thinking that querying for `Episode.where(needs_apple_delivery: true)` wouldn't exactly be accurate, since that returns shows that don't do delegated delivery.  Where `Apple::PodcastContainer.where(needs_apple_delivery: true)` would.)

cavis (approved):
One question on where the `needs_apple_delivery` column lives.  But otherwise looks good to me.

db/migrate/20230920153421_add_needs_apple_delivery_to_episode.rb
svevang: Really good point, let's move this off the episode.

app/models/apple/publisher.rb
```diff
@@ -264,7 +264,14 @@ def mark_delivery_files_uploaded!(eps)
         pdfs = eps.map(&:podcast_delivery_files).flatten
         ::Apple::PodcastDeliveryFile.mark_uploaded(api, pdfs)
 
+        # link the podcast container with the audio to the episode
         res = Apple::Episode.update_audio_container_reference(api, eps)
+        # update the feeder episode to indicate that delivery is no longer needed
+        eps.each do |ep|
+          Rails.logger.info("Marking episode as no longer needing delivery", {episode_id: ep.feeder_episode.id})
+          ep.feeder_episode.apple_has_delivery!
```
cavis, line 272: Could this be a race condition?

1. Media version A comes in, fires a publish job
2. We get partway through the job, media version B comes in, fires another publish job
3. We get here, mark delivered = true

I'm not sure that's actually a problem - probing for new media versions via #858 is the real fix for the why-did-the-user-swap-media-files-so-immediately conditions.

And this explicit control to force re-uploading is handy in any case.

app/models/episode.rb
```diff
@@ -230,14 +249,18 @@ def copy_media(force = false)
...
     apple_podcast_deliveries.reset
     apple_podcast_container&.podcast_deliveries&.reset
+    apple_needs_delivery!
   end
 
   def publish!
-    apple_mark_for_reupload!
-    podcast&.publish!
+    Rails.logger.tagged("Episode#publish!") do
+      apple_mark_for_reupload!
```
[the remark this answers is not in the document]
svevang: The intent here was to capture the fluctuations in media, and yes #858  should handle changes in media. But you're right, this needs to be removed to preserve the current/existing behavior.

Currently if the episode is [`synced_with_apple?`](https://github.com/PRX/feeder.prx.org/blob/main/app/models/apple/episode.rb#L456-L458) then we skip it completely in the publishing flow.

PR #863, opened 2023-09-28
Timepicker woes

### LINKED ISSUE #799: Date/Timepicker Woes
This component has some issues.

- [x] After loading the podcast planner, the Start Date field is prefilled with a default value. If you focus on the field but make no changes, when it loses focus, the value goes away (or becomes hidden)
- [x] Selecting timestamp by clicking a date results in angry red `!` icon for @farski (may need to check in for clues on how to recreate this)
- [x] When the Dropdate picker is active, the field doesn't display a timezone, but when you defocus, a timezone appears. If you try to type in basically any valid timezone (except maybe the local one?) the entire field gets cleared out.  (3ef1257 may have fixed this)
- [x] On the planner, typing in the Start Date or End Date fields results in Turbo frame and other errors
- [x] If you open a draft episode that has a Dropdate aleady set, when you click into the Dropdate field you don't get a date picker. The field seems to reject most values that are typed in by hand, so it becomes hard/impossible to change the Dropdate in this case (RETEST, per 3ef1257)

cavis:
Fixes #799.  Forever.

This covers a few UI aspects so I'll list them out.

### Episode Planner

image

1. Drastically simplifies `form.date_field` fields (the start/end dates).  And drops support for `form.time_field` altogether (we don't use them in Feeder).
2. Removes the short timezone name from the `Publish Time` field.  It's now just a plain `HH:MM` dropdown.
3. Adds a `Time Zone` field, containing the Rails "friendly" generic time zone names.  After JS loads, it changes it from UTC to your browser's local timezone.

### Episode Status Form

image

1. Starts out as a readonly field, displaying the `local_time` gem localized timestamp string.  (Including daylight savings).
2. When you click edit, shows the timestamp broken out into 3 fields - date, time, zone
3. Date is a basic (but typeable) flatpickr, with validations
4. Time is a basic (byt typeable) flatpickr, with validations
5. Zone is the same generic time zone names as above
6. Date/Time are combined in a hidden field and submitted to the server separately from the time zone.


app/controllers/episodes_controller.rb
```diff
@@ -134,6 +134,14 @@ def episode_params
       :original_guid,
       categories: [],
       images_attributes: %i[id original_url size alt_text caption credit _destroy _retry]
-    )
+    ).tap do |p|
+      p[:released_at] = released_at_zone.parse(p[:released_at]) if p[:released_at].present?
```
radical-ube, line 138: trying to understand this: if someone has chosen a release date, we need to also check the time zone that was selected, and then parse the final date off of that zone using the release date given.

app/helpers/feeder_form_builder.rb
```diff
@@ -51,19 +62,13 @@ def check_box(method, options = {})
...
 
-  def time_field(method, options = {})
-    value = options[:value] || object&.public_send(method)
-    options[:value] = value.try(:strftime, "%Y-%m-%d %-l:%M:%S %p %Z") || value
-    add_data(options, :timestamp, true)
-    add_flatpickr_controller(options)
-    text_field(method, options)
+  def time_zone_field(method, options = {})
+    add_data(options, :controller, TIME_ZONE_CONTROLLER)
+    select method, IMPORTANT_ZONES, {selected: "UTC"}, options
```
radical-ube, line 71: will we ever need other time zones as options here? or is it generally assumed that the selected timezone will either be a US time zone, and anything else that doesn't gets the `"UTC"` label? (unless i'm reading this `select` component incorrectly.)

radical-ube: also, was the `time_field` component removed out of here completely? i guess it makes sense since you would parse time off of a given value in the `time_field`, but that's not until a later step. first step is selecting a time; the time is used as an argument with a `zone.parse()` method in order to correctly give the full date and time, which is what we are trying to save in the db

app/javascript/controllers/dropdate_picker_controller.js
```diff
@@ -0,0 +1,81 @@
+import { Controller } from "@hotwired/stimulus"
+import flatpickr from "flatpickr"
+
+export default class extends Controller {
```
radical-ube, line 4: so my understanding of this controller is that it is specifically for the `dropdate` "field" of the Episode form, and it's main function is to manage hiding/showing/editing each of the separate date/time/zone fields that all contribute to the specific value presented in the `dropdate` field.

app/helpers/feeder_form_builder.rb
cavis: We could certainly add more, but we don't need to for this to work.  UTC is rendered server-side, but we change what's selected in JS.

If you look at the `time_zone_controller.js`, it handles this dropdown.  If it detects you're in one of these "important zones", it selects it.  Otherwise, it _adds a new option_ for your timezone string and selects that.  So should be compatible with any zone (I hope).

Yeah, I removed `time_field` entirely for now (it greatly simplifies the `flatpickr_controller.js`).  And the `dropdate_picker_controller.js` became its own thing - coordinating the date/time fields.  I could see making that a more generic component, if we find we want to reuse it somewhere else.

app/javascript/controllers/dropdate_picker_controller.js
cavis: You got it.  Though you'll notice, this controller doesn't _really_ deal with the time zone (other than adding a "change" action to it).  I abstracted all the timezone-dropdown handling into a different controller.

This controller handles the date+time fields, parsing/validating them in your locale styles (`M/D/YYYY` and `h:mm:ss AM`). And recombining them into a hidden ISO date field.

app/javascript/controllers/dropdate_picker_controller.js
```diff
@@ -0,0 +1,81 @@
...
+      allowInput: true,
+      dateFormat: "h:i:S K",
+      enableTime: true,
+      enableSeconds: true,
+      noCalendar: true,
+    })
+  }
+
+  editing(event) {
+    event.preventDefault()
```
radical-ube, line 35: just checking, but the default action here would be some sort of `submit`? because this `dropdate_picker#editing` action is used to toggle which fields the user sees (the showing field vs. the editing fields)

app/javascript/controllers/dropdate_picker_controller.js
```diff
@@ -0,0 +1,81 @@
...
+    this.showingTargets.forEach((el) => el.classList.add("d-none"))
+    this.editingTargets.forEach((el) => el.classList.remove("d-none"))
+  }
+
+  change() {
+    if (this.dateTarget.value && isNaN(new Date(this.dateTarget.value))) {
```
radical-ube, line 41: are these "validations" here because we can't otherwise validate these fields server-side? is the `is-invalid` class mainly for styling to inform the user something is incorrect? does it do anything on submission or otherwise raise an error when the request is made?

app/javascript/controllers/dropdate_picker_controller.js
```diff
@@ -0,0 +1,81 @@
...
+      this.dateTarget.classList.add("is-invalid")
+    } else {
+      this.dateTarget.classList.remove("is-invalid")
+    }
+
+    if (this.timeTarget.value && isNaN(new Date(`2020/01/01 ${this.timeTarget.value}`))) {
```
radical-ube, line 47: assuming that this date is arbitrary and we're only using the time value here anyway?

app/javascript/controllers/dropdate_picker_controller.js
```diff
@@ -0,0 +1,81 @@
...
+      this.timeTarget.classList.add("is-invalid")
+    } else {
+      this.timeTarget.classList.remove("is-invalid")
+    }
+
+    // set released_at field to datetime without zone
+    const fullUtcDate = new Date(`${this.dateTarget.value} ${this.timeTarget.value} UTC`)
+    if (isNaN(fullUtcDate)) {
```
radical-ube, line 55: oh i see, we just clear the field here if the date and time are "invalid"?

and then in the else case, we give an ISO but clear out the timezone "Z".

app/javascript/controllers/flatpickr_controller.js
```diff
@@ -1,123 +1,28 @@
...
-    if (this.isTimestamp) {
-      const name = this.element.name
-      this.element.removeAttribute("name")
-      this.element.insertAdjacentHTML("afterend", `<input type="hidden" name="${name}">`)
-      this.hiddenField = this.element.nextSibling
-      this.hiddenField.value = this.element.value
-    }
-
-    this.picker = flatpickr(this.element, {
+    flatpickr(this.element, {
```
radical-ube, line 8: so for the most part this controller is "gutted" so it has a fairly minimal amount of customization in it. this is still used in Podcast Planner? do we use (this version of) flatpickr anywhere else in Feeder? i know the dropdate picker also initializes flatpickr but it's a little differently used than here?

app/javascript/controllers/dropdate_picker_controller.js
cavis: I think the little edit-pencil is a `<a>` link, not a button.  So clicking it will by default try to navigate.  Which we don't want.

app/javascript/controllers/dropdate_picker_controller.js
cavis: Yep - just a way to validate the time string.  And there were no leap seconds on 2020-01-01 ... so should be good all the way up to `11:59:59 PM`!

app/javascript/controllers/dropdate_picker_controller.js
cavis: The server will also throw a rails validation error, if it isn't a valid timestamp.  Or in some cases, rails will clear out a timestamp field if it doesn't recognize the input string?  Or mis-parse it?

The fuzzy nature of how rails parses datetime strings made me want to handle as much as I could upfront.  Fast feedback for the user!

app/javascript/controllers/dropdate_picker_controller.js
cavis: Yeah, I think that was me getting "unexpected" results if you submit a weird looking string to Rails.  So I opted to just clear it out on the frontend.

app/javascript/controllers/flatpickr_controller.js
cavis: Yep, the Planner still uses this.  We can revisit making this more customizable if we add more date/time fields to Feeder.  But for now, it seemed like all my hundreds of lines of custom code in here weren't helping things.

app/javascript/controllers/time_zone_controller.js
```diff
@@ -0,0 +1,48 @@
+import { Controller } from "@hotwired/stimulus"
+import lookupFriendlyTimeZone from "util/lookupFriendlyTimeZone"
+
+export default class extends Controller {
+  connect() {
+    const resolvedZone = Intl.DateTimeFormat().resolvedOptions().timeZone
```
radical-ube, line 6: is `resolvedZone` via [resolvedOptions().timeZone](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/resolvedOptions) a way to get a user's default time zone? not fully sure how `Intl.DateTimeFormat()` works, but since nothing is passed in here, i'm assuming it either defaults to UTC or the user's time zone?

app/javascript/controllers/time_zone_controller.js
```diff
@@ -0,0 +1,48 @@
...
+    const friendlyZone = lookupFriendlyTimeZone(resolvedZone)
```
radical-ube, line 7: oh, ok. i read ahead to the `lookupFrinedlyTimeZone` util, which is a mapping of the local time zones to it's "general use"(?) counterpart. so maybe that answers my above question?

app/javascript/controllers/time_zone_controller.js
cavis: The docs state it's either a String, or undefined if it cannot be determined.  (Though I've only seen timezone strings testing locally - but this handles undefined as well).

app/javascript/controllers/time_zone_controller.js
cavis: You got it.  And right below this, we change the timezone picker to be either this friendlyZone (`Mountain Time`) or whatever string the browser returned (`America/Denver` style).

And if we have no `resolvedZone` - we leave the field alone.  And UTC will be selected, since that's what the server rendered in the view.

app/views/episodes/_form_dropdate.html.erb
```diff
@@ -0,0 +1,52 @@
+<% editing = policy(episode).update? && (episode.published_or_released_date.blank? || episode.released_at_changed?) %>
+
```
radical-ube, line 2: to my understanding this is the partial that houses all of the date/time/zone fields that are displayed/hidden through the "edit pencil" icon.

radical-ube (approved):
reviewed to the best of my understanding and seems to work.

app/views/episodes/_form_dropdate.html.erb
```diff
@@ -0,0 +1,52 @@
...
+<div class="row" data-controller="dropdate-picker">
+  <%= form.hidden_field :released_at, value: episode.published_or_released_date&.iso8601, data: {dropdate_picker_target: "input"} %>
```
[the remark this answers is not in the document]
cavis: Yep - making javascript reassemble an actual ISO8601 string, so we're pretty sure the server can parse it.

PR #868, opened 2023-10-02
Feeder episode deletion archives delegated delivery

svevang:


app/models/apple/publisher.rb
```diff
@@ -35,27 +35,50 @@ def podcast
...
+      # - episodes that have fallen off the end of the feed (Feed#display_episodes_count)
+      eps
+        .reject { |ep| eps_to_sync.include?(ep) }
+        .reject(&:apple_new?)
         .reject(&:archived?)
     end
 
+    def episodes_to_archive
+      # look at the global list of episodes, not just the private feed
+      filter_episodes_to_archive(show.podcast_episodes)
```
cavis, line 66: Okay, and here we're looking at any episode in the podcast that we've already synced to Apple, and archiving them if they fell out of the private feed.

cavis (approved):
👍 looks good to me

app/models/apple/show.rb
```diff
@@ -147,9 +148,26 @@ def get_show
       self.class.get_show(api, apple_id)
     end
 
+    def sort_by_episode_properties(eps)
+      eps.sort_by do |e|
+        [
+          e.deleted_at.nil? ? 1 : -1,
+          e.published_at.present? ? 1 : -1,
+          e.published_at || e.created_at,
+          e.created_at
```
cavis, line 157: Just seeing if I understand this - so with the `.reverse` this is sorting by:

1. created_at asc
2. published_at asc
3. non-blank published at, followed by blank (drafts)
4. non-deleted, followed by deleted

What's the reasoning for that?  So we always sync the oldest episodes first?

What happens in the case where you delete an episode, then create/publish a new version of it with the same <guid>?  Is this sorting so we always process the deleted one first?

test/models/apple/show_test.rb
```diff
@@ -19,37 +19,96 @@
...
+        recs = [
+          OpenStruct.new(deleted_at: now, published_at: now, created_at: now, id: 1),
+          OpenStruct.new(deleted_at: now, published_at: now + 1.second, created_at: now, id: 2),
+          OpenStruct.new(deleted_at: now, published_at: now + 1.second, created_at: now + 1.second, id: 3),
+          OpenStruct.new(deleted_at: nil, published_at: now + 1.second, created_at: now + 1.second, id: 4),
+          OpenStruct.new(deleted_at: nil, published_at: now + 1.second, created_at: now + 2.second, id: 5),
+          OpenStruct.new(deleted_at: nil, published_at: now + 2.second, created_at: now + 2.second, id: 6)
+        ]
+
+        assert_equal [6, 5, 4, 3, 2, 1], apple_show.sort_by_episode_properties(recs).map(&:id)
```
cavis, line 75: Okay, so this is non-deleted before deleted, then published_at asc, created_at asc.

cavis (approved):
👍 some questions about the sort, but I _think_ the tests turn out the way I expect anyways.

app/models/apple/show.rb
cavis: Update: i read this wrong.  The `reverse` is on the sort_by, not on the array.

But probably still worth a comment explaining why which of these is most important.

PR #869, opened 2023-10-03
Add slight debounce to redrawing chart

### LINKED ISSUE #855: Cache podcast downloads chart
Right now, every time you load up `podcasts#show`, it fires off a bunch of Castle API requests to get the last 30 days of download metrics.

For daily shows, this turns out to be a lot.  And makes the chart look somewhat comical as it loads.

Cache this, somewhere.  Maybe localstorage since this is all done client-side?  Or something else?

cavis:
Fixes #855.

Turns out, I just needed to debounce calling `chart.redraw()`.  Didn't bother to add any caching - Castle is returning these 40+ results real quick (IIRC, Castle has its own caching layer).


brandonhundt (approved):
Pretty snappy. Thanks!

PR #871, opened 2023-10-04
Index page speedups

### LINKED ISSUE #857: Podcast/Episode index improvements
Working on a show with a large number of episodes, it occurs to me we should:

- Cache parts of these in Redis (russian doll style)
- Allow filtering episodes by media status, so you can find incomplete media buried in the back catalog
- Other filters?
- Explicit page controls.  I feel like we could/should allow showing "page-size = all"?  It would be nice if these were sticky (in the session) as well.

cavis:
Fixes #857.

Adding caching and other improvements to index pages, so they work more good when you have a ton of stuff to sort through.

- Reorganize podcast/episode index a bit, so they match each other more closely in layout
- Use collection caching on index pages
- Add sorting, caching, and per-page settings
- Also cached the podcast-switcher dropdown a bit


brandonhundt (approved):
this is faster, and I think will be more helpful for producers to find episodes. I've added styling, when all this works for you, it works for me.

PR #872, opened 2023-10-04
Audio Segmenter Pre and Post Roll Controls

### LINKED ISSUE #832: Audio segmenter trims the start/end of the file
In addition to inserting midrolls in a file, we'll need the ability to (optionally) "trim" the start and/or end of the file.

Will need to switch from using the `ad_breaks` virtual field to the actual underlying database `segmentation` field.  Which is an array of arrays of times we're _keeping_.

So `segmentation = [[3, 4], [6, 10]` results in 2 segments.  From 3-4 seconds, and from 6-10 seconds.  And `nil` / `null` can be used to indicate the segment starts at 0 seconds, or ends at the end of the file.

rpeterman-gp:
Closes #832 

- Adds Pre and Post Roll controls to waveform and breakpoint inputs
- Changes ad breaks init and out put to segmentation arrays
- Fixes segmenter output change detection

REVIEW NOTES:
I left in the old ad_breaks code in the models and controllers. Wasn't sure what to lose and what to keep. Please point out what needs to go in the code review.


cavis (approved):
:+10: nice!  This is great functionality to add.

Issue #874, opened 2023-10-06
Preserve Uncut filenames


cavis:
Thinking of 2 cases:

1. If you upload a single Uncut segment and slice that, the resulting Contents all have an `original_url` pointing to the randomized-guid S3 filename of the Uncut.  So you end up with a Dovetail enclosure URL filename that's just a random guid.
2. Similarly, if you upload a Content segment and convert it to Uncut later ... the Uncut ends up with an `original_url` pointing to a randomized guid S3 filename.

cavis:
For case (1), I solved it by reusing the original_url https://github.com/PRX/feeder.prx.org/blob/main/app/models/episode.rb#L208.  At least if the episode is fresh.

I didn't solve case (2).  But that's far less likely to happen, and I don't think we generally care about those file names anyways.

Calling this good.

PR #875, opened 2023-10-06
Segment existing contents

cavis:
So previously, if you changed an episode from "Upload single file" to "Upload multiple files", any existing contents we sliced from your single file will show up as the "multiple files".

This kind of does the inverse: if your "Upload multiple files" episode has a completed file in position 1, we preset that as the new file.  But - without saving it yet - clicking save will reprocess it.

image

Did notice we tend to get guid filenames doing these types of conversions - opened #874 to figure that out.

### DIFF
--- a/app/models/episode.rb
+++ b/app/models/episode.rb
@@ -223,10 +223,20 @@ class Episode < ApplicationRecord
     super
 
     if medium_changed? && medium_was.present?
-      unless medium == "audio" && medium_was == "uncut"
+      if medium_was == "uncut" && medium == "audio"
+        uncut&.mark_for_replacement
+      elsif medium_was == "audio" && medium == "uncut"
+        if contents.first&.status_complete?
+          (uncut || build_uncut).tap do |u|
+            u.original_url = contents.first.url
+            u.file_size = contents.first.file_size
+            u.duration = contents.first.duration
+          end
+        end
+        contents.each(&:mark_for_replacement)
+      else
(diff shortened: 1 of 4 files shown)


app/models/episode.rb
```diff
@@ -223,10 +223,20 @@ def medium=(new_medium)
...
+      elsif medium_was == "audio" && medium == "uncut"
+        if contents.first&.status_complete?
+          (uncut || build_uncut).tap do |u|
+            u.original_url = contents.first.url
+            u.file_size = contents.first.file_size
+            u.duration = contents.first.duration
+          end
+        end
+        contents.each(&:mark_for_replacement)
+      else
```
cavis, line 237: Refactored to be more explicit, in how it handles uncut->audio and audio->uncut.

app/views/episode_media/media/_new.html.erb
```diff
@@ -55,15 +55,17 @@
   </div>
 
   <%# step 3a: hidden original_url field pointing to the uploaded temporary s3 file %>
-  <div class="form-floating d-none input-group" data-upload-target="success">
+  <div class="form-floating input-group <%= "d-none" unless upload_uploaded?(media) %>" data-upload-target="success">
```
cavis, line 58: Had to change a couple `d-none`s rendered by the server.  If we already have an uploaded file, skip straight to this step 3 which shows the uploaded file name, with the is-changed orange border.

PR #881, opened 2023-10-10
HOTFIX: post roll handle initial render

rpeterman-gp:
Ensures post roll handle doesn't render off canvas.

Noticed the post roll handle in wave form would show up initially on https://feeder.prx.test/episodes/1b1d0519-83b6-4150-ba1e-d4b362f69d45/media until edited via the ad break inputs. This PR should ensure the post roll segment has some renderable length in the waveform editor.


cavis (approved):
👍

Issue #864, opened 2023-09-28
Convert single Content to Uncut


cavis:
After an RSS import, we end up with a bunch of 1-segment `episode.contents`.

Sometimes, we want to slice that up.  Either programmatically adding timings, or converting/slicing in the UI.

Make some model helper methods to change the rails STI on these.  And expose in the UI somehow (cc: @brandonhundt).

PR #882, opened 2023-10-10
Podcast Import STI

cavis:
Setting up tooling for #864.

Getting ready for importing timings, and general bug fixes as I import a whole bunch of RSS feeds locally.

- Convert `PodcastImport` and `EpisodeImport` models to be STI base classes
- Simplify status fields, in prep for other flavors of import
- Move existing import code to `PodcastRssImport` and `EpisodeRssImport`


app/controllers/imports_controller.rb
```diff
@@ -15,24 +15,17 @@ def show
...
+    @imports = @podcast.podcast_imports.order(created_at: :asc)
     @import = @podcast.podcast_imports.new(import_params)
     authorize @import
-    begin
-      @import.get_feed
-    rescue URI::InvalidURIError
-      flash.now[:notice] = t(".failure_get_feed")
-      render :index, status: :unprocessable_entity
-      return
-    end
```
cavis, line 27: I moved this into a custom validation, in `PodcastRssImport`.  Since this controller is now generic to many types of imports.

app/controllers/imports_controller.rb
```diff
@@ -15,24 +15,17 @@ def show
...
 
     respond_to do |format|
       if @import.save
-        @import.import_later(import_metadata)
```
cavis, line 31: This `import_metadata` is now a field on the `config` store, saved to the db.

app/jobs/application_job.rb
```diff
@@ -10,6 +10,10 @@ class ApplicationJob < ActiveJob::Base
   rescue_from(StandardError) do |e|
     NewRelic::Agent.notice_error(e)
   ensure
-    raise e
+    if e.is_a? ActiveJob::DeserializationError
+      Rails.logger.warn(e.message)
```
cavis, line 14: Was hitting some errors with callbacks coming in for deleted records.  Seems safe enough to just log these as a warning and move on?

app/models/concerns/import_utils.rb
```diff
@@ -100,6 +94,21 @@ def files_match?(file, url)
     end
   end
 
+  def http_get(url)
+    uri = URI.parse(url)
+    http = Net::HTTP.new(uri.host, uri.port)
+    http.use_ssl = true if uri.scheme == "https"
+    req = Net::HTTP::Get.new(uri)
+    req["User-Agent"] = "PRX-Feeder-Import/1.0 (Rails-#{Rails.env})"
```
cavis, line 102: Moved any HTTP.get requests from the importers into this shared util.  And bonus, I can set a shared user agent string!

app/models/episode_import.rb
```diff
@@ -1,208 +1,50 @@
...
-require "itunes_category_validator"
-require "loofah"
-require "hash_serializer"
-
 class EpisodeImport < ApplicationRecord
   include ImportUtils
 
-  serialize :entry, HashSerializer
-  serialize :audio, HashSerializer
+  store :config, coder: JSON
```
cavis, line 4: I've migrated both `podcast_imports` and `episode_imports` table to have this "config" [store field](https://api.rubyonrails.org/classes/ActiveRecord/Store.html).  It's still JSON-encoded, but a bit more modern than the HashSerializer.

In general, these base classes are now pretty minimal.  The diff is large here, but this code all migrated to `app/models/imports/episode_rss_import`.

app/models/imports/episode_rss_import.rb
```diff
@@ -0,0 +1,130 @@
...
+    set_file_resources!
+
+    episode.save!
+    episode.copy_media
+
+    status_complete!
+  rescue => err
+    status_error!
+    raise err
+  end
```
cavis, line 26: Everything below here is copied from `episode_import.rb`.  The code above here was reorganized a bit, but the same logic as it used to have.

app/models/imports/podcast_rss_import.rb
```diff
@@ -0,0 +1,293 @@
...
+
+    create_or_update_podcast!
+    create_or_update_episode_imports!
+
+    status_importing!
+  rescue => err
+    status_error!
+    unlock_podcast!
+    raise err
+  end
```
cavis, line 85: Similarly, everything below here was copied over.

Changes above this:

1. `feed_rss` (the raw RSS string) is now saved in the `config` field as well.
2. `feed` (the parsed RSS) is memoized (provided some `feed_rss=` and `url=` setters to refresh it)
3. RSS parsing is now a custom validation `validate_rss`
4. `import_metadata` getter/setter to power the form checkbox (which is the inverse of the `episodes_only` db field).

app/models/podcast_import.rb
```diff
@@ -1,379 +1,61 @@
...
-    saved: SAVED,
-    failed: FAILED,
-    feed_retrieved: FEED_RETRIEVED,
+    started: STARTED,
     importing: IMPORTING,
-    retrying: RETRYING,
-    started: STARTED
+    complete: COMPLETE,
+    error: ERROR
   }, prefix: true
```
cavis, line 17: Again, setting up a simple structure for how the base classes work.  The rest of this code migrated to the `podcast_rss_import.rb` class.

app/views/imports/_form_main.html.erb
```diff
@@ -12,31 +12,30 @@
         </tr>
       </thead>
       <tbody>
-        <% if imports.first.persisted? %>
```
cavis, line 15: Just changed the indent here - I'm not entirely clear what this check was doing?

test/models/imports/episode_rss_import_test.rb
```diff
@@ -0,0 +1,166 @@
+require "test_helper"
+
+describe EpisodeRssImport do
```
cavis, line 3: This test and `podcast_rss_import_test.rb` are unchanged - just copied as-is out of the previous base class tests.

(Though I did also cleanup the `test/fixtures` that weren't being used - appeared to be some code copied out of CMS that wasn't necessary here).

app/models/episode_import.rb
```diff
@@ -1,208 +1,50 @@
...
-  def entry_description(entry)
-    atr = entry_description_attribute(entry)
-    clean_text(entry[atr])
+  def set_podcast_import_status
+    podcast_import.status_from_episodes! if status_previously_changed?
   end
 
-  def entry_description_attribute(entry)
-    [:content, :itunes_summary, :description, :title].find { |d| !entry[d].blank? }
+  def import!
```
radical-ube, line 36: i originally saw this empty method and was initially confused. but i think the idea is that the method needs to be defined in the parent, but in practicality the class type will have a more specific way to `import!` depending on what and how probably?

app/models/episode_import.rb
```diff
@@ -1,208 +1,50 @@
...
-    episode.itunes_type = entry[:itunes_episode_type] unless entry[:itunes_episode_type].blank?
-
-    episode
+    self.config ||= {}
   end
 
-  def entry_description(entry)
-    atr = entry_description_attribute(entry)
-    clean_text(entry[atr])
+  def set_podcast_import_status
```
radical-ube, line 32: for the "new" hook, but i think this is supposed to handle status updating before getting to the view. (i remember feeling confused about this when first working on imports; for some reason status was defined at the view level, i think)

app/models/episode_import.rb
cavis: Yep - fairly common pattern with inheritance.  If there was some "shared" work to do, I would have put it here.  But these importers seem pretty distinct.

app/models/episode_import.rb
cavis: Yep - as `episode_imports` change status, they have the parent `podcast_import` also recalculate-and-save its status.  So you can easily query the db for things like `PodcastImport.status_complete.count`.

app/models/podcast_import.rb
```diff
@@ -1,379 +1,61 @@
...
-    if (podtrac = link.try(:match, /\/(\w+\.podtrac.com\/.+?\.mp3\/)/))
-      prefix += podtrac[1]
-    end
-    if (blubrry = link.try(:match, /\/(media\.blubrry\.com\/[^\/]+\/)/))
-      prefix += blubrry[1]
-    end
-    prefix
-  end
+  def status_from_episodes!
+    stats = with_lock { episode_imports.group(:status).count }
```
radical-ube, line 28: what does `lock`ing do in this context?

app/views/imports/_form.html.erb
```diff
@@ -1,4 +1,4 @@
-<%= form_with(url: podcast_imports_path(import.podcast), model: [import.podcast, import], method: :post, html: {autocomplete: "off"}) do |form| %>
+<%= form_with(url: podcast_imports_path(import.podcast), model: [import.podcast, import.becomes(PodcastImport)], method: :post, html: {autocomplete: "off"}) do |form| %>
```
radical-ube, line 1: is the `becomes` method how the form can expect that the `type` of class might change?

app/models/podcast_import.rb
cavis: It puts a row lock on this single `podcast_import.id`.  So only one thread/process can run that "count by status" SQL query at a time.

Imagine there are only 2 episode imports, and this method is called at the exact same time from both of them.  So the query returns `{"importing": 1, "complete": 1}` to both, and neither completes this PodcastImport.

In hindsight - I should probably wrap this whole method in that `with_lock do`.  To prevent the podcast_import status update from similarly racing.  I'll make that update.

app/views/imports/_form_main.html.erb
radical-ube: i wanna say this had something to do with the case of an empty `imports` table besides the "new" one, which is defined in the controller so it means `imports` will never be "empty"? so i guess it's in essence an "empty" check?

config/application.rb
```diff
@@ -13,13 +13,15 @@ class Application < Rails::Application
     # Application configuration should go into files in config/initializers
     # -- all .rb files in that directory are automatically loaded.
     config.autoload_paths += %W[
+      #{config.root}/app/models/imports
```
radical-ube, line 16: i don't mess with config files much, but what does it do to `autoload` or `eagerload` these dirs?

cavis: By default in development, when you reference something like `SomeModel` in Rails, it's going to go looking in places like `app/models/some_model.rb` (converting camel case to underscores).  And for something like `Blah::SomeModel` - it's going to look in places with a subfolder like `app/models/blah/some_model.rb`.

And if it doesn't find the file there, it throws an error.

This would be problematic for the `PodcastRssImporter` ... because i threw it in the imports directory without calling it `Imports::PodcastRssImporter`.  It's an aesthetic thing - I tend to like leaving the namespace out, but keeping them organized in a subfolder.  You'll notice other sub-models (like `Content` and `Uncut`) are in the top-level directory.

app/views/imports/_form_main.html.erb
cavis: That was my guess as well.  At some point since then, the "new" import was moved to the `@import` variable, and not appended to `@imports` table.

radical-ube (approved):
looks good, reviewed to the best of my understanding +

PR #884, opened 2023-10-11
Setup new batch for delegated delivery

svevang:
Some fixes for another batch of shows.

Notably, this fixes the case where the feed item guid is set to the Apple episode id. In this case the remote Apple podcasts RSS poller will join the episode via the Apple episode id, instead of the guid.


app/models/apple/config.rb
```diff
@@ -55,14 +62,28 @@ def self.find_or_build_private_feed(podcast)
     end
 
     def self.build_apple_config(podcast, key)
-      ac = Apple::Config.new
+      if has_apple_config?(podcast)
+        Rails.logger.error("Found existing apple config for #{podcast.title}!")
+        Rails.logger.error("Do you want to continue? (y/N)")
+        raise "Stopping build_apple_config" if $stdin.gets.chomp.downcase != "y"
+      end
```
svevang, line 69: It's possible we can completely automate this at some point, and drop these checks.

app/models/apple/config.rb
```diff
@@ -55,14 +62,28 @@ def self.find_or_build_private_feed(podcast)
...
       ac
     end
 
+    def self.mark_as_delivered!(apple_publisher)
+      apple_publisher.episodes_to_sync.each do |episode|
+        if episode.podcast_container&.needs_delivery? == false
+          episode.feeder_episode.apple_has_delivery!
+        end
+      end
+    end
```
svevang, line 85: This is follow up from #853 
When looking at existing episodes, we need to mark the episode as having delivery, based on the container, so that we don't churn/reupload for all the episodes in the feed.

app/models/apple/episode.rb
```diff
@@ -58,23 +58,22 @@ def self.get_episodes(api, episodes)
...
-      episodes_to_sync = episodes.filter { |ep| guid_to_apple_json[ep.guid].present? }
+      episodes_to_sync = episodes.filter { |ep| show.find_apple_episode_json_by_guid(ep.guid).present? }
 
       bridge_params = episodes_to_sync.map do |ep|
-        id = guid_to_apple_json[ep.guid]["id"]
-        guid = guid_to_apple_json[ep.guid]["attributes"]["guid"]
-        Episode.get_episode_bridge_params(api, id, guid)
+        apple_json = show.find_apple_episode_json_by_guid(ep.guid)
+        apple_episode_id = apple_json["id"]
+        Episode.get_episode_bridge_params(api, ep.feeder_id, apple_episode_id)
```
svevang, line 71: Apple ID / GUID changes

This uses the new show level interface for looking up guids. This will handle the case where the episode has the apple episode id set as the feed item guid (in the RSS).

If the episode's item guid matches an Apple episode id, then join to the that.

At the end we upsert the "SyncLog", and subsequently we reference the remote episode only via it's `id` property.

db/migrate/20231011213406_remove_apple_excluded_category.rb
```diff
@@ -0,0 +1,19 @@
+class RemoveAppleExcludedCategory < ActiveRecord::Migration[7.0]
+  def change
+    Apple::Config.all.each do |ac|
+      pub = ac.build_publisher
+
+      eps =
+        pub.show.podcast_episodes.filter do |ep|
+          next false unless ep.feeder_episode.categories.include?("apple-excluded")
+          ep.feeder_episode.update(categories: ep.feeder_episode.categories - ["apple-excluded"])
```
svevang, line 9: Apple ID / GUID changes

This fixes up the existing episodes having remote episode ids as their guids. This removes the excluded category so that they end up in the set of episodes to sync (The original fix for this bad data was to exclude the episodes from `Apple::Publisher#episodes_to_sync`).

PR #888, opened 2023-10-16
Podcast timings import

cavis:
This adds another import subclass, for pulling in episode timings.

The timings csv/tsv you upload or paste in should have a format like:

|Title|Episodes Pubdate             |The Guid|And The Midroll Breaks                       |
|-----|-----------------------------|--------|---------------------------------------------|
|Whatev|2023-09-28                   |this-is-a-guid-1|{415.32,1246.65}                             |
|Second Episode| 2023-09-14                  | some-other-guid-2|{612.53}                                     |
|The Third Episode| 2023-09-12                  |last-guid-3|{}                                           |

The order of the columns doesn't matter, and extra columns with other data will be ignored.  The code will auto-detect which column contains the guids and the timings.  And return validation errors if it can't find them.

Once the import passes validation and saves - we fire off an `EpisodeTimingsImport` for each row in the CSV.  These try to find an episode by `<guid>`, and slice that episode's `Content` with these midroll timestamps.


app/models/media_resource.rb
```diff
@@ -142,8 +142,13 @@ def media_url
...
   end
 
+  def original_ext
+    without_query = (original_url || "").split("?").first
+    File.extname(without_query || "")
+  end
+
   def media_url_for_base(base_published_url)
-    ext = File.extname(original_url || "")
+    ext = original_ext
```
cavis, line 151: Fixes a bug where query parameters were getting into `f.prxu.org` urls.

app/controllers/imports_controller.rb
```diff
@@ -5,7 +5,9 @@ class ImportsController < ApplicationController
   # GET /imports
   def index
     @imports = @podcast.podcast_imports.order(created_at: :asc)
-    @import = @podcast.podcast_imports.new
+    @import = @podcast.podcast_imports.new(import_params)
+    @import.clear_attribute_changes(%i[type])
```
radical-ube, line 9: is `clear_attribute_changes` just a way to make sure that `type` stays what it was input as from `import_params`? does this have to do with the `reverse_merge` default that occurs below?

cavis: It's cosmetic in this case - I was trying to get the field to not have an orange border when the page first loads.

app/models/concerns/import_utils.rb
```diff
@@ -14,9 +14,31 @@ class HttpError < StandardError
   IMPORTING = "importing".freeze
   COMPLETE = "complete".freeze
   ERROR = "error".freeze
+  NOT_FOUND = "not_found".freeze
+  BAD_TIMINGS = "bad_timings".freeze
+  NO_MEDIA = "no_media".freeze
```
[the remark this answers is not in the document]
radical-ube: what is included in `bad_timings`? like timings that go out of range? or negative, NaN, nil, etc?

app/models/episode.rb
```diff
@@ -227,11 +227,14 @@ def medium=(new_medium)
...
-          (uncut || build_uncut).tap do |u|
-            u.original_url = contents.first.url
+        if (c = contents.first)
+          build_uncut.tap do |u|
             u.file_size = contents.first.file_size
             u.duration = contents.first.duration
+
+            # use the feeder cdn url for older completed files
+            is_old = (Time.now - c.created_at) > 24.hours
+            u.original_url = (c.status_complete? && is_old) ? c.url : c.original_url
```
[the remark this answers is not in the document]
radical-ube: so basically "updating" the `original_url` if it's old? is this related to the "regenerate publicfeeds.net url" issue in the feed fixup PR?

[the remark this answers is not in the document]
cavis: This is a bit of a workaround for #874.

So you import the RSS, your episode has original_url = `http://some.where/my_file.mp3`.  But it has url = `https://f.prxu.org/123/<episode_guid>/<media_guid>.mp3`.

If you import your timings CSV _within 24 hours_ ... the timings will set its original_url = `.../my_file.mp3`.  And the filename in your final dovetail enclosure will be `my_file.mp3`.

But if you import your timings CSV after 24 hours, your episode will end up with a file_name like `123e4567-e89b-12d3-a456-426614174000.mp3`

app/models/imports/episode_timings_import.rb
```diff
@@ -0,0 +1,61 @@
...
+      return status_no_media! unless episode.uncut.present?
+
+      episode.uncut.ad_breaks = parse_timings
+      return status_bad_timings! unless episode.uncut.valid?
+
+      episode.segment_count = episode.uncut.segmentation.count
+      episode.save!
+
+      # slice new contents (if any) and re-copy
+      episode.uncut.slice_contents!
```
radical-ube, line 52: i kind of understood up to here. is this different from the `segmentation` on L48? or is this method the actual segmenting of the audio file into separate segments? and maybe the other one is just the segment markers?

cavis: That `slice_contents!` method takes the `uncut.segmentation` field/array and converts it into a bunch of new `Content` records.

So for instance, if you have `uncut.segmentation = [[1, 20], [23, 40]]` ... calling `slice_contents!` would create 2 `Content` records.  One sliced from second 1 to 20 of the Uncut's mp3.  And one sliced from 23 to 40 seconds.

app/views/imports/_form.html.erb
```diff
@@ -1,12 +1,35 @@
-<%= form_with(url: podcast_imports_path(import.podcast), model: [import.podcast, import.becomes(PodcastImport)], method: :post, html: {autocomplete: "off"}) do |form| %>
-  <div class="row my-4">
-    <div class="col-lg-8">
-      <%= render "form_main", imports: imports, podcast: podcast, form: form %>
+<%= form_with(model: import.becomes(PodcastImport), method: :post, html: {autocomplete: "off"}) do |form| %>
+  <%# hacky - but we need to cast back into the child class %>
+  <% form.object = import.becomes(import.class) %>
```
radical-ube, line 3: what is the purpose of casting back into the child class?

app/views/imports/_form.html.erb
```diff
@@ -1,12 +1,35 @@
...
+      <%= turbo_frame_tag "import-settings" do %>
+        <div class="col-12 mb-4 prx-field-group">
+          <div class="form-floating" data-controller="dynamic-form">
+            <%= form.select :type, t("helpers.label.podcast_import.types").to_a.map(&:reverse), {}, data: {action: "dynamic-form#change"} %>
+            <%= form.label :type, required: true %>
+            <a hidden href="<%= request.fullpath %>" data-dynamic-form-target="link"></a>
+          </div>
+        </div>
+
+        <%# if form.object.is_a? PodcastRssImport %>
```
radical-ube, line 20: leftover comment?

app/views/imports/_form.html.erb
cavis: Okay, I had to try removing this to remember.

So we need to `.becomes(PodcastImport)` the parent class in the `form_with`... so that our field names are `podcast_import[url]` instead of `podcast_rss_import[url]`.

But then inside that, I needed to cast it back.  Otherwise calling something like `import.timings` would error - because that method doesn't exist on a `PodcastImport`.  Only on a `PodcastTimingsImport`.

I'll update the comment to be better.

app/views/imports/_form.html.erb
cavis: Ah, good catch.  Will remove.

test/models/imports/episode_timings_import_test.rb
```diff
@@ -0,0 +1,127 @@
+require "test_helper"
+
+describe EpisodeTimingsImport do
+  let(:import) { EpisodeTimingsImport.new }
+
+  describe ".parse_timings" do
+    it "parses timings" do
+      assert_equal [1.23], EpisodeTimingsImport.parse_timings("1.23")
+      assert_equal [1.23, 4.56], EpisodeTimingsImport.parse_timings(" 1.23 , 4.56")
+      assert_equal [1.23, 4.56, 6], EpisodeTimingsImport.parse_timings(" 1.23 , 4.56,6")
```
radical-ube, line 10: what makes the spacing behave this way? (the leading whitespace, one comma has whitespace on both sides while the other doesn't)

radical-ube (approved):
+1 reviewed to best of my ability. works as intended.

test/models/imports/episode_timings_import_test.rb
cavis: This test is just proving that the code ignores/strips whitespace.  It's a bit of a nonsensical string here - as if the user had done a bad job typing into excel.

PR #892, opened 2023-10-17
Fix/readonly feeds and episodes

### LINKED ISSUE #870: Readonly Feeds
Couple bugs for readonly users:

- Feed "add a feed" button still shows up
- Feed "Add Token" button still shows up on private feeds
- Feed "Delete Token" icon/button still shows up on private feeds
- Episode Categories adfree buttons show up
- Episode Categories "remove" X icons show up

radical-ube:
resolves #870 

All tasks from ticket are addressed except for the last one. It seems like this slim select field doesn't visually `disable` even though it's functionally `disabled` via policy. Not sure how to address.

### DIFF
--- a/app/views/episodes/_form_tags.html.erb
+++ b/app/views/episodes/_form_tags.html.erb
@@ -12,10 +12,12 @@
       </div>
     </div>
 
-    <div class="card-footer" data-controller="tooltip">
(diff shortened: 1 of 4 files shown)


app/views/episodes/_form_tags.html.erb
```diff
@@ -6,16 +6,18 @@
 
     <div class="card-body">
       <div class="form-floating input-group">
-        <%= form.tag_select :categories, episode.categories, {}, data: {allow_new: true, allow_clear: true, category_target: "field"} %>
+        <%= form.tag_select :categories, episode.categories, {}, data: {allow_new: policy(episode).update?, allow_clear: policy(episode).update?, category_target: "field"} %>
```
radical-ube, line 9: technically didn't need to also add the policy to the `allow_new` param, but it seemed like an opportunity to keep it consistent.

cavis (approved):
:+4: looks and works good!

PR #895, opened 2023-10-20
Feed fixups

### LINKED ISSUE #840: Keep publicfeeds.net urls in-sync with file_name/slug
If you create a new feed and provide no public url for it, we generate a `publicfeeds.net` url for you by default.

But - if you change the file_name or slug of that feed, we need to re-generate that `publicfeeds.net` url.  Otherwise it'll be stuck pointing to an old/stale file in S3, which no longer gets updated.

### LINKED ISSUE #852: Cleanup renamed feed rss in S3
Right now, if you change a Feed file_name or slug, we just leave the old file in S3.  But it no longer gets updates.

Would be better if we removed the old RSS file, to avoid confusion.  Maybe fire off a job after_save, to delete it?  Or just SNS Porter to do it?  Any gotchas if we did that?

This is one case where we "lose track" of a file in S3.  Unlike images/audio, which have paranoid db models so we always know where they are.

### LINKED ISSUE #883: Public feed link wrong for non-default feeds
The "open in new tab" link for non-default public feeds is wrong.  It opens the default feed instead of the value in the input.

image

cavis:
Fixes #840.
Fixes #852.
Fixes #883.

Fixes all the things.

### DIFF
--- a/app/models/feed.rb
+++ b/app/models/feed.rb
@@ -82,12 +82,12 @@ class Feed < ApplicationRecord
     elsif ENV["PUBLIC_FEEDS_URL_PREFIX"].present? && podcast.present?
       public_feeds_url = "#{ENV["PUBLIC_FEEDS_URL_PREFIX"]}/#{path}"
 
-      # hacky, but only default publicfeeds urls on newer records
-      if url.blank? && (new_record? || created_at >= "2023-10-01")
-        self.url = public_feeds_url
-      elsif url.present? && url.include?(ENV["PUBLIC_FEEDS_URL_PREFIX"])
-        self.url = public_feeds_url
-      end
+      # if already publicfeeds... keep file_name/slug changes in sync
+      self.url = public_feeds_url if url.present? && url.include?(ENV["PUBLIC_FEEDS_URL_PREFIX"])
+
+      # otherwise, just default back to publicfeeds when blank
+      # TODO: after https://github.com/PRX/feeder.prx.org/issues/896 remove the date condition
+      self.url = public_feeds_url if url.blank? && (new_record? || created_at >= "2023-10-01")
     end
   end
 
--- a/app/models/feed.rb
+++ b/app/models/feed.rb
@@ -219,6 +219,12 @@ class Feed < ApplicationRecord
   def copy_media(force = false)
     feed_images.each { |i| i.copy_media(force) }
     itunes_images.each { |i| i.copy_media(force) }
+
+    # remove old feed rss
+    if !previously_new_record? && (slug_previously_changed? || file_name_previously_changed?)
+      old_path = [podcast.path, slug_previously_was, file_name_previously_was].compact.join("/")
+      UnlinkJob.perform_later(old_path)
+    end
   end
 
   def ready_feed_image
--- a/app/models/feed.rb
+++ b/app/models/feed.rb
@@ -80,7 +80,7 @@ class Feed < ApplicationRecord
     if private?
       self.url = nil
     elsif ENV["PUBLIC_FEEDS_URL_PREFIX"].present? && podcast.present?
-      public_feeds_url = "#{ENV["PUBLIC_FEEDS_URL_PREFIX"]}/#{podcast.path}/#{published_path}"
+      public_feeds_url = "#{ENV["PUBLIC_FEEDS_URL_PREFIX"]}/#{path}"
 
       # hacky, but only default publicfeeds urls on newer records
       if url.blank? && (new_record? || created_at >= "2023-10-01")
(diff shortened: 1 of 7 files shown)


app/models/feed.rb
```diff
@@ -78,15 +77,20 @@ def sanitize_text
...
 
   def set_public_feeds_url
-    if public? && url.blank? && ENV["PUBLIC_FEEDS_URL_PREFIX"].present?
-      self.url = "#{ENV["PUBLIC_FEEDS_URL_PREFIX"]}/#{podcast.path}/#{published_path}"
+    if private?
+      self.url = nil
+    elsif ENV["PUBLIC_FEEDS_URL_PREFIX"].present? && podcast.present?
+      public_feeds_url = "#{ENV["PUBLIC_FEEDS_URL_PREFIX"]}/#{path}"
+
+      # hacky, but only default publicfeeds urls on newer records
```
cavis, line 85: Maybe this is a terrible idea.  But I've seen `f.prxu.org` urls being used as public feeds in the wild.  And I didn't want them to suddenly have an `<atom:link>` pointing somewhere else.

app/models/feed.rb
```diff
@@ -211,6 +219,12 @@ def mime_type
   def copy_media(force = false)
     feed_images.each { |i| i.copy_media(force) }
     itunes_images.each { |i| i.copy_media(force) }
+
+    # remove old feed rss
+    if !previously_new_record? && (slug_previously_changed? || file_name_previously_changed?)
```
cavis, line 224: Terrible idea #2 - after a controller saves a Feed and calls `.copy_media`, check if we just changed one of these fields.

Removing the old RSS file is nice, but not critical.  So I think this is good enough?

app/models/feed.rb
radical-ube: what does the conditional do here if the output is the same on either branch? is it a way to check these cases separately?

cavis: Hrmm... I guess that was a funny way to write that.  I was checking 2 cases, but the result is the same either way:

1. Is this a newer Feed (created this October or later)?  Then default to publicfeeds.net when blank.
2. Does this feed _already_ have a publicfeeds.net url?  Then keep updating it every time it's saved, _just in case_ the file_name or slug changed.

I'll refactor that a bit to clarify.

app/views/feeds/_form_distribution.html.erb
```diff
@@ -13,7 +13,7 @@
           <% data = {confirm_field_target: "field", confirm_with: with, confirm_delete: delete} %>
           <%= form.text_field :url, data: feed.url_was.present? ? data : {} %>
           <%= form.label :url %>
-          <%= field_link podcast.url if podcast.url.present? %>
+          <%= field_link feed.url if feed.url.present? %>
```
radical-ube, line 16: it does make sense for this to be the feed url, but was there a reason this used to be the podcast url?

cavis: This was the bug #883 - the `podcast.url` is one of those fields that's proxied to `podcast.default_feed.url`.  So this link would be correct _only_ if you're looking at the default feed.

(I probably originally copied/pasted this code from somewhere else, and forgot to change this line?)

radical-ube (changes requested):
just the one conditional clarification nitpick on the feed model, but otherwise this looks good.

app/models/feed.rb
```diff
@@ -78,13 +77,18 @@ def sanitize_text
...
+      public_feeds_url = "#{ENV["PUBLIC_FEEDS_URL_PREFIX"]}/#{path}"
 
-  def remove_url
-    self.url = nil
+      # if already publicfeeds... keep file_name/slug changes in sync
+      self.url = public_feeds_url if url.present? && url.include?(ENV["PUBLIC_FEEDS_URL_PREFIX"])
+
+      # otherwise, just default back to publicfeeds when blank
+      # TODO: after https://github.com/PRX/feeder.prx.org/issues/896 remove the date condition
+      self.url = public_feeds_url if url.blank? && (new_record? || created_at >= "2023-10-01")
```
radical-ube, line 90: so the cases where `self.url` would not be changed are:

- new record, but the `url` has been filled in
- not new record, but it's not using the `PUBLIC_FEEDS_URL_PREFIX`

because we would want those `url`s to be whatever the user has set them to in those cases?
so number 2 is a case where `PUBLIC_FEEDS_URL_PREFIX` is present (L82) but is not used in the `url` (L86)?

cavis: Exactly - we want to push users to _either_ use their own domain (like `https://feed.songexploder.net/SongExploder`) ... or use the `publicfeeds.net` url we generate for them.  We _don't_ want them to use `f.prxu.org` as the public feed url.

radical-ube: but this logic doesn't change the url if they still chose to use `f.prxu.org`? like the option still needs to exist even though it's not encouraged?

radical-ube (approved):
reviewed to best ability. looks good to me!

PR #897, opened 2023-10-23
RSS import parity

### LINKED ISSUE #894: RSS import parity
When importing an RSS feed, I'm seeing that:

1. ~New Feed Url is only parsed if set in that source feed.  Is that right?~ yes
2. The `default_feed.url` ends up blank - due to some of this feedburner parsing code perhaps?  This causes us to have a different `atom:link` than the source feed.
3. We're not parsing itunes:authors from the channel/items
4. Do we need to strip all these `podcastchoices.com/adchoices` links from descriptions?

cavis:
Closes #894

### DIFF
--- a/app/models/concerns/import_utils.rb
+++ b/app/models/concerns/import_utils.rb
@@ -93,6 +93,7 @@ module ImportUtils
   def clean_text(text)
     return nil if text.blank?
     result = remove_feedburner_tracker(text)
+    result = remove_podcastchoices_link(result)
     sanitize_html(result)
   end
 
--- a/app/models/concerns/import_utils.rb
+++ b/app/models/concerns/import_utils.rb
@@ -81,7 +81,7 @@ module ImportUtils
       end
     end
 
-    {name: name, email: email}
+    {name: name, email: email}.with_indifferent_access
   end
 
   def clean_string(str)
--- a/app/models/concerns/import_utils.rb
+++ b/app/models/concerns/import_utils.rb
@@ -102,6 +102,12 @@ module ImportUtils
     str.sub(regex, "").strip
   end
 
+  def remove_podcastchoices_link(str)
+    if str.present?
+      str.sub(/(<p>)?learn more about your ad choices.+podcastchoices.com\/adchoices\.?(<\/a>)?(<\/p>)?/i, "").strip
+    end
+  end
+
   def sanitize_html(text)
     return nil if text.blank?
     sanitizer = Rails::Html::WhiteListSanitizer.new
--- a/app/models/imports/podcast_rss_import.rb
+++ b/app/models/imports/podcast_rss_import.rb
@@ -145,8 +145,7 @@ class PodcastRssImport < PodcastImport
     podcast_attributes[:explicit] = explicit(feed.itunes_explicit, "false")
     podcast_attributes[:new_feed_url] = clean_string(feed.itunes_new_feed_url)
     podcast_attributes[:enclosure_prefix] ||= enclosure_prefix(feed.entries.first)
-    podcast_attributes[:feedburner_url] ||= feedburner_url(feed.feedburner_name)
-    podcast_attributes[:url] ||= feedburner_url(feed.feedburner_name)
+    podcast_attributes[:url] ||= clean_string(feed.feed_url)
 
     podcast_attributes[:author] = person(feed.itunes_author)
     podcast_attributes[:managing_editor] = person(feed.managing_editor)
@@ -203,10 +202,6 @@ class PodcastRssImport < PodcastImport
       prefix += blubrry[1]
     end
     prefix
-  end
-
-  def feedburner_url(fb_name)
-    fb_name ? "https://feeds.feedburner.com/#{clean_string(fb_name)}" : nil
   end
 
   def owner(itunes_owners)
--- a/app/models/imports/podcast_rss_import.rb
+++ b/app/models/imports/podcast_rss_import.rb
@@ -185,23 +185,22 @@ class PodcastRssImport < PodcastImport
     podcast
   end
 
-  def enclosure_prefix(podcast_item)
-    prefix = ""
-    link = [podcast_item.feedburner_orig_enclosure_link,
-      podcast_item.enclosure.try(:url),
-      podcast_item.media_contents.first.try(:url)].find do |url|
-      url.try(:match, /podtrac/) || url.try(:match, /blubrry/)
+  def enclosure_prefix(item)
+    redirectors = [
+      /\/(www|dts)\.podtrac\.com(\/pts)?\/redirect\.mp3\//,
+      /\/media\.blubrry\.com\/[^\/]+\//,
+      /\/chrt\.fm\/track\/[^\/]+\//,
+      /\/chtbl\.com\/track\/[^\/]+\//,
+      /\/pdst\.fm\/e\//
+    ]
+
+    urls = [item.feedburner_orig_enclosure_link, item.enclosure.try(:url), item.media_contents.first.try(:url)]
+    url = urls.compact.find { |u| redirectors.any? { |r| u.match?(r) } }
+
+    if url.present?
+      end_index = redirectors.map { |r| url.index(r) + url[r].length if url.match?(r) }.compact.max
+      url[0...end_index]
     end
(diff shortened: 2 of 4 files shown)


app/models/concerns/import_utils.rb
```diff
@@ -102,6 +103,12 @@ def remove_feedburner_tracker(str)
     str.sub(regex, "").strip
   end
 
+  def remove_podcastchoices_link(str)
+    if str.present?
+      str.sub(/(<p>)?learn more about your ad choices.+podcastchoices.com\/adchoices\.?(<\/a>)?(<\/p>)?/i, "").strip
```
cavis, line 108: Bit magical, but this regex will strip off both the HTML and plain-text version of these sentences:

```
Learn more about your ad choices. Visit podcastchoices.com/adchoices

<p>Learn more about your ad choices. Visit <a href="https://podcastchoices.com/adchoices">podcastchoices.com/adchoices</a></p>
```

And there are tests!

app/models/concerns/import_utils.rb
```diff
@@ -81,7 +81,7 @@ def person(arg)
       end
     end
 
-    {name: name, email: email}
+    {name: name, email: email}.with_indifferent_access
```
cavis, line 84: Found two of these places, where the hash assembled in the importer used symbols as keys.  But the `author=` method on the model was assuming string keys.  This fixes setting the itunes-author name.

app/models/imports/podcast_rss_import.rb
```diff
@@ -145,8 +145,7 @@ def build_podcast_attributes
     podcast_attributes[:explicit] = explicit(feed.itunes_explicit, "false")
     podcast_attributes[:new_feed_url] = clean_string(feed.itunes_new_feed_url)
     podcast_attributes[:enclosure_prefix] ||= enclosure_prefix(feed.entries.first)
-    podcast_attributes[:feedburner_url] ||= feedburner_url(feed.feedburner_name)
-    podcast_attributes[:url] ||= feedburner_url(feed.feedburner_name)
+    podcast_attributes[:url] ||= clean_string(feed.feed_url)
```
cavis, line 148: I just nuked the "feedburner name" parsing code, which looks like it was meant to strip the slug from feedburner urls, and use it as the `podcast.path` in our system.  (We deprecated those string paths, and just use the podcast ID in the S3 paths now).

And we now parse/pull the "self" link href into this feed_url property (see the feedjira parser file).

app/models/imports/podcast_rss_import.rb
```diff
@@ -186,32 +185,27 @@ def create_or_update_podcast!
...
+    ]
+
+    urls = [item.feedburner_orig_enclosure_link, item.enclosure.try(:url), item.media_contents.first.try(:url)]
+    url = urls.compact.find { |u| redirectors.any? { |r| u.match?(r) } }
 
-  def feedburner_url(fb_name)
-    fb_name ? "https://feeds.feedburner.com/#{clean_string(fb_name)}" : nil
+    if url.present?
+      end_index = redirectors.map { |r| url.index(r) + url[r].length if url.match?(r) }.compact.max
+      url[0...end_index]
```
cavis, line 202: Refactored to support more redirect prefixes.

This code (1) looks for an enclosure/contents/etc url with any redirect prefix in it.  And then (2) returns the entire redirect chain (in case multiple), minus the final Mp3-hosting domain.

lib/feedjira/parser/podcast.rb
```diff
@@ -50,6 +48,10 @@ class Podcast
       elements :link, as: :hubs, value: :href, with: {rel: "hub"}
       elements :"atom10:link", as: :hubs, value: :href, with: {rel: "hub"}
 
+      element :link, as: :feed_url, value: :href, with: {rel: "self"}
+      element :"atom10:link", as: :feed_url, value: :href, with: {rel: "self"}
+      element :"atom:link", as: :feed_url, value: :href, with: {rel: "self"}
```
cavis, line 53: That `attr_accessor` was apparently doing nothing?  I pulled it down here (copying some FeedJira other parsers) to look for any "self" link in the RSS.

app/models/concerns/import_utils.rb
radical-ube: where does this text come from? it seems specific.

app/models/imports/podcast_rss_import.rb
radical-ube: if i'm reading correctly, (1) is handled by L198, and (2) is L201-202? is the `url` mutated on L202 with that range method?

lib/feedjira/parser/podcast.rb
radical-ube: what is the FeedJira library and what do these `elements` do?

test/models/imports/podcast_rss_import_test.rb
```diff
@@ -134,23 +144,23 @@
...
     let(:sample_link3) do
-      "http://media.blubrry.com/some_name/www.podtrac.com/pts/redirect.mp3/blah"
+      "https://pts.podtrac.com/redirect.mp3/pdst.fm/e/chtbl.com/track/7E7E1F/blah"
     end
 
     it "can make a good guess for an enclosure prefix" do
       item = feed.entries.first
-      _(importer.enclosure_prefix(item)).must_equal "https://dts.podtrac.com/redirect" \
-        ".mp3/media.blubrry.com/transistor/"
+      _(importer.enclosure_prefix(item)).must_equal "https://dts.podtrac.com/redirect.mp3/media.blubrry.com/transistor/"
```
radical-ube, line 152: i think i'm still not sure where this prefix comes from. Transistor is the podcast name, and so it's guessing the enclosure by using the podcast title?

radical-ube: i guess the prefix is everything in the sample links up to `redirect.mp3`?

radical-ube: and then what follows is a redirect url?

app/models/concerns/import_utils.rb
cavis: It's on all megaphone generated RSS (apparently).  And that url has very megaphone/spotify specific language.

app/models/imports/podcast_rss_import.rb
cavis: yep, exactly

lib/feedjira/parser/podcast.rb
cavis: FeedJira is an RSS parsing gem - and these `element` / `elements` are part of its [syntax](https://github.com/feedjira/feedjira#adding-attributes-to-only-one-class).

I sort of pieced these together from the other parts of this parser file, and looking at how [the builtin Atom namespace parser](https://github.com/feedjira/feedjira/blob/main/lib/feedjira/parser/atom.rb#L13) works.

test/models/imports/podcast_rss_import_test.rb
cavis: These "prefixes" vary a bit by host - some of them are longer/shorter than others.  But when you request them, you get a bunch of redirects - here's a more realistic example:

```
GET https://pts.podtrac.com/redirect.mp3/pdst.fm/e/chtbl.com/track/7E7E1F/dovetail.prxu.org/123/abc/file.mp3
302 -> https://pdst.fm/e/chtbl.com/track/7E7E1F/dovetail.prxu.org/123/abc/file.mp3
302 -> https://chtbl.com/track/7E7E1F/dovetail.prxu.org/123/abc/file.mp3
302 -> https://dovetail.prxu.org/123/abc/file.mp3
302 -> https://dovetail-cdn.prxu.org/123/abc/def/file.mp3
200 -> finally we get to download the file!
```

Along each of those, the redirect hosts are tracking something about the request.  Until we finally get to dovetail.

Since Feeder only knows the `dovetail.prxu.org` URL, we need to save the rest of the chain in the `feed.enclosure_prefix` field as:

```
https://pts.podtrac.com/redirect.mp3/pdst.fm/e/chtbl.com/track/7E7E1F/
```

So we can add that to the beginning of every dovetail mp3 url in the feed.

radical-ube (approved):
looks good to me.

Issue #793, opened 2023-08-18
Set up uniqueness constraint on Apple::Config's podcast


svevang:
Until https://github.com/PRX/feeder.prx.org/issues/735 lands, it's not safe to have multiple Apple::Config instances for a single podcast.

Make sure this can't happen:

- Set up a `podcast_id` column on the Apple::Config that's populated in a rails hook
- Set up a uniqueness validation/constraint on that column

PR #898, opened 2023-10-23
Apple Config as singular to podcast

kookster:
- [x] #793


db/migrate/20231011213406_remove_apple_excluded_category.rb
```diff
@@ -14,6 +14,8 @@ def change
       pub.poll!(eps)
 
       Apple::Config.mark_as_delivered!(pub)
+    rescue => e
```
kookster, line 17: This older migration won't run with the new code and older db structure

app/models/apple/config.rb
```diff
@@ -61,19 +60,21 @@ def self.find_or_build_private_feed(podcast)
       )
     end
 
+    # this doesn't seem to be called anywhere?
+    # is this a helper for setup via console, with that stdin.get?
```
svevang, line 64: Yep, this is used to onboard new shows

Eventually, we can retrofit this as a programmatic pathway for connecting feeds to delegated delivery.

kookster:
the changes to the models now only, but I'm missing adding a unique constraint in the db.

app/models/apple/config.rb
```diff
@@ -61,19 +60,20 @@ def self.find_or_build_private_feed(podcast)
       )
     end
 
+    # TODO: this a helper for onboarding via console, retrofit when the UX catches up
     def self.build_apple_config(podcast, key)
-      if has_apple_config?(podcast)
+      if podcast.apple_config
```
svevang, line 65: Do we need this `has_apple_config?` method anymore?

app/models/apple/config.rb
```diff
@@ -103,27 +103,18 @@ def self.setup_delegated_delivery(podcast, key: nil, apple_config: nil, apple_sh
...
-      if podcast.feeds.map(&:apple_configs).flatten.compact.present?
-        errors.add(:podcast, "can only have one apple config")
-      end
-    end
-
     def feed_podcasts_match
       return unless public_feed.present? && private_feed.present?
 
-      if public_feed.podcast != private_feed.podcast
+      if (public_feed.podcast_id != podcast_id) || (private_feed.podcast_id != podcast_id)
```
svevang, line 109: Good to see that we're validating, the podcasts / feeds line up hierarchically under the podcast_id

app/models/apple/config.rb
```diff
@@ -103,27 +103,18 @@ def self.setup_delegated_delivery(podcast, key: nil, apple_config: nil, apple_sh
...
         errors.add(:public_feed, "must belong to the same podcast as the private feed")
       end
     end
 
     def publish_to_apple?
       return false unless key&.valid?
 
-      public_feed.publish_to_apple?(self)
+      public_feed.publish_to_apple?
```
svevang, line 117: So much cleaner!

app/models/apple/show.rb
```diff
@@ -31,6 +31,15 @@ def self.get_show(api, show_id)
       api.unwrap_response(resp)
     end
 
+    def self.for_podcast(podcast)
```
svevang, line 34: Not a big deal,  but the pattern up to now has been to build derived objects _from_ some other object. so `from_podcast(podcast)` here for the signature?

db/migrate/20231114155433_add_podcast_apple_config_unique_constraint.rb
```diff
@@ -0,0 +1,5 @@
+class AddPodcastAppleConfigUniqueConstraint < ActiveRecord::Migration[7.0]
+  def change
+    add_index :apple_configs, [:podcast_id], unique: true
```
svevang, line 3: And with rails validations on unique ids across public/private feeds then this transitively applies to the feeds as well `{:podcast_id, :public_feed_id}` etc

svevang (approved):
Looks good!

Just a couple minor nitpicks on method name and unused class method

app/models/apple/config.rb
kookster: nope! will remove

app/models/apple/show.rb
kookster: for sure!

PR #900, opened 2023-10-25
Multiple itunes categories

### LINKED ISSUE #890: Multiple itunes categories / subcategories
Noting that the RSS importer will create multiple itunes categories / subcategories.

And our RSS will render up to 3 itunes categories, and all of their subcategories.

But - as soon as you save the Podcast in the Feeder UI, it will delete all but 1 category and its first subcatagory.

We should do... better?

cavis:
Closes #890.

Allow setting multiple itunes categories/subcategories from the UI.

image

To make this work, I had to do a couple tricky things:

1. The Itunes Categories select is always sorted by the order you select them in.  So when you save, the top-most category in your RSS is the same as the leftmost one in this select.  NOT alphabetical.
2. The subcategories select updates every time you add a category.  To reflect all the subcategories under any of those categories.
3. Subcategories I just alphabetized.  We'll see if anyone cares how those are ordered, within the category.  I think it's most important to be able to pick which top-level category Apple sees.

### DIFF
--- a/app/assets/stylesheets/shared/form.scss
+++ b/app/assets/stylesheets/shared/form.scss
@@ -68,6 +68,9 @@
   .invalid-feedback {
     order: 10;
   }
+  input[type="hidden"] + .invalid-feedback {
+    display: none;
+  }
 }
 
 label[required="required"],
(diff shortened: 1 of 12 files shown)


brandonhundt (approved):
:shipit: Great.

PR #901, opened 2023-10-30
Save media versions

### LINKED ISSUE #858: Delegated delivery probes for media version
- Apple::Publisher checks for `Apple::Show#episodes`  having a changed media version.
- Mark the Episode as `Episode#apple_needs_delivery!`
- Update the media version and enclosure metadata to Podcast Container (via https://github.com/PRX/feeder.prx.org/blob/main/app/models/apple/podcast_container.rb#L285-L292)
- Send a request to DTR HEAD
- Poll for the DTR redirect with the expected version.
- Enter regular publish flow -- replace the media in the podcast container

svevang:
Closes #858 

This PR:

- Combines all the mutable state into a single "status" record `Episode#apple_episode_delivery_status`. This status record is derived from an append-only set `Episode#apple_episode_delivery_statuses`, so now we can have a complete record of each attempt to upload, what size and location metadata the CDN returned etc.
- Sets up a `source_media_version_id` that we keep alongside `source_url` and `source_size`.  This is initialized on reset and probe in the `Apple::PodcastContainer` class.
- Modifies `Apple::Episode#needs_delivery?` to include a `has_media_version?` test.  If the feeder episode media version that we've fetched in a prior publishing flow does not match the _current_ episode media version, then we must redeliver.  And of course `Apple::Episode#needs_delivery?` filters up to `Apple::Episode#synced_with_apple?` and so the episode end back up in `Apple::Publisher#episodes_to_sync`

Testing:
1) Turn off apple publishing in staging
2) Create and publish a new episode
3) coerce media_version_ids for the episode
4) dump the database
5) load and migrate
6) `Apple::Publisher#publish!`


db/schema.rb
```diff
@@ -147,8 +153,8 @@
     t.string "feedburner_orig_link"
     t.string "feedburner_orig_enclosure_link"
     t.boolean "is_perma_link"
-    t.datetime "source_updated_at", precision: nil
     t.string "keyword_xid"
+    t.datetime "source_updated_at", precision: nil
```
cavis, line 157: Weird this field keeps flipping around.  I'm guessing the order of the fields is different in staging vs prod?

app/models/apple/podcast_container.rb
```diff
@@ -34,34 +35,56 @@ def self.reset_source_file_metadata(episodes)
...
+      raise "Missing podcast container for episode" if episodes.map(&:podcast_container).any?(&:nil?)
+
+      wait_for(episodes, wait_interval: wait_interval, wait_timeout: wait_timeout) do |remaining_episodes|
+        containers = Apple::PodcastContainer.reset_source_file_metadata(remaining_episodes)
+        Rails.logger.info("Reset container source metadata", {reset_count: containers.length})
+
+        containers = Apple::PodcastContainer.probe_source_file_metadata(api, remaining_episodes)
+        Rails.logger.info("Updated container source metadata.", {count: containers.length})
+
+        finished = remaining_episodes.group_by(&:has_media_version?)
```
cavis, line 82: Is it possible for DTR to return a newer `media_version_id` than we were expecting?  I guess most likely, this call should always return the newest version.  Even if we started with an older one when this job started.

cavis (approved):
👍 looks good to me!

PR #903, opened 2023-10-31
Raise exception if there are any processing errors

svevang:
Fixes a case where the episode's podcast delivery files had processing errors, but because we were not using exception semantics (only logging errors),  we did not block the publishing of RSS.

This unites the exception handling / publishing blocking with the verification of the PodcastDeliveryFile processing status.


cavis (approved):
👍

PR #904, opened 2023-10-31
Clean up old delivery records upon retry

svevang:
Fallout from https://github.com/PRX/feeder.prx.org/pull/853

Fixes up a case where old delivery records (Apple::PodcastDelivery, Apple::PodcastDeliveryFile) are not cleaned up upon retry of delivery.

In #853 we switched from using the podcast delivery/delivery files as sentinels for determining when the audio needed reuploaded to using a flag that's independent of any remote state (primarily because the delivery records are ephemeral in Apple's systems).  But, various places in the codebase assume that there is only a single set of delivery related models at a time.

This PR sets up a top level guard in the publishing routine that:

- filters for episodes that need delivery
- removes deliveries and delivery files for those episodes to-be-uploaded


cavis (approved):
👍

PR #905, opened 2023-10-31
Publishing Pipeline expire! can handle deleted podcasts

svevang:
Handles the case where the publishing pipeline cannot expire podcasts marked as deleted.

### DIFF
--- a/app/models/publishing_pipeline_state.rb
+++ b/app/models/publishing_pipeline_state.rb
@@ -166,7 +166,7 @@ class PublishingPipelineState < ApplicationRecord
   end
 
   def self.expire_pipelines!
-    Podcast.where(id: expired_pipelines.select(:podcast_id)).each do |podcast|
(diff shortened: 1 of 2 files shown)


app/models/publishing_pipeline_state.rb
```diff
@@ -166,7 +166,7 @@ def self.expire!(podcast)
   end
 
   def self.expire_pipelines!
-    Podcast.where(id: expired_pipelines.select(:podcast_id)).each do |podcast|
+    Podcast.with_deleted.where(id: expired_pipelines.select(:podcast_id)).each do |podcast|
```
cavis, line 169: Has this happened?  We don't often delete podcasts in prod.

cavis (approved):
👍 looks good to me

app/models/publishing_pipeline_state.rb
svevang: In staging they pile up -- as part of CI

PR #906, opened 2023-10-31
Dropdate errors

### LINKED ISSUE #902: Allow episode "publish now"
If you go to a future-planned episode (has a `released_at` date set) and try to go directly to Status = Published, you get a validation error.  Because "published" episodes cannot have a drop date in the future.

This is problematic for some shows, which plan episodes ahead of time but then go "publish now" at some point on the drop day.

How to fix?

cavis:
Fixes #902.

Two changes:

1. The episode dropdate form was previously hidden (behind that edit-pencil) when you had a validation error.  So it was hard to see why the save failed.  This shows the form automatically, if there's an error on the dropdate field.  (Try scheduling an episode with a past-dropdate to see).
2. IF the episode was draft but had a future-dropdate ... resets that to `Time.now`.  So changing the publishing status from Draft to Published should always "just work" now.

### DIFF
--- a/app/models/concerns/publishing_status.rb
+++ b/app/models/concerns/publishing_status.rb
@@ -34,11 +34,14 @@ module PublishingStatus
     @publishing_status = value
 
     if value == "draft"
+      self.released_at = published_at if published_at.present?
       self.published_at = nil
     elsif value == "scheduled"
       self.published_at = released_at
     elsif value == "published"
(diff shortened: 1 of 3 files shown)


brandonhundt (approved):
I've reviewed all use cases, and didn't get a single error. Onwards. :shipit:

PR #908, opened 2023-11-08
Auth API caching

cavis:
Some changes ahead of https://github.com/PRX/dovetail-router.prx.org/pull/260, so DTR can start using the auth endpoints for all its data.  Instead of the public endpoints.

Overview:

1. Enables redis caching on auth endpoints.  Hit some gotchas, but this appears to be working fine now.  Since none of the actions we're caching are specific to a user.  You can either read-private or you can't.  And the HAL collection cache is already doing a [digest of ids/timestamps](https://github.com/PRX/hal_api-rails/blob/master/lib/hal_api/paged_collection.rb#L44), so `/api/v1/authorization/episodes` is already going to vary by user.
2. Changes `FEEDS_TOKEN` env to a more generic `API_ADMIN_TOKENS` list-of-tokens env.  And enables read-only access to _all_ auth endpoints if you have that token.

Don't merge until the `API_ADMIN_TOKENS` has been setup in infrastructure.


app/controllers/api/auth/episodes_controller.rb
```diff
@@ -28,6 +28,6 @@ def sorted(res)
   end
 
   def resources_base
-    @resources_base ||= super.merge(authorization.token_auth_episodes)
+    authorization.token_auth_episodes
```
cavis, line 31: I ran into a bug, where memoizing this would break things in the auth endpoints once I enabled the "admin token".  Inside `hal_api-rails` it was checking for pagination - but since we memoized `resources_base` without pagination - it would error.

In any case - not memoizing doesn't appear to hurt anything?  I don't see any additional SQL being run.

app/controllers/api/authorizations_controller.rb
```diff
@@ -8,4 +8,8 @@ class Api::AuthorizationsController < Api::BaseController
   def resource
     authorization
   end
+
+  def show_cache_path
+    authorization.cache_key
```
cavis, line 13: I removed all the `count`s from the `/api/v1/authorization` JSON.  So now the only thing that varies is the user_id - so it's safe to cache based on this key.

app/controllers/api/base_controller.rb
```diff
@@ -33,7 +34,11 @@ def pundit_user
   end
 
   def authorization
-    Authorization.new(prx_auth_token) if prx_auth_token
+    if prx_auth_token
+      Authorization.new(prx_auth_token)
+    elsif api_admin_token?
+      Authorization.new(nil, true)
```
cavis, line 40: Kinda hacky - but added a new flavor of `Authorization` when you have one of those API admin tokens (like DTR does).

This grants you read-only access to the authorized API endpoints.  Though if you try to edit anything there, it's going to 500, since Pundit will assume you have a `prx_auth_token`.  (I'll open a ticket after this, to address that - but I don't really want to give these API admins write access anyways).

app/controllers/api/feeds_controller.rb
```diff
@@ -1,8 +1,9 @@
-class Api::FeedsController < ApplicationController
-  skip_before_action :authenticate!
-  before_action :authenticate_feeds_token!
+class Api::FeedsController < Api::BaseController
```
cavis, line 1: This controller (which returns _all_ "non default settings" Feeds in the system) isn't long for this world. 
 [PRX/dovetail-router](https://github.com/PRX/dovetail-router.prx.org/pull/260) is going to move to using the `auth/podcast/123/feeds` endpoints, which is more scalable.

I made some changes in here to consolidate the api_admin_token code... but this functions exactly the same as before.

app/controllers/concerns/api_authenticated.rb
```diff
@@ -7,15 +7,20 @@ module ApiAuthenticated
...
   def authenticate_user!
-    user_not_authorized unless prx_auth_token
+    user_not_authorized unless authenticated?
   end
 
-  def cache_show?
-    false
+  # don't bother calculating cache keys if user will be 401'd anyways
+  def index_cache_path
+    super if authenticated?
```
cavis, line 20: Here was another gotcha - some part of `hal_api-rails` was checking the `index_cache_path` _before_ we had determined if you're logged in or not.  This was firing off some SQL to get the items in the collection, which failed because `authorization` is null.

config/application.rb
```diff
@@ -56,6 +56,9 @@ class Application < Rails::Application
       config.session_store :feeder_active_record_store, key: "_feeder_session"
     end
 
+    # Version the cache by the application version
+    ENV["RAILS_APP_VERSION"] = VERSION
```
cavis, line 60: We need to start cycling our redis cache on deploys, when the representers have changed.  This adds the app version (`1.0.0` at the moment) to the redis key.

app/models/authorization.rb
```diff
@@ -20,38 +21,48 @@ def persisted?
   end
 
   def cache_key
-    token_key = OpenSSL::Digest::MD5.hexdigest([token.scopes, token.resources].map(&:to_s).join(""))
+    token_key =
+      if api_admin
+        "api-admin"
```
kookster, line 26: is this value used elsewhere? should it be a constant?

app/representers/api/auth/episode_representer.rb
```diff
@@ -1,6 +1,17 @@
...
+  collection :media,
+    decorator: Api::Auth::MediaResourceRepresenter,
+    class: MediaResource
+
+  collection :complete_media,
+    as: :ready_media,
+    decorator: Api::Auth::MediaResourceRepresenter,
+    class: MediaResource,
+    writeable: false,
+    if: ->(_o) { !media_ready? && complete_media? }
```
kookster, line 13: *not* `media_ready?`?
That seems wrong?

app/representers/api/media_resource_representer.rb
```diff
@@ -2,10 +2,14 @@ class Api::MediaResourceRepresenter < Roar::Decorator
...
+  property :href, readable: false
+  property :file_name
   property :mime_type, as: :type
   property :file_size, as: :size
   property :duration
   property :status, writeable: false
+
+  # TODO: deprecate
+  property :href
+  property :original_url, writeable: false
```
kookster, line 14: oh, I expected these to be removed, and only available in he `auth` version?
I guess they will be, but not yet in this PR.

app/representers/api/media_resource_representer.rb
```diff
@@ -2,10 +2,14 @@ class Api::MediaResourceRepresenter < Roar::Decorator
...
-  property :original_url, writeable: false
+  property :href, readable: false
+  property :file_name
   property :mime_type, as: :type
   property :file_size, as: :size
   property :duration
   property :status, writeable: false
+
+  # TODO: deprecate
+  property :href
```
kookster, line 13: Duplicate of line 5?

kookster (changes requested):
I saw a couple of things, like in the representers, that looked like they should change.

app/models/authorization.rb
cavis: Only used here, but yeah, probably good practice to make it a constant.  Will update!

app/representers/api/media_resource_representer.rb
kookster: This is minor, but probably worth fixing.

app/controllers/api/auth/episodes_controller.rb
cavis: Okay, I may have slightly mis-analyzed this... the error thrown was:

```
NoMethodError (undefined method `prev_page' for #<ActiveRecord::Relation ...
```

So that's the [CollectionPaging](https://github.com/PRX/hal_api-rails/blob/main/lib/hal_api/representer/collection_paging.rb#L15) representer trying to render a link.  AND importantly - this was only happening on the `auth/podcasts#index`, not on this `auth/episodes#index`.  So, it's related to how podcasts was defining `@podcasts ||=` whereas this defines `@resources_base ||=`.

I _think_ what was happening was, [this logic](https://github.com/PRX/hal_api-rails/blob/main/lib/hal_api/controller/resources.rb#L62) was grabbing the memoized `@podcasts` variable, which didn't have paging on it.  So I _should_ be able to fix this by just changing the podcasts controller to `@resources_base ||=`.

That make sense?

app/representers/api/media_resource_representer.rb
cavis: Ah, when I was initially going to remove href from the public API, I left it as writeable above.  So you can PUT media via the public endpoint, but cannot read the href.

Then I figured out we need some DTR changes (move it to use the auth endpoints) before we can do this.  So I added it back again.  I'll comment out line 5 to clarify.

app/representers/api/media_resource_representer.rb
cavis: Yep, this PR doesn't change the API yet.  But it enables us to deploy https://github.com/PRX/dovetail-router.prx.org/pull/260, moving DTR to use all auth endpoints.

After that DTR change, we can come back and deprecate the href here.

app/controllers/api/auth/episodes_controller.rb
cavis: Update: memoization happens in other places.  This seems to be related to a combination of memoizing that model name variable, having index caching enabled, and the index cache path manually constructing the filtered/paged/sorted query.

In any case, the best path seems to be _not_ memoizing here.

cavis:
Okay @kookster - think I addressed all of that?

kookster (approved):
lgtm!
lots we talked about, looks good!

PR #909, opened 2023-11-08
Filter auth feeds by authorized podcasts

cavis:
Feeds were not being filtered correctly.

### DIFF
--- a/app/models/authorization.rb
+++ b/app/models/authorization.rb
@@ -40,12 +40,20 @@ class Authorization
     end
   end
 
+  def token_auth_feeds
+    if token.globally_authorized?("read-private")
+      Feed.with_deleted.all
+    else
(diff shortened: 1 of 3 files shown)


app/models/authorization.rb
```diff
@@ -40,12 +40,20 @@ def token_auth_podcasts
...
+    end
+  end
+
   # avoid joining podcasts here, as it breaks a bunch of other queries
   def token_auth_episodes
     if token.globally_authorized?("read-private")
       Episode.with_deleted.all
     else
-      Episode.where(podcast_id: token_auth_podcasts.pluck(:id))
+      Episode.where("podcast_id IN (SELECT id FROM podcasts WHERE prx_account_uri IN (?))", token_auth_account_uris)
```
cavis, line 56: Not entirely related - but it's slightly faster to do a subquery here, rather than 2 separate queries.

svevang (approved):

kookster (commented):
had a question about "deleted" feeds?

PR #918, opened 2023-11-14
Just overwrite instead of deleting

cavis:
This or https://github.com/PRX/Infrastructure/pull/726/files


cavis:
We merged that infra PR instead.

PR #919, opened 2023-11-14
Filter episodes by dropdate

### LINKED ISSUE #914: Episode API publishedAt filter
Add a filter to the public (and authorized?) episode APIs to filter by publishedAt.  `?before=` and `?after=` or something - i think we have similar in Augury.

Maybe also filter unpublished episodes by `releasedAt`, in the non-public API?

cavis:
Fixes #914.

Adds 2 params to the episodes APIs (both public and auth'd)...

- `?after=2023-11-14` filters by publishedAt `>=` (or releasedAt, for draft/scheduled episodes in the auth API)
- `?before=2023-11-14` filters by publishedAt `<` (exclusive)

Date strings are parsed with `DateTime.parse` - which is generally fairly permissive.

### DIFF
--- a/app/controllers/concerns/api_updated_since.rb
+++ b/app/controllers/concerns/api_updated_since.rb
@@ -9,7 +9,7 @@ module ApiUpdatedSince
     end
   end
 
-  def resources_base
+  def filtered(resources)
     if updated_since?
       super.where("updated_at >= ?", updated_since).order(updated_at: :asc)
     else
--- a/app/models/episode.rb
+++ b/app/models/episode.rb
@@ -67,7 +67,8 @@ class Episode < ApplicationRecord
   scope :draft, -> { where("episodes.published_at IS NULL") }
   scope :scheduled, -> { where("episodes.published_at IS NOT NULL AND episodes.published_at > now()") }
   scope :draft_or_scheduled, -> { draft.or(scheduled) }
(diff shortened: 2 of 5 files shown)


app/models/episode.rb
```diff
@@ -67,7 +67,8 @@ class Episode < ApplicationRecord
   scope :draft, -> { where("episodes.published_at IS NULL") }
   scope :scheduled, -> { where("episodes.published_at IS NOT NULL AND episodes.published_at > now()") }
   scope :draft_or_scheduled, -> { draft.or(scheduled) }
-  scope :after, ->(time) { where("#{DROP_DATE} > ?", time) }
+  scope :after, ->(time) { where("#{DROP_DATE} >= ?", time) }
```
cavis, line 70: This was only used in one place in the UI - so  shouldn't matter that I changed this from exclusive to inclusive.

app/controllers/concerns/api_updated_since.rb
```diff
@@ -9,7 +9,7 @@ module ApiUpdatedSince
     end
   end
 
-  def resources_base
+  def filtered(resources)
```
kookster, line 12: nice use of `filtered`, as intended!

kookster (approved):
lgtm!

PR #920, opened 2023-11-16
Timings import fixups

cavis:
🚧 WIP

### DIFF
--- a/app/models/imports/podcast_timings_import.rb
+++ b/app/models/imports/podcast_timings_import.rb
@@ -33,7 +33,7 @@ class PodcastTimingsImport < PodcastImport
   end
 
   def maximum_guid_length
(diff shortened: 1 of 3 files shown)


app/models/imports/podcast_timings_import.rb
```diff
@@ -33,7 +33,7 @@ def minimum_guid_length
   end
 
   def maximum_guid_length
-    @maximum_guid_length ||= [podcast&.episodes&.minimum("length(original_guid)"), default_guid_length].compact.max
+    @maximum_guid_length ||= [podcast&.episodes&.maximum("length(original_guid)"), default_guid_length].compact.max
```
kookster, line 36: ah, yeah, that's a bug

kookster (approved):
lgtm!

app/models/imports/podcast_timings_import.rb
cavis: My bad - test didn't catch this!

PR #921, opened 2023-11-17
Display nil explicit as "inherit"

### LINKED ISSUE #907: Set default "explicit" value to "inherit"
https://prx.slack.com/archives/G0FKLK2Q5/p1699026105188249

This is a little confusing. I've stumbed over this a few times in demos/walkthroughs. This field is technically empty and will inherit the series value but it looks as if its set to "explicit". Can we set the visual default to "inherit" to clear up the confusion?

Screenshot 2023-11-03 at 10 39 16 AM

cavis:
Fixes #907.

Adds a new getter/setter virtual field to the Episode model.

### DIFF
--- a/app/controllers/episodes_controller.rb
+++ b/app/controllers/episodes_controller.rb
@@ -141,7 +141,7 @@ class EpisodesController < ApplicationController
       :subtitle,
       :description,
       :production_notes,
-      :explicit,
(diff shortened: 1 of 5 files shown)


brandonhundt (approved):
Thank you. Works as described.

PR #924, opened 2023-11-21
Use nil embed player urls in the database

### LINKED ISSUE #922: Default landing urls getting out of sync
If you change the filename of a default feed, suddenly all of the embed player urls in the db are broken.  (The podcast "Homepage Link", and episode "Episode URL" fields).

Fixit fixit fixit

cavis:
Fixes #922.

- [x] Return embed player urls for nil `podcast.link` and `episode.url`
- [x] Make sure we don't save any `play.prx.org` urls to the db (also fixes up `episode.item_guid` for a similar issue)
- [x] Migration to nil out existing embed urls


app/helpers/embed_player_helper.rb
```diff
@@ -18,6 +18,10 @@ module EmbedPlayerHelper
   EMBED_PLAYER_SEASON = "se"
   EMBED_PLAYER_CATEGORY = "ct"
 
+  def embed_url?(value)
+    value.blank? || value.include?(ENV["PLAY_HOST"])
```
cavis, line 22: Maybe this `include?` is over-simplistic?  But hard to imagine a different and real url would have `play.prx.org` in it somewhere.

app/models/episode.rb
```diff
@@ -214,11 +217,19 @@ def explicit_content
   end
 
   def item_guid
-    original_guid || self.class.generate_item_guid(podcast_id, guid)
+    original_guid || generate_item_guid
   end
 
   def item_guid=(new_guid)
-    self.original_guid = new_guid.blank? ? nil : new_guid
+    self.original_guid = (new_guid.blank? || new_guid == generate_item_guid) ? nil : new_guid
```
cavis, line 224: Unrelated, but I realized that the UI was saving that generated `prx_123_abcd` value back to the database.  Changing the column from nil to non-nil.

Wasn't really hurting anything... but to be consistent, we should keep these nil.

db/migrate/20231121144340_nil_out_embed_player_urls.rb
```diff
@@ -0,0 +1,20 @@
+class NilOutEmbedPlayerUrls < ActiveRecord::Migration[7.0]
+  def up
+    Podcast.with_deleted.where("link LIKE ?", "%#{ENV["PLAY_HOST"]}%").update_all(link: nil)
+    Episode.with_deleted.where("url LIKE ?", "%#{ENV["PLAY_HOST"]}%").update_all(url: nil)
+  end
```
cavis, line 5: Realizing we're going to get some stale caches after this.  And we don't have a cache_version yet (that's in #908).  So I guess... note to self: go nuke the redis cache when this deploys.

PR #932, opened 2023-12-14
Include enclosures for replaced Uncut media

### LINKED ISSUE #915: Published episode with no enclosure
Ran into a published episode that _had_ media (single segment).  Then was converted to 3 segment (via either a timings import, or through the episode form UI).

Now it's still published, still "feed ready" (has an old/complete media version to serve), but the RSS has no enclosure.  That seems wrong.

cavis:
Fixes #915.  This tackles leaving the `<enclosure>` in the RSS while we're waiting for the user to segment an Uncut file.

This does not tackle RSS/API metadata (duration, content type, file size) matching the exact files Dovetail Router is going to serve you.

### DIFF
--- a/app/controllers/api/podcasts_controller.rb
+++ b/app/controllers/api/podcasts_controller.rb
@@ -13,7 +13,13 @@ class Api::PodcastsController < Api::BaseController
   end
 
   def show
-    super if visible?
+    if visible?
+      if request.format.rss?
+        render plain: FeedBuilder.new(show_resource).to_feed_xml
+      else
+        super
+      end
+    end
   end
 
   def visible?
--- a/app/models/concerns/episode_media.rb
+++ b/app/models/concerns/episode_media.rb
@@ -31,8 +31,8 @@ module EpisodeMedia
     complete_media.any?
   end
 
-  def no_media?
-    medium.nil? && segment_count.nil? && !media?
+  def media?
+    medium.present? || segment_count.present? || media.present?
   end
 
   def media
@@ -79,10 +79,6 @@ module EpisodeMedia
         self.medium = "video"
       end
     end
-  end
-
-  def media?
-    media.any?
   end
 
   def media_content_type(feed = nil)
--- a/app/models/enclosure_url_builder.rb
+++ b/app/models/enclosure_url_builder.rb
@@ -38,9 +38,9 @@ class EnclosureUrlBuilder
   end
 
   def podcast_episode_expansions(podcast, episode, feed)
-    media = episode.contents.first
+    media = episode.uncut || episode.contents.first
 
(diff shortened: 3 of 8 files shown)


app/controllers/api/podcasts_controller.rb
```diff
@@ -13,7 +13,13 @@ def included(relation)
   end
 
   def show
-    super if visible?
+    if visible?
+      if request.format.rss?
+        render plain: FeedBuilder.new(show_resource).to_feed_xml
```
cavis, line 18: For debugging.  Note that this only responds when you add a `/api/v1/podcasts/123.rss` extension (something about hal_api-rails doesn't let you request RSS here via `Accepts:`).  And this is cached.

The old (pre-Feeder-HTML) `/podcasts/123` endpoint worked a bit different.  But I couldn't quite get this to render that way, so ended up doing it manually.

app/models/concerns/episode_media.rb
```diff
@@ -31,8 +31,8 @@ def complete_media?
     complete_media.any?
   end
 
-  def no_media?
-    medium.nil? && segment_count.nil? && !media?
+  def media?
+    medium.present? || segment_count.present? || media.present?
```
cavis, line 35: The reason the `<enclosure>` wasn't appearing in the RSS - is because there was a difference in the `no_media?` and `media?` logic.

I consolidated to a single method here.  Episodes are always expected to have media if they have a medium, segment_count, or media (contents).

app/models/enclosure_url_builder.rb
```diff
@@ -38,9 +38,9 @@ def base_enclosure_url(podcast, episode, feed)
   end
 
   def podcast_episode_expansions(podcast, episode, feed)
-    media = episode.contents.first
+    media = episode.uncut || episode.contents.first
 
-    original_url = media.try(:original_url) || ""
+    original_url = media.try(:original_url) || "media"
```
cavis, line 43: Makes more sense to prefer the Uncut filename, over the Contents?  Also, when you replace your `episode.uncut`, that's going to soft-delete all the previous `episode.contents`.  And we'd end up with a blank filename here, which doesn't make a great enclosure url.

Left the default of `"media"` here just in case.  But in theory, the UI shouldn't allow you to save a published episode without replacing the old file.

app/models/episode.rb
```diff
@@ -406,7 +406,7 @@ def published_or_released_date
   end
 
   def validate_media_ready
-    return if published_at.blank? || no_media?
+    return unless published_at.present? && media?
```
cavis, line 409: Just inverting conditions in this file.

kookster (approved):
lgtm!
simpler even.

PR #939, opened 2023-12-15
updating help text and labels

### LINKED ISSUE #934: Clean up help text
A little help text clean up recommended by @farski from the Feeder/Publish parity spreadsheet:

- [x] Feed owner help text: Probably best to avoid the term iTunes
- [x] A lot of fields have help text that provides no additional information. For example, the Managing Editor Name says "set the managing editor name for this podcast". I don't think that helps a user who has already seen the input field and knows it's called Managing Editor Name. May be better to remove the (?) buttons when there's no significant help text available.
- [x] Title help text is the only one on the whole page I've found that is in the form of a question
- [x] Category help text, should probably call them "Apple Podcast categories" and link to https://podcasters.apple.com/support/1691-apple-podcasts-categories
- [x] Description help text says you can add images, but that doesn't seem to be true
- [x] Production notes, if this is just internal and never visible to listeners, I would explicitly say that in the help text
- [x] Explicit help text, replace iTunes with Apple Podcast
- [x] Inconsistency around using the terms Link and URL

brandonhundt:
Closes #934

- [x]  Feed owner help text: Probably best to avoid the term iTunes
- [x]  A lot of fields have help text that provides no additional information. For example, the Managing Editor Name says "set the managing editor name for this podcast". I don't think that helps a user who has already seen the input field and knows it's called Managing Editor Name. May be better to remove the (?) buttons when there's no significant help text available.
- [x]  Title help text is the only one on the whole page I've found that is in the form of a question
- [x]  Category help text, should probably call them "Apple Podcast categories" and link to https://podcasters.apple.com/support/1691-apple-podcasts-categories
- [x]  Description help text says you can add images, but that doesn't seem to be true
- [x]  Production notes, if this is just internal and never visible to listeners, I would explicitly say that in the help text
- [x]  Explicit help text, replace iTunes with Apple Podcast
- [x]  Inconsistency around using the terms Link and URL

I'm open to more suggestions. As things come up.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -388,7 +388,7 @@ en:
         segment_count: Number of Segments
         subtitle: Subtitle
         title: Title
-        url: Episode URL
+        url: Episode Link
       feed:
         audio_type: Format
         audio_bitrate: Bitrate
@@ -406,8 +406,8 @@ en:
         include_donation_url: Include Donation Form
         include_podcast_value: Include Micropayments Wallet
         include_tags: Include Categories
-        itunes_category: Category
-        itunes_subcategory: Sub-Category
+        itunes_category: Apple Podcast Category
+        itunes_subcategory: Apple Podcast Sub-Category
         new_feed_url: New Feed URL
         paid: Paid Ads
         private: Private (unlisted)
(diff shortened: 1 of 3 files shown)


config/locales/en.yml
```diff
@@ -388,7 +388,7 @@ en:
         segment_count: Number of Segments
         subtitle: Subtitle
         title: Title
-        url: Episode URL
+        url: Episode Link
```
farski, line 391: From what I can see, we use "URL" in many more places than "link", so I would probably change the few "links" to "URL", unless we want to go back and change everything.

Screenshot 2023-12-28 at 11 47 53 AM

config/locales/en.yml
```diff
@@ -709,7 +710,7 @@ en:
     form_auth:
       title: Feed Authorization
       help:
-        auth_tokens: Auth tokens help ensure your feed stays private. They must be appended to the feed URL for the audio to work. Remove them to revoke access to your feed.
+        auth_tokens: Auth tokens help ensure your feed stays private. They must be appended to the feed link for the audio to work. Remove them to revoke access to your feed.
```
farski, line 713: Same as above

farski (changes requested):

farski (changes requested):

farski (approved):

PR #953, opened 2024-01-18
Fix copyable links for unpublished episode players

### LINKED ISSUE #948: Bad embed player preview urls
The `Embeddable Player URL` field is showing the hardcoded values, when previewing an unpublished episode.

image

Don't do that!

Also (sort of related) ... can we preview the `Media URL` even before a file is uploaded?  I made a change awhile back to default enclosures to `file.mp3`... so I _think_ it's possible?

cavis:
Fixes #948.

There was a regression in 1466bc7e04ee3825ba22417869906c4ecb59fe43, where we stopped returning "preview" urls for the episode player.  Which wouldn't work until your episode hits the public RSS, but was the correct url.  We also lost the "hey, it looks like your episode may not be published" warning next to these links.

This PR fixes that, along with _always_ returning the Media URL.  Even if you haven't uploaded any files yet.  (We've gotten requests for this, and I recently defaulted the filename to just `/media` in these cases).

I need to QA this in staging some more (you need the actual feed-cdn to work for some pieces), but there should be no functional changes.  Summary:

1. _All_ player url/iframe generation is now handled server-side by feeder.  This involved adding the newest Theme / Accent query params, and porting the [generateEmbedHtml](https://github.com/PRX/play.poc.prx.org/blob/main/lib/generate/html/generateEmbedHtml.ts#L10) logic we needed from Play.
2. All form updates are now handled by Turbo (and [morphdom](https://github.com/patrick-steele-idem/morphdom), which we've been using more lately)
3. To re-fix the iframe flickering as you change styles (Turbo was re-rendering it) ... I added `data-morph=false` to the Iframe container, so it wouldn't get modified.  And the scope of the `embed_preview_controller.js` is now only to modify the already on page iframe/wrapper styles and relay configuration changes to the iframe via `postMessage`.  There is no round-trip of params coming back.

There does seem to be another regression, where even via the `play.prx.org/preview` path, the iframe is still showing the incorrect copyable iframe/url links.  But we'll have to tackle that over there.


app/controllers/episode_player_controller.rb
```diff
@@ -5,10 +5,14 @@ def show
...
-    @player_options[:episode_guid] = Episode.find_by_guid!(params[:episode_id])
+    @player_options = episode_player_params
+  end
+
+  private
+
+  def episode_player_params
+    nilify params
+      .permit(:embed_player_type, :max_width, :embed_player_theme, :accent_color)
+      .with_defaults(EmbedPlayerHelper::DEFAULT_OPTIONS)
```
cavis, line 16: So now, Turbo will GET this action with `?embed_player_type=card`, we regenerate the url/iframe, and re-render the page.

app/helpers/embed_player_helper.rb
```diff
@@ -111,19 +92,83 @@ def embed_player_category_options(podcast, selected)
...
 
-  def embed_player_all_episodes_options(selected)
-    opts = %w[all number].map { |v| [t("helpers.label.podcast_player.episodes_options.#{v}"), v] }
-    options_for_select(opts, selected)
-  end
-
   private
 
-  def embed_params(params, preview = false)
+  def embed_params(options_with_defaults = {})
```
cavis, line 97: This method was confusingly named before, so I hijacked it to generate the embed query params.  The above methods (`_episode_url` / `_podcast_url`) call through to this.

app/helpers/embed_player_helper.rb
```diff
@@ -111,19 +92,83 @@ def embed_player_category_options(podcast, selected)
...
+
+    # calculate height for playlists
+    height =
+      if options[:episode_number].to_i.between?(1, 5)
+        HEIGHT_BASE + HEIGHT_PLAYLIST_HEADER + HEIGHT_PLAYLIST_ROW * options[:episode_number].to_i
+      elsif options[:playlist]
+        (HEIGHT_BASE + HEIGHT_PLAYLIST_HEADER + HEIGHT_PLAYLIST_ROW * 5.5).round
+      else
+        HEIGHT_BASE
+      end
```
cavis, line 139: This method gets in the weeds a bit.  But it's ported from the embed player repo.  And there are advantages to being able to generate these completely in ruby (return API results, programmatically generate for a bunch of episodes, etc).

app/views/episode_player/_form.html.erb
```diff
@@ -1,27 +1,39 @@
...
   <div class="col-8">
+    <a hidden href="<%= request.fullpath %>" data-dynamic-form-target="link"></a>
 
     <section class="container-fluid mb-5">
       <header>
         <h3 class="fw-bold"><%= t(".title.preview") %></h3>
       </header>
 
-      <%= render "form_preview", player_options: @player_options %>
+      <%= render "form_preview", embed_html: embed_player_episode_iframe(episode, player_options, true) %>
```
cavis, line 10: And here's the trick - the `form_preview`s pass true at the end of this method, indicating they're _previewing_.  The below copyable form fields are all using the final non-preview urls.

rpeterman-gp (approved):
It would be nice if by default if the user didn't select a specific config value, that those params would default to `undefined` or `nil` and not get added to the copied URL/HTML. That maybe its own ticket. This PR looks good. :shipit:

app/helpers/embed_player_helper.rb
cavis: That's not a bad idea.  That would get us away from having users copy/paste iframes with hardcoded styles right?  So we could say, change the heights of things in the play repo.

app/controllers/episode_player_controller.rb
cavis: Yep - started with all serve side rendering, then added that `data-morph=false` thing to prevent turbo from clobbering the iframe.  And delegate those iframe updates to the stimulus controller.

app/helpers/embed_player_helper.rb
```diff
@@ -17,22 +17,38 @@ module EmbedPlayerHelper
   EMBED_PLAYER_PLAYLIST = "sp"
   EMBED_PLAYER_SEASON = "se"
   EMBED_PLAYER_CATEGORY = "ct"
+  EMBED_PLAYER_THEME = "th"
+  EMBED_PLAYER_ACCENT_COLOR = "ac"
+
+  DEFAULT_OPTIONS = {
```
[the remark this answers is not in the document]
cavis: Hrmm... might be possible.  But at the moment, those <select>s and the color picker need a default.

FWIW, these don't get added to the play.prx.org url.  If your options hash has one of these values, `def embed_params` strips it out before assembling the query params.

app/javascript/controllers/embed_preview_controller.js
```diff
@@ -1,141 +1,52 @@
...
+      const iframe = this.wrapperTarget.querySelector("iframe")
+      const tpl = document.createElement("template")
+      tpl.innerHTML = value
+
+      if (tpl.content.firstChild.tagName === "IFRAME") {
+        this.updateWrapper(wrapper, null)
+        this.updateIframe(iframe, tpl.content.firstChild)
+      } else {
+        this.updateWrapper(wrapper, tpl.content.firstChild)
+        this.updateIframe(iframe, tpl.content.firstChild.firstChild)
```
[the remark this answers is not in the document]
cavis: Yeah, the idea here is the string `embedHtmlValue` changes, and its this controllers responsibility to make those changes to the `wrapperTarget` div containing the iframe.

It seemed easiest to turn the string into an Element, and interrogate that for what styles/attributes we should change.

Yeah, there could be a cleaner way to do this?  I thought maybe I could call [morphdom](https://github.com/patrick-steele-idem/morphdom) manually here, and it would basically do all this for me?  But couldn't quite figure it out.

PR #954, opened 2024-01-26
Feat/new podcast slack notifications

### LINKED ISSUE #917: Slack notifications for new podcasts
from Paloma:

> Hey team, when a podcast is created/imported using Publish, we get a notification in the [#support-data-monitoring](https://prx.slack.com/archives/CE9E6SSBW) channel. Can we get the same when a show is setup/imported in feeder?

radical-ube:
resolves #917


app/controllers/podcasts_controller.rb
```diff
@@ -130,4 +132,17 @@ def podcast_params
...
+    title = "<#{podcast_url(podcast.id)}|#{sub_escapes(podcast.title)}>"
+    intro = "A new podcast has been created: #{title}"
+    id = "ID: #{podcast.id}"
+    planner = "<#{podcast_planner_url(podcast.id)}|Schedule episodes>"
+    imports = "<#{podcast_imports_url(podcast.id)}|Start a new import>"
+    [intro, id, planner, imports].join("\n")
+  end
+
+  def sub_escapes(text)
+    text.gsub(/[&<>]/, "&" => "&amp;", "<" => "&lt;", ">" => "&gt;")
```
radical-ube, line 146: these 3 were the only escape symbols that the Slack docs said needed to be substituted. do i need to prepare or consider any other formatting issues?

app/controllers/podcasts_controller.rb
```diff
@@ -130,4 +132,17 @@ def podcast_params
...
     )
   end
+
+  def compose_message(podcast)
+    title = "<#{podcast_url(podcast.id)}|#{sub_escapes(podcast.title)}>"
+    intro = "A new podcast has been created: #{title}"
+    id = "ID: #{podcast.id}"
+    planner = "<#{podcast_planner_url(podcast.id)}|Schedule episodes>"
+    imports = "<#{podcast_imports_url(podcast.id)}|Start a new import>"
+    [intro, id, planner, imports].join("\n")
```
radical-ube, line 142: wasn't sure what all was going to be helpful for distro, so i included the planner and the importer links if i were guessing where someone might need to go next.

env-example
```diff
@@ -70,3 +70,7 @@ APPLE_PROVIDER_ID=
 APPLE_KEY_ID=
 APPLE_KEY_PEM_B64=
 APPLE_API_BRIDGE_URL=
+
+# slack notifications
+SLACK_SNS_TOPIC=
```
radical-ube, line 75: when it comes time to add these variables in to stag/prod, i forget how that is done?

lib/slack_helper.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  def send_slack(message, options = {})
+    if SlackHelper.slack_sns_arn && SlackHelper.slack_channel
+      SlackHelper.slack_sns_publish(slack_default_options.merge(options).merge(text: message))
+    end
+  end
+
+  def slack_default_options
+    {
+      channel: SlackHelper.slack_channel || "#tech-dev-testing",
```
radical-ube, line 31: will change this channel to `#support-data-monitoring` once the PR is approved, but for now i used a test channel that i created.

app/controllers/podcasts_controller.rb
cavis: Nope - I think that'll do it.

lib/slack_helper.rb
cavis: +1, though I'd say you should copy the channel_id out of slack.  So things keep working if the channel gets renamed.

env-example
cavis: You'll need to wire through the SNS topic to the Infrastructure feeder.yml stack, and set the channel id.  Here's that same change for Augury - https://github.com/PRX/Infrastructure/commit/8bf3c3d787ee549cbc252906c573c2f554f4e630

app/controllers/podcasts_controller.rb
cavis: I think this is a good start.  We'll find out!

cavis (approved):
👍 looks good to me!

PR #956, opened 2024-01-31
Nil out itunes summary

### LINKED ISSUE #949: ITunes Summary handling
In the new Feeder UI, we decided not to include the itunes:summary field anymore.  Since it no longer appears [in the apple docs](https://help.apple.com/itc/podcasts_connect/#/itcb54353390).

But it turns out - Apple Podcasts is still using/preferring that field, over the description.  What to do?  Things to decide:

1. The PodcastRSSImporter still pulls in the summary - should we stop doing that?
2. For shows transitioning from Publish to Feeder UI, they won't be able to edit the existing summary.  Should we just nil the field out, when you save from the non-API controller?  Or what?
3. The RSS builder still falls back to sanitizing the description, to render the `<itunes:summary>` flag.  Should we stop doing that?

cavis:
Fixes #949.

Per that ticket/thread, we decided to continue including `<itunes:summary>` in the RSS for now.  But just set it to the sanitized description, instead of allowing a different value.

This PR adds a hidden `:summary` field to the podcast/feed/episode forms.  So the Publish app can continue to have a separate/overridden summary (though I don't think any shows do).  But as soon as you switch to Feeder UI ... as soon as you save the form, we nil out the field.

From there, the existing [def itunes_summary](https://github.com/PRX/feeder.prx.org/blob/main/app/helpers/podcasts_helper.rb#L27) method will fallback to sanitizing the description.


brandonhundt (approved):
I think this is a reasonable approach. thanks Ryan.

PR #960, opened 2024-02-06
Remove media_resources.replaced_at

### LINKED ISSUE #835: Delete media_resources.replaced_at
Once it's obvious #816 `media_versions` are working as intended, we can probably just get rid of the "replaced_at" field/logic on media_resources.

- [x] Nuke the field
- [x] Remove any "replacing" methods (marked_for_replacement? etc)
- [x] Leave the field on images though - we still use that.  Since images aren't actually required, there's still a distinction between replacing and deleting.

cavis:
Closes #835.

Now that we have the `media_versions` table, we don't need the `media_resources.replaced_at` field.

This removes it from the db, and just does a normal "mark for destruction" to set `deleted_at`.


kookster (approved):
lgtm

PR #961, opened 2024-02-09
Fix bad mp3 durations

### LINKED ISSUE #940: Move audio slicing timestamps to output options
Apparently it's more accurate to set the ffmpeg audio-slicing timestamps as output options, rather than input.

### LINKED ISSUE #941: Create new version of uncut audio for waveform
Encountered an issue where Chrome is incorrectly seeking on the waveform editor.  Apparently the file has bad duration metadata, and the [waveform does not match playback](https://github.com/bbc/peaks.js/issues/239#issuecomment-424734309).  Safari/Firefox work fine.

To work around this, we should just create a new derivative copy of the original uncut file, and use that for playback.  Probably transcode it to an mp3 of some sort?

### LINKED ISSUE #942: Replaced uncut media has "complete" badge
1. Upload uncut media to an episode with > 0 breaks.  Mark ad breaks, save, wait for processing until complete.
2. Delete the uncut media. Save.
3. Upload a new uncut media.  Save.
4. After processing completes, episode appears as "complete".  Both on the edit and index pages.  Should be incomplete.

cavis:
Fixes #941.  Fixes #940.  Fixes #942.

Porter is now returning a `"DurationDiscrepancy": 1234` milliseconds for audio (mp3s) where `mpck` and `ffprobe` disagree.

This PR takes a minimal-fix approach, and fires off a 2nd `Tasks::FixMediaTask` job when we detect a discrepancy > 500ms.  Which avoids running _everything_ through ffmpeg (slow), and storing basically duplicate copies of every original file.

### DIFF
--- a/app/controllers/episode_media_controller.rb
+++ b/app/controllers/episode_media_controller.rb
@@ -25,6 +25,9 @@ class EpisodeMediaController < ApplicationController
     authorize @episode, :update?
     @episode.assign_attributes(parsed_episode_params)
 
+    # when an uncut is destroyed, also destroy sliced contents
+    @episode.contents.each(&:mark_for_destruction) if @episode.uncut&.marked_for_destruction?
+
     respond_to do |format|
       if @episode.save
         @episode.uncut&.slice_contents!
--- a/app/models/tasks/copy_media_task.rb
+++ b/app/models/tasks/copy_media_task.rb
@@ -48,8 +48,13 @@ class Tasks::CopyMediaTask < ::Task
         media_resource.duration = porter_callback_transcode[:Duration]&.to_f&./ 1000
       end
 
-      # change status, if metadata doesn't pass validations
-      media_resource.status = "invalid" if media_resource.invalid?
+      # change status, for invalid and bad-duration files
+      if media_resource.invalid?
+        media_resource.status = "invalid"
+      elsif info[:Audio].try(:[], :DurationDiscrepancy).to_i > 500
+        media_resource.status = "processing"
+        Tasks::FixMediaTask.create! { |t| t.owner = media_resource }.start!
+      end
     end
 
     media_resource.save!
--- a/app/models/tasks/copy_media_task.rb
+++ b/app/models/tasks/copy_media_task.rb
@@ -87,10 +87,9 @@ class Tasks::CopyMediaTask < ::Task
   end
 
   def porter_slice_task
-    input_opts = []
-    input_opts << "-ss #{media_resource.slice_start}" if media_resource.slice_start.present?
-    input_opts << "-to #{media_resource.slice_end}" if media_resource.slice_end.present?
     output_opts = ["-map_metadata 0"]
+    output_opts << "-ss #{media_resource.slice_start}" if media_resource.slice_start.present?
+    output_opts << "-to #{media_resource.slice_end}" if media_resource.slice_end.present?
 
     {
       Type: "Transcode",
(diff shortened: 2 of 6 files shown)


app/models/tasks/fix_media_task.rb
```diff
@@ -0,0 +1,33 @@
...
+
+  def porter_tasks
+    [
+      {
+        Type: "Transcode",
+        Format: "INHERIT",
+        Destination: {
+          Mode: "AWS/S3",
+          BucketName: ENV["FEEDER_STORAGE_BUCKET"],
+          ObjectKey: porter_escape(media_resource.path)
```
cavis, line 20: Note this ends up overwriting our copy of the original in S3.  But I think this is fine?  I can't think of a reason we'd ever want/need it.  And we can always go back to the source url.

It'll also be easy to monitor these bad mp3s, by querying `Tasks::FixMediaTask.count`.

app/controllers/episode_media_controller.rb
```diff
@@ -25,6 +25,9 @@ def update
     authorize @episode, :update?
     @episode.assign_attributes(parsed_episode_params)
 
+    # when an uncut is destroyed, also destroy sliced contents
+    @episode.contents.each(&:mark_for_destruction) if @episode.uncut&.marked_for_destruction?
```
cavis, line 29: This is for #942 - opted to fix this in the controller, so it's less magical.  But this fixes the status displayed in the UI, by destroying sliced-contents when the uncut segment they came from is destroyed.

app/models/tasks/copy_media_task.rb
```diff
@@ -82,10 +87,9 @@ def porter_copy_task
   end
 
   def porter_slice_task
-    input_opts = []
-    input_opts << "-ss #{media_resource.slice_start}" if media_resource.slice_start.present?
-    input_opts << "-to #{media_resource.slice_end}" if media_resource.slice_end.present?
     output_opts = ["-map_metadata 0"]
+    output_opts << "-ss #{media_resource.slice_start}" if media_resource.slice_start.present?
+    output_opts << "-to #{media_resource.slice_end}" if media_resource.slice_end.present?
```
cavis, line 92: And this is #940 - still works, and I don't see a noticeable difference in resulting files.  So ... success?

app/models/tasks/fix_media_task.rb
kookster: I think it is fine? it's such a minimal change. it preserves id3 even, right?

app/models/tasks/fix_media_task.rb
```diff
@@ -0,0 +1,33 @@
...
+        },
+        FFmpeg: {
+          OutputFileOptions: "-acodec copy"
```
kookster, line 23: and with a copy, it should be all kinds of fast and not cause any issues with losing quality

app/models/tasks/copy_media_task.rb
```diff
@@ -48,8 +48,13 @@ def update_media_resource
         media_resource.duration = porter_callback_transcode[:Duration]&.to_f&./ 1000
       end
 
-      # change status, if metadata doesn't pass validations
-      media_resource.status = "invalid" if media_resource.invalid?
+      # change status, for invalid and bad-duration files
+      if media_resource.invalid?
+        media_resource.status = "invalid"
+      elsif info[:Audio].try(:[], :DurationDiscrepancy).to_i > 500
```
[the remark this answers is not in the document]
kookster: Is there any case this could accidentally get called on existing audio?
Older files won't have a DurationDiscepancy, so it would just be a 0, which is less than 500, so that's ok.

Seems like it should be fine, just worrying about any unnecessary processing of old files or worse processing that might overwrite files and cause an issue...

kookster (approved):
lgtm - I'm worrying a bit about the file overwriting, but it seems like it should be fine

PR #963, opened 2024-02-13
Release episodes for multiple feeds

### LINKED ISSUE #955: Negative feed episode publishing support
Support for Feeds with negative episode publish offsets (they publish before the default feed).

- [ ] Add UI options for negative values, in the Feed form
- [ ] Refactor [Episode.episodes_to_release](https://github.com/PRX/feeder.prx.org/blob/main/app/models/episode.rb#L83) method to account for Feed offsets ... negative or positive.

That second bit is going to be tricky.  May need to do a query for _all_ Feeds with non-null/zero offsets, and check for episodes that need release.  Or move this logic to the Feed model?  Or something else?

cavis:
Closes #955.

Currently, our `release_episodes` logic does a comparison of episode `published_at` vs `created_at`, to determine what needs to be released on the 5-minute background cron.

This is maybe a bit brittle?  And doesn't work when you consider multiple feeds being published with different episode offsets.

So instead - this uses some SQL to figure out _all_ the timestamps every episode should publish at (joining `episodes` to `feeds.episode_offset_seconds`).  And figures out the latest `publishing_queue_item.created_at` we've kicked off for that podcast_id.  If we don't have a publishing queue >= the timestamp the episode drops ... create one!


app/models/concerns/release_episodes.rb
```diff
@@ -0,0 +1,101 @@
...
+      end
+    end
+  end
+
+  def self.to_release
+    ActiveRecord::Base.connection.execute(EPISODES_TO_RELEASE)
+  end
+
+  # human readable output of what we need to release
+  def self.to_release_debug
```
cavis, line 68: I wrote this while testing, and seems handy to have even in prod?

image


May turn out we never need it, if everything gets a publishing queue within ~5 minutes.

test/models/concerns/release_episodes_test.rb
```diff
@@ -0,0 +1,116 @@
+require "test_helper"
+
+class ReleaseEpisodesTest < ActiveSupport::TestCase
+  let(:episode) { create(:episode, published_at: 30.minutes.from_now) }
+  let(:podcast) { episode.podcast }
+  let(:feed) { podcast.default_feed }
+
+  describe ".to_release" do
+    it "returns episodes/podcasts that need release" do
```
cavis, line 9: Apologies if these tests are a bit opaque.  I started out trying to set them up with static data, and then incrementing/stubbing `Time.now` to watch how things react over time.

But - since this relies on Postgres `NOW()`, that didn't really work out.  So I ended up modifying the other 2 variables: `episode.published_at` and `publishing_queue_item.created_at`.

app/jobs/release_episodes_job.rb
```diff
@@ -1,15 +0,0 @@
-class ReleaseEpisodesJob < ApplicationJob
```
cavis, line 1: As far as I can tell - this hasn't been used for 7+ years?  This was the old style SQS jobs we used to use, that would "reschedule" themselves.

But these days, our episode releasing is via `config/initializers/say_when.rb` - which calls `Podcast.release!` every 5 minutes.

PR #965, opened 2024-02-14
Bad audio bytes

cavis:
Couple changes:

- Also trigger "fixing" audio files if Porter returns any `UnidentifiedBytes`
- Adds helper methods/scopes for finding `Task`s with bad durations or unidentified bytes
- Don't slice episode segments until the parent Uncut segment is fixed
- Use the `ffprobe` detected `Format` (usually mp3) when fixing files.  Instead of INHERIT.


app/models/tasks/copy_media_task.rb
```diff
@@ -48,18 +48,29 @@ def update_media_resource
...
 
-      # change status, for invalid and bad-duration files
-      if media_resource.invalid?
-        media_resource.status = "invalid"
-      elsif info[:Audio].try(:[], :DurationDiscrepancy).to_i > 500
-        media_resource.status = "processing"
-        Tasks::FixMediaTask.create! { |t| t.owner = media_resource }.start!
-      end
+      # change status, if metadata doesn't pass validations
+      media_resource.status = "invalid" if media_resource.invalid?
```
cavis, line 52: Reverted this to a couple PRs ago, and moved the "fix this file" logic down lower.  Functionally pretty similar, but reads a bit better.

app/models/tasks/copy_media_task.rb
```diff
@@ -48,18 +48,29 @@ def update_media_resource
...
+    end
+
+    media_resource.save!
+
+    if media_resource.status_complete? && (bad_audio_duration? || bad_audio_bytes?)
+      fix_media!
+    else
+      slice_media!
```
cavis, line 60: Slight logic change here - if we need to fix the file, _don't_ slice it.  Wait until after we fix it (added this same slice_media logic to the fix_media_task).

farski (approved):

Issue #833, opened 2023-09-15
Media Versions and Dovetail Router


cavis:
When publishing to the Apple API - we want to know exactly what `media_version` (#816) we're sending.  So we know when we need to upload new files.

But - Dovetail Router has its own cache of the Feeder Episode JSON.  Which it read-behind refreshes.  So there's always a chance it serves you a stale media_version.

This isn't an issue for brand new episodes (which aren't in its cache yet - so it reads-through).  But for existing episodes, we might push a stale arrangement to Apple, and never pick up the new audio.

How might we force (for just this request from us) DTR to read-through?

- I _think_ passing a JWT already makes DTR read through?  Maybe that?
- Or some other magical secret header?
- Or should DTR return a header indicating the media version?  Is that a good idea to do anyways?

cavis:
I think this is maybe fine.  Going to close this (and the draft PR) - can resurrect if we decide to go this route later on.

PR #966, opened 2024-02-15
Uncached feeds

cavis:
For #833.

Adds a boolean indicator (only settable from the Rails console right now) to indicate that Dovetail Router should _not_ cache episode json for the feed.  This ensures requests through the Feed are always fresh, not read-behind cached.

Will need to update DTR to interpret this flag, after deploy.

### DIFF
--- a/app/models/apple/config.rb
+++ b/app/models/apple/config.rb
@@ -57,6 +57,7 @@ module Apple
         audio_format: DEFAULT_AUDIO_FORMAT,
         include_zones: ["billboard", "sonic_id"],
         tokens: [FeedToken.new(label: DEFAULT_TITLE)],
+        uncached: true,
         podcast: podcast
       )
     end
--- a/db/migrate/20240215160548_add_uncached_to_feed.rb
+++ b/db/migrate/20240215160548_add_uncached_to_feed.rb
@@ -1,0 +1,8 @@
+class AddUncachedToFeed < ActiveRecord::Migration[7.0]
+  def change
+    add_column :feeds, :uncached, :boolean, default: false
+
+    # make all apple delegated delivery feeds uncached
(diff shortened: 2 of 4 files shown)


app/models/apple/config.rb
```diff
@@ -57,6 +57,7 @@ def self.find_or_build_private_feed(podcast)
         audio_format: DEFAULT_AUDIO_FORMAT,
         include_zones: ["billboard", "sonic_id"],
         tokens: [FeedToken.new(label: DEFAULT_TITLE)],
+        uncached: true,
```
kookster, line 60: What about a cache timeout value instead of a boolean? would that ever be needed - to set a cache value that could be nothing, or some other value? hmm, maybe more complicated for no good reason, this is good.

db/migrate/20240215160548_add_uncached_to_feed.rb
```diff
@@ -0,0 +1,8 @@
+class AddUncachedToFeed < ActiveRecord::Migration[7.0]
+  def change
+    add_column :feeds, :uncached, :boolean, default: false
+
+    # make all apple delegated delivery feeds uncached
+    Feed.where(id: Apple::Config.pluck(:private_feed_id)).update_all(uncached: true)
```
kookster, line 6: nice.

kookster (approved):
lgtm, could use a cache value instead of a bool, but this works great!

cavis:
Update: i'm starting to wonder if this is even necessary.  Since:

1. Delegated delivery already handles DTR returning the wrong media_version header (it fails/retries).
2. DTR would only cache an episode json for up to 5 minutes
3. Producers generally don't replace audio right as an episode is first publishing
4. We can live with a 5 minute delay if a producer is replacing audio

Converting to draft for now.  May revisit later (need to make some other caching fixes in DTR).

PR #967, opened 2024-02-15
Fix trix editor error messages

cavis:
Fix a couple issues with the trix editor and bootstrap-field-errors:

- [x] Fix regex so it doesn't botch adding the `is-invalid` class to the tag
- [x] Add invalid message below trix editor
- [x] Hide invalid message on the hidden field trix renders

image

### DIFF
--- a/config/initializers/bootstrap_field_errors.rb
+++ b/config/initializers/bootstrap_field_errors.rb
@@ -4,7 +4,7 @@
 # wrapping them in a div that breaks bootstrap layouts
 ActionView::Base.field_error_proc = proc do |html_tag, instance|
   feedback = ""
-  if html_tag.match(/^<(input|textarea|select)/) && instance.error_message.present?
+  if html_tag.match(/^<(input|textarea|select|trix-editor)/) && instance.error_message.present?
     msg = instance.error_message.join(", ").capitalize
     feedback = "<div class=\"invalid-feedback\">#{msg}</div>"
   end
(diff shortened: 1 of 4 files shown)


config/initializers/bootstrap_field_errors.rb
```diff
@@ -4,15 +4,15 @@
...
-  if html_tag.match(/^<(input|textarea|select)/) && instance.error_message.present?
+  if html_tag.match(/^<(input|textarea|select|trix-editor)/) && instance.error_message.present?
     msg = instance.error_message.join(", ").capitalize
     feedback = "<div class=\"invalid-feedback\">#{msg}</div>"
   end
 
   if html_tag.match?(/class="(.*?)"/)
     (html_tag.sub(/class="(.*?)"/, 'class="\1 is-invalid"') + feedback).html_safe
   else
-    (html_tag.sub(/(\/>|>)/, 'class="is-invalid" \1') + feedback).html_safe
```
cavis, line 15: The previous regex was tripping up on `data-action="whatev->something"` stimulus actions.  So it would insert this invalid-feedback div right in the middle of the data attributes.

config/initializers/bootstrap_field_errors.rb
```diff
@@ -4,15 +4,15 @@
 # wrapping them in a div that breaks bootstrap layouts
 ActionView::Base.field_error_proc = proc do |html_tag, instance|
   feedback = ""
-  if html_tag.match(/^<(input|textarea|select)/) && instance.error_message.present?
+  if html_tag.match(/^<(input|textarea|select|trix-editor)/) && instance.error_message.present?
```
kookster, line 7: oh good call, yeah that was a miss on integrating that trix-editor tag

config/initializers/bootstrap_field_errors.rb
kookster: better!

kookster (approved):
lgtm

PR #972, opened 2024-02-16
Add some missing cache headers on media

cavis:
Realizing I forgot cache (and other) headers on the more recent porter tasks.

### DIFF
--- a/app/models/tasks/copy_media_task.rb
+++ b/app/models/tasks/copy_media_task.rb
@@ -116,7 +116,7 @@ class Tasks::CopyMediaTask < ::Task
         ContentType: "REPLACE",
         Parameters: {
           CacheControl: "max-age=86400",
-          ContentDisposition: "attachment; filename=\"#{porter_escape(media_resource.waveform_file_name)}\""
+          ContentDisposition: "attachment; filename=\"#{porter_escape(media_resource.file_name)}\""
         }
       },
       FFmpeg: {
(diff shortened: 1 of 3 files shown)


app/models/tasks/copy_media_task.rb
```diff
@@ -112,7 +112,12 @@ def porter_slice_task
       Destination: {
         Mode: "AWS/S3",
         BucketName: ENV["FEEDER_STORAGE_BUCKET"],
-        ObjectKey: porter_escape(media_resource.path)
+        ObjectKey: porter_escape(media_resource.path),
+        ContentType: "REPLACE",
```
cavis, line 116: Since ffmpeg is just slicing the file, not changing the format, the content type of the sliced file should match the input (uncut) file.

app/models/tasks/copy_media_task.rb
```diff
@@ -128,7 +133,12 @@ def porter_waveform_task
       Destination: {
         Mode: "AWS/S3",
         BucketName: ENV["FEEDER_STORAGE_BUCKET"],
-        ObjectKey: porter_escape(media_resource.waveform_path)
+        ObjectKey: porter_escape(media_resource.waveform_path),
+        Parameters: {
+          CacheControl: "max-age=86400",
+          ContentDisposition: "attachment; filename=\"#{porter_escape(media_resource.file_name)}\"",
+          ContentType: "application/json"
```
cavis, line 140: But this one I need to set manually, right?  Since the input file was audio.

PR #976, opened 2024-02-28
Fix unarchived episodes published

svevang:
Fixes the case where a fully delivered episode is unpublished and then republished but left in a remote `DRAFTING` state. In this case the default `episodes_to_sync` params were being calculated prior to the unarchiving of an episode. Since `ARCHIVED` episodes are not published, the main delivery routine was skipping these unarchived episodes.

### DIFF
--- a/app/models/apple/publisher.rb
+++ b/app/models/apple/publisher.rb
@@ -106,20 +106,32 @@ module Apple
       end
     end
 
-    def publish!(eps = episodes_to_sync)
+    def publish!
       show.sync!
       raise "Missing Show!" unless show.apple_id.present?
 
-      # delete or unpublished episodes
+      # Archive deleted or unpublished episodes.
+      # These episodes are no longer in the private feed.
       poll_episodes!(episodes_to_archive)
       archive!(episodes_to_archive)
       show.reload
 
+      # Un-archive episodes that are re-published.
+      # These episodes are in the private feed.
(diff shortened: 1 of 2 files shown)


app/models/apple/publisher.rb
```diff
@@ -106,20 +106,30 @@ def poll!(eps = episodes_to_sync)
       end
     end
 
-    def publish!(eps = episodes_to_sync)
```
svevang, line 109: Removing this default parameter fixes up the caching of episode state. So when we un-archive the episode, the subsequent call to `episodes_to_sync` then freshly filters episode state (derived from the private feed).

app/models/apple/publisher.rb
```diff
@@ -106,20 +106,30 @@ def poll!(eps = episodes_to_sync)
...
+    def publish!
       show.sync!
       raise "Missing Show!" unless show.apple_id.present?
 
-      # delete or unpublished episodes
+      # Archive deleted or unpublished episodes.
+      # These episodes are no longer in the private feed.
       poll_episodes!(episodes_to_archive)
       archive!(episodes_to_archive)
-      show.reload
```
svevang, line 116: These calls are only needed when episodes move from `apple_new? == true` to `apple_new? == false`. Or when an episode moves in and out of a private feed (which should not be happening during a publishing flow)

kookster (approved):
lgtm!

PR #980, opened 2024-03-11
Change Podcast "profile image" to "cover image"

### LINKED ISSUE #978: Change "Profile Image" to "Cover Image"
From Greg https://prx.slack.com/archives/CDKL9CUQ3/p1709320684327519
> I had a training session in Feeder with a producer this morning and she had some feedback about the UI I told her I’d pass along. She asked why in the metadata under “settings” it says “profile image” rather than “cover art/image” and felt it makes more sense and is more consistent with other formats to have it be “cover” rather than “profile”. 

In both Publish and Feeder we refer to the podcast-level art as a "Profile Image", but the help text clearly refers to it as "the cover image for this series". Cover Image appears to be more industry standard.

I propose we change it to "Cover Image", but wanted to check if there's a reason we call it the Profile Image. Its a 1 or 2pt ticket.

Thoughts welcomed. Thanks.

brandonhundt:
closes #978

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -818,7 +818,7 @@ en:
     title:
       episode_image: Cover Image
       feed_image: Thumbnail Image
-      itunes_image: Profile Image
+      itunes_image: Cover Image
   imports:
     create:
       success: Beginning import.


config/locales/en.yml
```diff
@@ -818,7 +818,7 @@ en:
     title:
       episode_image: Cover Image
       feed_image: Thumbnail Image
-      itunes_image: Profile Image
+      itunes_image: Cover Image
```
brandonhundt, line 821: Its a little strange that there are three labels defined here, but only two use cases. Regardless, this is the quick fix.

cavis: I think all 3 are needed, since we infer the title from the [model name](https://github.com/PRX/feeder.prx.org/blob/main/app/views/images/_new.html.erb#L4).

cavis (approved):
👍

PR #983, opened 2024-03-15
More logging

cavis:
Trying to track down some API traffic, and realized we're lacking some logging.


config/application.rb
```diff
@@ -116,17 +116,18 @@ def call(severity, time, progname, data)
...
-    config.lograge.custom_options = lambda do |event|
-      exceptions = %w[controller action format id]
+    config.lograge.custom_payload do |controller|
       {
-        user_id: event.payload[:request].session.dig("prx.auth.info", "sub"),
-        params: event.payload[:params].except(*exceptions)
+        params: controller.request.params.except(*%w[controller action format id]),
+        user_agent: controller.request.user_agent,
+        user_id: controller.try(:api_admin_token?) ? "admin-token" : controller.prx_auth_token&.user_id&.to_i,
+        user_ip: controller.request.ip,
```
cavis, line 124: So changes here:

1. Switch to the `custom_payload` newer lograge config, which is a bit simpler to access the `controller` object.
2. The `params` is functionally equivalent - though for some reason, I had to make it drop the `unpermitted_params` it was now trying to log for every API request (we have a lot of unpermitted params).
3. Adds `user_agent`
4. Fixes `user_id`, which wasn't working before
5. Adds `user_ip` ... for now, just the simple rails-parsed ip.  We'll see if it parses the XFF correctly in prod, or if we need something more verbose.

config/application.rb
```diff
@@ -116,17 +116,18 @@ def call(severity, time, progname, data)
...
-    config.lograge.custom_options = lambda do |event|
-      exceptions = %w[controller action format id]
+    config.lograge.custom_payload do |controller|
       {
-        user_id: event.payload[:request].session.dig("prx.auth.info", "sub"),
-        params: event.payload[:params].except(*exceptions)
+        params: controller.request.params.except(*%w[controller action format id]),
+        user_agent: controller.request.user_agent,
+        user_id: controller.try(:api_admin_token?) ? "admin-token" : controller.prx_auth_token&.user_id&.to_i,
+        user_ip: controller.request.ip
```
cavis, line 124: So changes here:

1. Switch to the `custom_payload` newer lograge config, which is a bit simpler to access the `controller` object.
2. The `params` is functionally equivalent - though for some reason, I had to make it drop the `unpermitted_params` it was now trying to log for every API request (we have a lot of unpermitted params).
3. Adds `user_agent`
4. Fixes `user_id`, which wasn't working before
5. Adds `user_ip` ... for now, just the simple rails-parsed ip.  We'll see if it parses the XFF correctly in prod, or if we need something more verbose.

svevang (approved):

PR #993, opened 2024-04-10
Apple::Config only belongs to the private feed

cavis:
Our Apple::Config schema is fighting a bit with the formbuilder/UI.  Since it shows up on the _private_ Feed form UI.  But only the `public_feed has_one :apple_config`.  And there are 3 foreign keys to populate/validate.

This simplifies down to a single assocation: `private_feed has_one :apple_config`.  And then I alias/delegate everything else when possible, to keep interfaces the same.

Should definitely QA this one carefully.  Because it makes me nervous!

### DIFF
--- a/app/jobs/publish_feed_job.rb
+++ b/app/jobs/publish_feed_job.rb
@@ -15,7 +15,8 @@ class PublishFeedJob < ApplicationJob
     Rails.logger.info("Starting publishing pipeline via PublishFeedJob", {podcast_id: podcast.id, publishing_queue_item_id: pub_item.id})
 
     PublishingPipelineState.start!(podcast)
-    podcast.feeds.each { |feed| publish_feed(podcast, feed) }
+    podcast.feeds.each { |feed| publish_apple(podcast, feed) }
+    podcast.feeds.each { |feed| publish_rss(podcast, feed) }
     PublishingPipelineState.complete!(podcast)
   rescue => e
     PublishingPipelineState.error!(podcast)
@@ -25,19 +26,14 @@ class PublishFeedJob < ApplicationJob
     PublishingPipelineState.settle_remaining!(podcast)
   end
 
-  def publish_feed(podcast, feed)
-    publish_apple(feed)
-    publish_rss(podcast, feed)
-  end
-
-  def publish_apple(feed)
+  def publish_apple(podcast, feed)
     return unless feed.publish_to_apple?
     res = PublishAppleJob.perform_now(feed.apple_config)
-    PublishingPipelineState.publish_apple!(feed.podcast)
+    PublishingPipelineState.publish_apple!(podcast)
     res
   rescue => e
     NewRelic::Agent.notice_error(e)
-    res = PublishingPipelineState.error_apple!(feed.podcast)
+    res = PublishingPipelineState.error_apple!(podcast)
     raise e if feed.apple_config.sync_blocks_rss?
     res
   end
--- a/app/models/apple/config.rb
+++ b/app/models/apple/config.rb
@@ -55,12 +55,7 @@ module Apple
         raise "Stopping build_apple_config" if $stdin.gets.chomp.downcase != "y"
       end
 
-      Apple::Config.new(
-        podcast: podcast,
-        public_feed: podcast.default_feed,
-        private_feed: find_or_build_private_feed(podcast),
-        key: key
-      )
+      Apple::Config.new(feed: find_or_build_private_feed(podcast), key: key)
     end
 
     def self.mark_as_delivered!(apple_publisher)
--- a/app/models/apple/config.rb
+++ b/app/models/apple/config.rb
@@ -104,9 +104,7 @@ module Apple
     end
 
     def publish_to_apple?
-      return false unless key&.valid?
-
-      public_feed.publish_to_apple?
+      !!key&.valid? && publish_enabled?
     end
 
     def build_publisher
--- a/app/models/apple/config.rb
+++ b/app/models/apple/config.rb
@@ -6,32 +6,23 @@ module Apple
     DEFAULT_TITLE = "Apple Delegated Delivery Subscriptions"
     DEFAULT_AUDIO_FORMAT = {"f" => "flac", "b" => 16, "c" => 2, "s" => 44100}.freeze
 
-    belongs_to :podcast, class_name: "Podcast"
-    belongs_to :public_feed, class_name: "Feed"
-    belongs_to :private_feed, class_name: "Feed"
-    belongs_to :key, class_name: "Apple::Key", optional: true
+    belongs_to :feed
+    belongs_to :key, class_name: "Apple::Key", optional: true, validate: true, autosave: true
 
-    delegate :title, to: :podcast, prefix: "podcast"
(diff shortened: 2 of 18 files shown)


app/jobs/publish_feed_job.rb
```diff
@@ -15,7 +15,8 @@ def perform(podcast, pub_item)
     Rails.logger.info("Starting publishing pipeline via PublishFeedJob", {podcast_id: podcast.id, publishing_queue_item_id: pub_item.id})
 
     PublishingPipelineState.start!(podcast)
-    podcast.feeds.each { |feed| publish_feed(podcast, feed) }
+    podcast.feeds.each { |feed| publish_apple(podcast, feed) }
+    podcast.feeds.each { |feed| publish_rss(podcast, feed) }
```
cavis, line 19: Okay, so here is kind of the crux of it.

It used to be the _public/default feed_ that had a `f.apple_config`.  But now it's tied to the `private_feed.apple_config`.

So - we can't rely on the `podcast.feeds[0]` (the default_feed) processing first, which blocks RSS until we publish to apple.  Instead, we do _all_ the apple publishing first, before moving onto RSS.

app/jobs/publish_feed_job.rb
```diff
@@ -25,19 +26,14 @@ def perform(podcast, pub_item)
...
 
-  def publish_feed(podcast, feed)
-    publish_apple(feed)
-    publish_rss(podcast, feed)
-  end
-
-  def publish_apple(feed)
+  def publish_apple(podcast, feed)
     return unless feed.publish_to_apple?
     res = PublishAppleJob.perform_now(feed.apple_config)
```
cavis, line 31: Took me awhile to realize - this class is a job, but we _only_ call it synchronously.  So we block right here in this thread until Apple publishing works/fails.

app/models/apple/config.rb
```diff
@@ -99,18 +90,21 @@ def self.setup_delegated_delivery(podcast, key: nil, apple_config: nil, apple_sh
...
+      if Apple::Config.where(feed_id: all_feeds).where.not(id: id).any?
+        errors.add(:feed, "podcast already has an apple config")
       end
     end
 
     def publish_to_apple?
-      return false unless key&.valid?
-
-      public_feed.publish_to_apple?
+      !!key&.valid? && publish_enabled?
```
cavis, line 107: The logic was circular here - it called up to the `public_feed`, which called back down into this `apple_config.publish_enabled` field.  Cleaned it up so this model gets to decide directly.

app/models/feed.rb
```diff
@@ -173,9 +167,7 @@ def normalize_category(cat)
   end
 
   def publish_to_apple?
-    apple_config.present? &&
-      apple_config.public_feed == self &&
-      apple_config.publish_enabled?
+    !!apple_config&.publish_to_apple?
```
cavis, line 170: Note this method now returns true on the _private_ feed, but not on the default/public feed.

test/jobs/publish_apple_job_test.rb
```diff
@@ -3,9 +3,9 @@
 describe PublishAppleJob do
   let(:episode) { create(:episode, prx_uri: "/api/v1/stories/87683") }
   let(:podcast) { episode.podcast }
-  let(:feed) { create(:feed, podcast: podcast, slug: "adfree") }
+  let(:feed) { podcast.default_feed }
   let(:private_feed) { create(:private_feed, podcast: podcast) }
-  let(:apple_config) { create(:apple_config, podcast: podcast, public_feed: feed, private_feed: private_feed) }
+  let(:apple_config) { create(:apple_config, feed: private_feed) }
```
cavis, line 8: So common pattern in the test refactor... the public `:feed` is now just the `podcast.default_feed`.  And the `:apple_config` only gets the private feed association.

Similarly, some method args change from the public-feed to the private-feed, since now it's the `private_feed.publish_apple? == true`.

test/models/apple/show_test.rb
```diff
@@ -256,7 +256,7 @@
 
     it "returns an authed url if private" do
       assert_equal apple_show.feed_published_url,
-        "https://p.prxu.org/#{public_feed.podcast.path}/#{public_feed.slug}/feed-rss.xml?auth=" + public_feed.tokens.first.token
+        "https://p.prxu.org/#{public_feed.podcast.path}/feed-rss.xml?auth=" + public_feed.tokens.first.token
```
cavis, line 259: This test was a bit atypical - IRL you'd never have a `public_feed` that wasn't the `default_feed` (with a nil slug).

app/models/apple/config.rb
kookster: I saw this, and almost changed it in my PR, but decided it was a bit out of scope - glad you saw it too!

app/models/apple/config.rb
```diff
@@ -6,32 +6,23 @@ class Config < ApplicationRecord
     DEFAULT_TITLE = "Apple Delegated Delivery Subscriptions"
     DEFAULT_AUDIO_FORMAT = {"f" => "flac", "b" => 16, "c" => 2, "s" => 44100}.freeze
 
-    belongs_to :podcast, class_name: "Podcast"
-    belongs_to :public_feed, class_name: "Feed"
-    belongs_to :private_feed, class_name: "Feed"
-    belongs_to :key, class_name: "Apple::Key", optional: true
+    belongs_to :feed
```
kookster, line 9: So the only feed associated is the private apple feed, and I assume when I get to the migrations you are renaming that as `feed_id`

app/models/podcast.rb
```diff
@@ -67,6 +67,19 @@ def default_feed
     super || build_default_feed(podcast: self, private: false)
   end
 
+  def apple_config
+    if defined?(@apple_config)
+      @apple_config
+    else
+      @apple_config = Apple::Config.where(feed_id: feeds.pluck(:id)).first
```
kookster, line 74: wait, but that's looking for the configs based on the first feed's id?
but the config is associated with the private feed now, that won't be the first feed, the default public feed will be.
This should be looking for the apple private feed id, not just the first in the list, right?

Like, this should find the `feed.publish_apple? == true`, not just the first one?

db/migrate/20240410212602_simplify_apple_config_foreign_keys.rb
```diff
@@ -0,0 +1,32 @@
...
+  def up
+    Apple::Config.find_each do |ac|
+      unless ac[:public_feed_id] == Podcast.find(ac[:podcast_id]).default_feed.id
+        raise "Apple::Config[#{ac.id} is not using the podcast.default_feed"
+      end
+    end
+
+    remove_column :apple_configs, :podcast_id
+    remove_column :apple_configs, :public_feed_id
+    rename_column :apple_configs, :private_feed_id, :feed_id
```
kookster, line 11: looks right.

kookster (changes requested):
I think the `Podcast#apple_config` method has a problem

app/models/podcast.rb
kookster: nvm, cleared this up in slack, my bad.

kookster (approved):
lgtm!

Issue #632, opened 2023-05-02
Convert Feed include/exclude tags to Episode has_and_belongs_to_many Feeds


cavis:
Stop using tags to determine which feed an episode is in, instead add a join table specifically for this.

kookster:
It might be possible to do both - use a habtm for feeds, but also with tags.

brandonhundt:
DESIGNS:

If the Apple Subscriptions Feed is selected AND the episode is published, we can display an Apple Status of either Syncing, Retrying, or Complete

If a show ONLY has a "default feed" - we don't need to display this field.

related to #796 

![Publish Box](https://github.com/PRX/feeder.prx.org/assets/369739/bf74d4fb-450d-4652-919c-4e8ecc3c6ffd)

PR #994, opened 2024-04-11
Add Apple publish status

kookster:
Implements part of #632 to display the apple publishing  status on the episode

### DIFF
--- a/app/models/concerns/apple_delivery.rb
+++ b/app/models/concerns/apple_delivery.rb
@@ -17,6 +17,10 @@ module AppleDelivery
     alias_method :apple_status, :apple_episode_delivery_status
   end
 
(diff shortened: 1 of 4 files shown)


app/models/concerns/apple_delivery.rb
```diff
@@ -17,6 +17,10 @@ module AppleDelivery
     alias_method :apple_status, :apple_episode_delivery_status
   end
 
+  def publish_to_apple?
+    podcast.apple_config&.publish_to_apple?
```
cavis, line 21: Ooh, I deleted the `podcast.apple_config` association in my PR.  I should re-add that as a memoized method or something.

cavis (approved):
Nice, looks good to me!

PR #999, opened 2024-04-17
Warn producers when new draft episodes exist

### LINKED ISSUE #992: Validate for existing drafts before allowing a one off episode to be published
Jeanne let me know a few producers are creating new drafts on dates when drafts already exist, causing issues with previously targeted faked-in flights. Jeanne highlighted the following use cases:

- producer created a new draft on a date when a draft already existed with flights targeted to it.

So, to me, possible solutions lie in a few places, but we should probably look at adding something to alert producers that a draft for "today" already exists and creating a new draft, when one already exists, may cause issues with ads.

Something like that???

brandonhundt:
Closes #992 

Adds a modal that warns producers there's an existing draft and they should use that.

- [ ] Create an episode draft for today or tomorrow
- [ ] Create a new episode
- [ ] Confirm you see this modal

Screenshot 2024-04-17 at 11 51 14 AM


cavis (approved):
👍

PR #1001, opened 2024-04-22
Replace SNS with EventBridge for Slack messages

farski:
Not tested. Needs policy update in Infra


cavis (approved):
Should verify in staging, but 👍 looks good to me.

PR #1005, opened 2024-04-24
Non-mp3s in default feed

### LINKED ISSUE #1002: Validate for WAV upload wav files when the default feed audio type isn't set
Snap uploaded a wav for the audio file and Spotify and Apple don't do wavs. Causing a missed drop. 

1. We require audio type field in the Default feed. this would prevent this on future shows
2a. We should warn the producer if a wav/flac file is uploaded without a default audio type set.
2b. Manually update the default audio type for existing shows based on what we know.

[Slack thread](https://prx.slack.com/archives/C04DUTFL4/p1713821177513419)

[Support ticket (example)](https://prx.zendesk.com/agent/tickets/197818):

cavis:
Fixes #1002.

When you upload a non-mp3 (really, a non-`audio/mpeg` format) file and have no `default_feed.audio_format`... we warn you.  Exact wording TBD.

Pretty easy to game the system by changing your `.wav` file extension.  But I think this is enough to prevent accidentally publishing WAVs into the public RSS.

image

### DIFF
--- a/app/javascript/controllers/disable_controller.js
+++ b/app/javascript/controllers/disable_controller.js
@@ -17,14 +17,14 @@ export default class extends Controller {
     if (disable) {
       this.uploadCount++
     } else {
-      this.uploadCount--
+      this.uploadCount = Math.max(0, this.uploadCount - 1)
     }
     this.submitButtons().forEach((el) => {
       if (this.uploadCount > 0) {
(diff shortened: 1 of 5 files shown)


brandonhundt (approved):
Works as expected. Thanks @cavis

PR #1006, opened 2024-04-25
Optimistic locking

### LINKED ISSUE #1000: Reject saving stale form data
So for the case:

1. User A and User B load an episode form
2. User A publishes episode
3. User B edits the title and saves

Now the episode is _unpublished_ again, with the new title from 3.  This kind of conflict could really happen on any objects/fields.

We should handle with optimistic locking... the changes in (3) should be rejected.  Rails/ActiveRecord has some built-ins for this, if you add a column.  Or DIY it with updated_at?

Then we have some decisions after (3).  What error message does User B get?  Are their changes still in the unsaved-form?  Can they opt to overwrite User A by clicking save a second time?

cavis:
Closes #1000.

Adds optimistic locking to _UI updates only_.  This avoids locking a bunch of browser-forms every time an async worker callback touches a record.  (Turns out, rails `lock_version`s bubble up, along with the updated_at timestamps we use to break caches).

If a user updates a form and it's been edited by someone else in the interim, throw up a warning.  Which lets them either discard their changes or manually confirm they want to overwrite.

Added to the pages:

- [x] Podcast
- [x] Podcast -> Engagement Settings
- [x] Feed 
- [x] Episode
- [x] Episode -> Media Files

image

### DIFF
--- a/app/views/layouts/_stale_record_modal.html.erb
+++ b/app/views/layouts/_stale_record_modal.html.erb
@@ -1,4 +1,4 @@
-<% if form.object.lock_version_changed? %>
+<% if form.object.stale? %>
   <div class="modal hide" id="stale-record" tabindex="-1" aria-labelledby="stale-record-title" aria-hidden="true">
     <div class="modal-dialog modal-dialog-centered">
       <div class="modal-content">
--- a/app/views/layouts/_stale_record_modal.html.erb
+++ b/app/views/layouts/_stale_record_modal.html.erb
@@ -3,14 +3,14 @@
     <div class="modal-dialog modal-dialog-centered">
       <div class="modal-content">
         <div class="modal-header">
-          <h5 class="modal-title" id="stale-record-title"><%= t(".title") %></h5>
+          <h5 class="modal-title fw-bold" id="stale-record-title"><%= t(".title") %></h5>
         </div>
         <div class="modal-body">
           <p><%= t ".body_html", model: form.object.model_name.human, ago: local_time_ago(form.object.updated_at) %></p>
         </div>
         <div class="modal-footer">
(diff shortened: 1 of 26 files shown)


app/views/layouts/_stale_record_modal.html.erb
```diff
@@ -3,14 +3,14 @@
...
-          <h5 class="modal-title" id="stale-record-title"><%= t(".title") %></h5>
+          <h5 class="modal-title fw-bold" id="stale-record-title"><%= t(".title") %></h5>
         </div>
         <div class="modal-body">
           <p><%= t ".body_html", model: form.object.model_name.human, ago: local_time_ago(form.object.updated_at) %></p>
         </div>
         <div class="modal-footer">
           <%= link_to t(".discard"), discard_path, class: "btn btn-discard-changed", data: {action: "unsaved#discard", unsaved_target: "discard"} %>
-          <button type="button" class="btn btn-danger" data-bs-dismiss="modal"><%= t(".continue") %></button>
+          <button type="button" class="btn btn-warning" data-bs-dismiss="modal"><%= t(".continue") %></button>
```
brandonhundt, line 13: Its a warning, not an error...

config/locales/en.yml
```diff
@@ -917,10 +917,10 @@ en:
       aside: (Not Recommended)
       label: Overwrite
     stale_record_modal:
-      body_html: While you were sleeping, someone else updated this %{model} <b>%{ago}</b>.</br></br>You can either discard your changes and make them again in a fresh form, or continue editing at your own risk.
+      body_html: Another user updated this %{model} <b>%{ago}</b>.</br></br>You can either discard your changes and make them again in a fresh form, or continue editing at your own risk.
       continue: Continue Editing
       discard: Discard changes
-      title: Woh - someone else updated this!
+      title: Warning! Another user has updated this form.
```
brandonhundt, line 923: I think this verbiage works enough. Onwards.

PR #1008, opened 2024-04-29
Remove the noImp decorator in enclosure builder

svevang:
Delegated delivery:

- Removes the `noImp` query param to enable impression logging in Dovetail
- in conjunction with https://github.com/PRX/prx-podagent/pull/43 to enable filtering as a bot

### DIFF
--- a/test/models/apple/episode_test.rb
+++ b/test/models/apple/episode_test.rb
@@ -56,12 +56,6 @@ describe Apple::Episode do
   end
 
   describe "#enclosure_url" do
-    it "should add a noImp query param" do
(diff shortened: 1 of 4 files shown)


test/models/apple/episode_test.rb
```diff
@@ -62,8 +62,8 @@
   end
 
   describe "#enclosure_url" do
```
cavis, line 64: Huh - looks like there's a dup test case just above this?

cavis (approved):
Missed a test, but otherwise looks good to me!

PR #1010, opened 2024-04-29
Better apple status

### LINKED ISSUE #995: Apple status "complete" before episode publishes
Doing some QA-ing last week in staging, I noticed that the "complete" badge was showing up 15+ minutes before the episode successfully pushed to Apple (looking at the publishing pipeline state).

kookster:
fixes #995 

After #1019 can check if the episode is in an apple subscription feed before showing status or retrieving status, but that can be a new PR (or added to this if it isn't merged yet)


app/helpers/episodes_helper.rb
```diff
@@ -16,21 +16,26 @@ def episode_explicit_options
...
+    if !apple_episode
       "not_found"
-    elsif episode.apple_status.delivered?
+    elsif apple_episode.needs_delivery?
+      "incomplete"
+    elsif apple_episode.waiting_for_asset_state?
+      "processing"
+    elsif apple_episode.audio_asset_state_error?
+      "error"
+    elsif apple_episode.synced_with_apple?
```
cavis, line 28: Are these API requests?  I don't see an `apple_episodes` table in the db, so not entirely clear where these statuses are stored.

app/models/episode.rb
```diff
@@ -254,6 +254,10 @@ def apple_mark_for_reupload!
     apple_needs_delivery!
   end
 
+  def apple_episode
+    podcast.apple_config&.build_publisher&.show&.episodes&.find { |e| e.feeder_episode.id == id }
```
cavis, line 258: So just tracing through this...

1. Get the `Apple::Config` for this podcast
2. Gets a new `Apple::Api` from that config
3. Gets a new `Apple::Publisher` with that api
4. Gets a new `Apple::Show` with that api
5. Gets many new `Apple::Episode` for most `Episode`s in the podcast
6. Filters those down to episodes in the apple subscription `Feed`
7. Finds this 1 episode within that

Kind of convoluted?  But doesn't seem like it makes any API requests along the way, so probably ok functionally.

app/helpers/episodes_helper.rb
kookster: Much of this status comes from the Apple API, but uses the latest responses which are stored.

app/models/episode.rb
kookster: yes, I agree, it's convoluted, but that's the point of this helper, to jump through the hoops from the episode -> apple_episode in this one place.

I could break up those steps into additional lines or helper methods, but this is fairly concise?

cavis: I do like how you kept it all together/concise.  Maybe just a comment, that this doesn't make API requests?  That's mainly why i traced through all that - wasn't obvious at face value if this was making remote calls.

app/helpers/episodes_helper.rb
svevang: Yep that's right. Almost all the Apple models mix in a `Apple::ApiResponse` + `has_one :apple_sync_log`.  The "SyncLog" is the last recorded remote state (usually from polling or updating the remote resource).

Apple returns json:api style responses from their endpoints. We record the last response for a remote resource in the local resource's "sync log".  The ApiResponse mixin wraps the SyncLog, and [gives some simple helpers](https://github.com/PRX/feeder.prx.org/blob/main/app/models/apple/api_response.rb#L77-L112) to unwrap json:api data.

Since the Apple::Episode is not a db table, [that class references the base ::Episode model's SyncLog by overriding the attr reader](https://github.com/PRX/feeder.prx.org/blob/main/app/models/apple/episode.rb#L219-L221).

Lastly there is the media status. [This has it's own table that records attributes of the media (version, filename, CDN source, etc](https://github.com/PRX/feeder.prx.org/blob/main/app/models/concerns/apple_delivery.rb#L33-L35).

So the result here is that `synced_with_apple?` here wraps delivery/podcast container/apple_episode state as well as uploaded media state. It uses a variety of helper methods to look down into the last stored apple api responses as well as the records of what media has been uploaded so far.

app/models/episode.rb
kookster: looking at this again, one part we may not need are steps 5 - 7.
once you have the `show`, you can construct an apple episode:
```
def apple_episode
    show = podcast.apple_config&.build_publisher&.show
    Apple::Episode.new(api: show.api, show: show, feeder_episode: self) if show
end
```
@svevang do you see an issue with skipping those steps and instantiating from the feeder episode and apple show?

svevang: Up to this point there has been no ::Episode -> Apple::Episode logic, but ultimately we need to do a join episodes that are "apple synced" to this episode.

Per your comment @cavis 1,2,3 are concomitant/procedural and (in any case) can't really be split up, so maybe that could be a helper? Then a guard on the publisher `return nil unless apple_publisher.present?` or even `apple_show.present?`, since we'd be modeling this as a relation almost (not a publishing endeavor) .

svevang: @kookster Instantiation sounds fine, but might not be what you want?

If this method is always returning an apple episode, then there is no way to indicate that this episode would never be considered for publishing to apple  like in the case of an episode that is not in the private feed.  We'd likely want to display no status in that case? (`Apple::Show#episodes` are the episodes in the private feed)

kookster: There is a guard on the view logic before this is called, where it checks if there should be apple status:
https://github.com/PRX/feeder.prx.org/blob/main/app/views/episodes/_form_status.html.erb#L34
```
 <% if episode.persisted? && episode.publish_to_apple? %>
```

I could pull that in to this helper as well, so we don't try to find an apple episode if there should not be one.

svevang: `Episode::publish_to_apple?` might need to be tweaked a bit too. https://github.com/PRX/feeder.prx.org/blob/cbd501db3cd912d742e3cacbd5984f75cf80e0e2/app/models/concerns/apple_delivery.rb#L20-L22 as it's just wrapping the apple_config check (not incorporating information about which feed the episode is in)

Looking at https://github.com/PRX/feeder.prx.org/blob/cbd501db3cd912d742e3cacbd5984f75cf80e0e2/app/models/apple/show.rb#L208-L217 Is it too onerous to keep leveraging this? So that we don't mirror the interface / have to handle the edge cases that this solves for (like an episode that's in a third unrelated feed would never be published to apple).

cavis: (Related: we're dangerously close to having a `feed.episodes` association in #1019 - would simplify finding episodes that should be in this `Feeds:AppleSubscription`)

svevang: Direct instantiation would work with a join the Feeds::AppleSubscription, though I think we'd need a new branch in the switch

```
if apple_episode.nil?
  'not found'
elsif apple_episode.apple_new?
  'new'
```
And maybe a branch for apple archived?

Joining to the Apple::Episode via the `Feeds::AppleSubscription` instance will decouple the join from the apple specific filtering that happens in `Apple::Show`. Off the top of my head, things like archived episodes will not fall into the "not found" branch if we do that. (Assuming that`feed.episodes` and `feed.feed_episodes` would not be the same right? e.g. in the case of feed limits)

It seems like we'd want the Episode -> Apple::Episode join done in one place -- and avoid duplicating the filtering logic.

kookster: ok, I am going to wait a bit on this for the other `Feeds::AppleSubscription` work to get merged, as it seems like that will simplify knowing if this episode ought to or will eventually be sent to Apple.

cavis (approved):
👍 looks good to me!  May be able to slightly improve the number of queries after #1019 - but I'm not sure it's even worth it for a `#show`.

PR #1012, opened 2024-04-30
Cautiously optimistic locking (attempt 2)

cavis:
Redoing #1006.

We were hitting issues in prod where users could not upload files.  The server was throwing `:conflict`s, but failing to render the modal.  Plus nobody else had edited the object in the background.

This appeared to be a combination of:

1. The default `lock_version` for these tables in the db is `NULL`.  Which rails casts to `0` - but for some reason this was tripping/failing the [constraint](https://github.com/rails/rails/blob/main/activerecord/lib/active_record/locking/optimistic.rb#L108) rails uses when incrementing the column.
2. Saving a nested form object (such as images/audio) _without_ changing anything on the parent Podcast/Feed/Episode.

Setting a default of `0` in the migration seems to resolve the issue for me.  That's the only change I've made to #1006.

### DIFF
--- a/app/controllers/episode_media_controller.rb
+++ b/app/controllers/episode_media_controller.rb
@@ -49,6 +49,8 @@ class EpisodeMediaController < ApplicationController
         format.html { render :show, status: :unprocessable_entity }
       end
     end
+  rescue ActiveRecord::StaleObjectError
+    render :show, status: :conflict
   end
 
   private
@@ -56,11 +58,13 @@ class EpisodeMediaController < ApplicationController
   def set_episode
     @episode = Episode.find_by_guid!(params[:episode_id])
     @episode.strict_validations = true
+    @episode.locking_enabled = true
     @podcast = @episode.podcast
   end
 
   def episode_params
     nilify params.fetch(:episode, {}).permit(
+      :lock_version,
       :medium,
       :ad_breaks,
       contents_attributes: %i[id position original_url file_size _destroy _retry],
--- a/app/javascript/controllers/toggle_value_controller.js
+++ b/app/javascript/controllers/toggle_value_controller.js
@@ -1,0 +1,12 @@
+import { Controller } from "@hotwired/stimulus"
+
+export default class extends Controller {
+  static targets = ["field"]
+  static values = { next: String }
+
+  toggle() {
+    const prev = this.fieldTarget.value
+    this.fieldTarget.value = this.nextValue
+    this.nextValue = prev
+  }
+}
--- a/app/models/application_record.rb
+++ b/app/models/application_record.rb
@@ -11,4 +11,14 @@ class ApplicationRecord < ActiveRecord::Base
   def error_message_aliases
     @@error_message_aliases
   end
+
+  def locking_enabled?
+    !!(super && @locking_enabled)
+  end
+
+  attr_writer :locking_enabled
+
+  def stale?
+    try(:lock_version_changed?) || false
+  end
 end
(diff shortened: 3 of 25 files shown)


db/migrate/20240424214558_add_locking_columns.rb
```diff
@@ -0,0 +1,7 @@
+class AddLockingColumns < ActiveRecord::Migration[7.0]
+  def change
+    add_column :episodes, :lock_version, :integer, null: false, default: 0
+    add_column :feeds, :lock_version, :integer, null: false, default: 0
+    add_column :podcasts, :lock_version, :integer, null: false, default: 0
```
cavis, line 5: This is the only change - no null lock_versions.

To QA this locally, you can:

1. Browse to an episode with no media
2. Open `psql` and manually make the lock_version null
    ```
    feeder_development=# ALTER TABLE episodes ALTER COLUMN lock_version DROP NOT NULL;
    ALTER TABLE

    feeder_development=# UPDATE episodes SET lock_version = NULL WHERE guid = '<your_guid>';
    UPDATE 1
    ```
3. In your browser, try to upload and save a media file. See the server return 409 Conflict
4. Alter another field such as segment count and upload the media file again. Save succeeds.
5. Delete audio file and save.
6. Set the NOT NULL again from psql:
    ```
    feeder_development=# UPDATE episodes SET lock_version = 0 WHERE guid = '<your_guid>';
    UPDATE 1

    feeder_development=# ALTER TABLE episodes ALTER COLUMN lock_version SET NOT NULL;
    ALTER TABLE
    ```
7. Retest just uploading a file as in (3).  See that it succeeds this time.

app/javascript/controllers/toggle_value_controller.js
```diff
@@ -0,0 +1,12 @@
+import { Controller } from "@hotwired/stimulus"
+
+export default class extends Controller {
+  static targets = ["field"]
+  static values = { next: String }
+
+  toggle() {
```
kookster, line 7: I see, for handling the modal on conflict error.

app/models/application_record.rb
```diff
@@ -11,4 +11,14 @@ def self.alias_error_messages(to_field, from_field)
...
   end
+
+  def locking_enabled?
+    !!(super && @locking_enabled)
+  end
+
+  attr_writer :locking_enabled
+
+  def stale?
+    try(:lock_version_changed?) || false
```
kookster, line 22: `|| false` to make sure this is a boolean?
If you always want a boolean, I'd add !!, with this, if try(:lock_version_changed?) is an integer > 0 or something, you aren't always getting back a boolean - I guess this prevents getting back a `nil`?

kookster (approved):
had a couple of questions, but lgtm

app/controllers/episode_media_controller.rb
```diff
@@ -49,18 +49,22 @@ def update
...
+  rescue ActiveRecord::StaleObjectError
+    render :show, status: :conflict
   end
 
   private
 
   def set_episode
     @episode = Episode.find_by_guid!(params[:episode_id])
     @episode.strict_validations = true
+    @episode.locking_enabled = true
```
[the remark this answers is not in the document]
cavis: Yeah, I did try it that way, but discovered it really needs to default to `false`. Otherwise all the `belongs_to ... touch: true` throughout the app will lock all the parent records as updated_ats bubble up.

app/models/application_record.rb
cavis: Ah, good call.  I'm always inconsistent about that.

PR #1013, opened 2024-05-02
Feat/apple feed subclass

radical-ube:
adds a `Feed::AppleSubscription` subclass in preparation for Apple Subscription UI work.


app/models/apple/config.rb
```diff
@@ -24,40 +20,6 @@ class Config < ApplicationRecord
     delegate :public_feed, to: :podcast, allow_nil: true
     alias_method :private_feed, :feed
 
-    def self.find_or_build_private_feed(podcast)
```
radical-ube, line 27: not really needed in this particular PR but moved to the new subclass. looking at it now, wondering if it's even needed now? seems like its usage was to create a feed specifically for an `apple_config` and now the subclass serves that purpose, i think.

app/models/apple/config.rb
```diff
@@ -24,40 +20,6 @@ class Config < ApplicationRecord
...
-        title: DEFAULT_TITLE,
-        audio_format: DEFAULT_AUDIO_FORMAT,
-        include_zones: ["billboard", "sonic_id"],
-        tokens: [FeedToken.new(label: DEFAULT_TITLE)],
-        podcast: podcast
-      )
-    end
-
-    # TODO: this a helper for onboarding via console, retrofit when the UX catches up
-    def self.build_apple_config(podcast, key)
```
radical-ube, line 51: same for this? maybe won't be needed once the UI PR goes up?

app/models/feed.rb
```diff
@@ -25,8 +25,6 @@ class Feed < ApplicationRecord
   alias_attribute :tokens, :feed_tokens
   accepts_nested_attributes_for :feed_tokens, allow_destroy: true, reject_if: ->(ft) { ft[:token].blank? }
 
-  has_one :apple_config, class_name: "::Apple::Config", dependent: :destroy, autosave: true, validate: true
```
radical-ube, line 28: specific to a `Feed::AppleSubscription`, right?

app/policies/feed/apple_subscription_policy.rb
```diff
@@ -0,0 +1,5 @@
+class Feed::AppleSubscriptionPolicy < ApplicationPolicy
+  def show?
```
radical-ube, line 2: maybe could leave this out until UI PR? i think this got added in a commit before realizing which branch i was in.

test/test_helper.rb
```diff
@@ -19,6 +19,7 @@
 ENV["APPLE_KEY_PEM_B64"] =
   "LS0tLS1CRUdJTiBFQyBQUklWQVRFIEtFWS0tLS0tCk1IY0NBUUVFSUhHWUUvUVBZVWtkVUFmczcyZ1FUQkE5aTVBNkRndklFOGlpV3RrQzFScDdvQW9HQ0NxR1NNNDkKQXdFSG9VUURRZ0FFaHFJSFVZUDN3QmxMdnMvQVpLM1ZHdW0vai8rMkhnVVF6dDc4TFQ0blMrckkxSlZJT0ZyVQpSVUZ6NmtSZ0pFeGxyZjdvSGZxZkxZanZGM0JvT3pmbWx3PT0KLS0tLS1FTkQgRUMgUFJJVkFURSBLRVktLS0tLQ=="
 ENV["APPLE_API_BRIDGE_URL"] = "http://localhost:3000"
+ENV["SLACK_CHANNEL_ID"] = ""
```
radical-ube, line 22: was causing an issue in an unrelated test on the podcast controller.

app/models/apple/config.rb
cavis: Yeah, I think some of that can leave now.  Will comment below.

app/policies/feed/apple_subscription_policy.rb
cavis: Yep - I think you can delete this.  You shouldn't need it at all, since you set `self.model_name` to Feed in the model.  That causes the child-model to use the policy of the parent.

app/models/feed/apple_subscription.rb
```diff
@@ -0,0 +1,90 @@
...
+
+  validate :unchanged_defaults
+  validate :only_apple_feed
+  validate :must_be_private
+
+  def set_defaults
+    self.slug ||= DEFAULT_FEED_SLUG
+    self.title ||= DEFAULT_TITLE
+    self.audio_format ||= DEFAULT_AUDIO_FORMAT
+  end
```
cavis, line 26: You should call `super` here, since the parent `Feed.set_defaults` has some things this also needs (file name, etc).

app/models/feed/apple_subscription.rb
```diff
@@ -0,0 +1,90 @@
...
+    end
+    default_feed = podcast.default_feed
+
+    Feed.new(
+      display_episodes_count: default_feed.display_episodes_count,
+      slug: DEFAULT_FEED_SLUG,
+      title: DEFAULT_TITLE,
+      audio_format: DEFAULT_AUDIO_FORMAT,
+      include_zones: ["billboard", "sonic_id"],
+      tokens: [FeedToken.new(label: DEFAULT_TITLE)],
```
cavis, line 50: Yeah, I think this whole method can go away.  Since now you can just `Feed::AppleSubscription.new` and it comes out like this.

(But you should also move `tokens` `include_zones` and `display_episodes_count` into `set_defaults`)

app/models/feed/apple_subscription.rb
```diff
@@ -0,0 +1,90 @@
...
+
+  # TODO: this a helper for onboarding via console, retrofit when the UX catches up
+  def self.build_apple_config(podcast, key)
+    if podcast.apple_config
+      Rails.logger.error("Found existing apple config for #{podcast.title}!")
+      Rails.logger.error("Do you want to continue? (y/N)")
+      raise "Stopping build_apple_config" if $stdin.gets.chomp.downcase != "y"
+    end
+
+    Apple::Config.new(feed: find_or_build_private_feed(podcast), key: key)
```
cavis, line 63: This method you may want to keep and modify.  Since the other command line helper in `Apple::Config.setup_delegated_delivery` relies on this.

But you can just replace the `find_or_build...` method with some sort of if/else to check if the podcast already has a `Feed::AppleSubscription` or not.

app/models/feed/apple_subscription.rb
```diff
@@ -0,0 +1,90 @@
...
+  end
+
+  def unchanged_defaults
+    return unless persisted?
+
+    if title_changed? || slug_changed?
+      errors.add(:feed, "cannot change properties once set")
```
cavis, line 70: So as far as business rules: you should not be able to change the `title`, `slug`, `file_name` or `audio_format`.  Everything else is fair game.

And that first param you pass is the name of the field to attach the error to.  Typically, use either the specific field name or `:base`.  (Using the actual field name might be nice, so then the error message shows up in the UI if you tried to change a field).

app/models/feed/apple_subscription.rb
```diff
@@ -0,0 +1,90 @@
...
+    end
+  end
+
+  def only_apple_feed
+    existing_feed = Feed.where(podcast_id: podcast_id, type: "Feed::AppleSubscription")
+    if existing_feed.present?
+      errors.add(:podcast, "cannot have more than one apple subscription")
+    end
```
cavis, line 78: Hrmm... I'm unable to update any existing `Feed:AppleSubscription`, since .present? returns true.  You may need to exclude this feed.id from your query.

cavis (changes requested):
I'm liking the direction of this!  A couple comments/gotchas (mostly trying to untangle some existing logic/rules).

But I think this will be a good direction to go in!

app/models/apple/config.rb
```diff
@@ -24,27 +20,9 @@ class Config < ApplicationRecord
...
-        display_episodes_count: default_feed.display_episodes_count,
-        slug: DEFAULT_FEED_SLUG,
-        title: DEFAULT_TITLE,
-        audio_format: DEFAULT_AUDIO_FORMAT,
-        include_zones: ["billboard", "sonic_id"],
-        tokens: [FeedToken.new(label: DEFAULT_TITLE)],
-        podcast: podcast
-      )
+    def find_or_build_apple_feed(podcast)
+      existing_feed = Feed.where(podcast_id: podcast.id, type: "Feed::AppleSubscription")
```
cavis, line 24: Hrm... was your intention to keep the existing method around for now?  Looks like you changed it from a class method to an instance method.

Also, this is going to return a relation, not a single record.  So the next line could return an array of feeds.  Maybe use the subclass directly, plus a rails magic finder?

```ruby
Feed::AppleSubscription.find_by_podcast_id(podcast.id)
```

app/models/feed/apple_subscription.rb
```diff
@@ -0,0 +1,72 @@
...
+      errors.add(:file_name, "cannot change once set")
+    end
+    if audio_format_changed?
+      errors.add(:audio_format, "cannot change once set")
+    end
+  end
+
+  def only_apple_feed
+    existing_feed = Feed::AppleSubscription.where(podcast_id: podcast_id).where.not(id: id)
+    if existing_feed.present?
```
cavis, line 58: Minor nitpick- calling `.present?` on an activerecord relation will load all the records from the db, and then basically call `[...array-of-things...].present?` on it.  Which would be slow if the query had 100s of results or something.

Call `.any?` instead _doesn't_ load all the records, but gives you the truthy value back.  (see https://guides.rubyonrails.org/active_record_querying.html#existence-of-objects)

cavis (approved):
👍 looks good to me!

Issue #1014, opened 2024-05-05
Apple episodes not yet synced to the public feed can have updates


kookster:
Had a weird problem with a SJ episode - it was created in Apple, but had empty title and description.
That shouldn't happen - those fields should have values by the time we're publishing and then publishing to Apple.
But, somehow it did, but then there were no calls to `Apple::Episode.update_episodes`, so the attributes stayed blank.

So, while we try to figure out the mechanism that an invalid episode got created, we can change the episode sync to update episodes that are still DRAFTING publish state in Apple, which should be prior to sync with the public episode that locks down updates.

When this happened, attempts to publish the episode failed with this info:
```
Error message:
[
  {
      "request_metadata": {
          "apple_episode_id": "1000654364957"
      },
      "api_parameters": {
          "data": {
              "type": "episodePublishingRequests",
              "attributes": {
                  "action": "PUBLISH"
              },
              "relationships": {
                  "episode": {
                      "data": {
                          "id": "1000654364957",
                          "type": "episodes"
                      }
                  }
              }
          }
      },
      "api_url": "https://api.podcastsconnect.apple.com/v1/episodePublishingRequests",
      "api_response": {
          "ok": false,
          "err": true,
          "val": {
              "data": {
                  "body": {
                      "errors": [
                          {
                              "id": "a61ebc06-d92b-4b00-a894-7a00b2cd01ac",
                              "status": "409",
                              "code": "ENTITY_ERROR.ATTRIBUTE.INVALID",
                              "title": "The provided entity includes an attribute with an invalid value",
                              "detail": "Should be non-null, non-empty, and no longer than 150 characters",
                              "source": {
                                  "pointer": "/data/attributes/title"
                              }
                          },
                          {
                              "id": "52dad9d5-48c2-4e2e-b4be-ebb67d072e2b",
                              "status": "409",
                              "code": "ENTITY_ERROR.ATTRIBUTE.INVALID",
                              "title": "The provided entity includes an attribute with an invalid value",
                              "detail": "Should be non-null, non-empty, and no longer than 4000 characters",
                              "source": {
                                  "pointer": "/data/attributes/description"
                              }
                          },
                          {
                              "id": "0ddcb670-f05c-4daf-8c9f-2b59eda2f448",
                              "status": "409",
                              "code": "ENTITY_ERROR.ATTRIBUTE.INVALID",
                              "title": "The provided entity includes an attribute with an invalid value",
[...]

PR #1015, opened 2024-05-05
Update episodes still drafting in Apple

kookster:
implements #1014


svevang (approved):
This covers two drafting cases

1) the episode is new, has never been published before
2) the episode has been unarchived, was previously published

The potential error here is a remote episode state that has blank attributes. When we go to publish the remote resource, the remote validation triggers and results in a 4xx api response.

In the case of 2, I stepped through unpublish -> archive, and then manually unarchived the episode into drafting state.  I tried in staging and the `updateEpisodes` flow works in this case!

PR #1018, opened 2024-05-09
Combine categories and keywords

cavis:
Cleaning up and simplifying categories and keywords:

- [x] Make `podcast.categories` and `episode.categories` an array field
- [x] Port all existing categories/keywords to the new single field
- [x] Drop `podcast.keywords` and `podcast.categories`
- [x] Have the RSS importer combine incoming data into the single array field
- [x] Add a Categories field to the podcast settings form


app/models/concerns/text_sanitizer.rb
```diff
@@ -22,4 +22,16 @@ def sanitize_text_only(text)
     return nil if text.blank?
     Loofah.fragment(text).scrub!(:prune).text(encode_special_chars: false)
   end
+
+  def sanitize_keywords(kws, strict)
+    Array(kws).map { |kw| sanitize_keyword(kw, kw.length, strict) }.uniq.reject(&:blank?)
```
cavis, line 27: Combining a bunch of logic from elsewhere.

This takes in any input, and returns an array of unique/stripped/non-blank strings.

Optionally with `strict`, it applies our lowercasing/only-allowed-chars logic.  Which is used for `episode.keyword_xid` and matching `feed.include_tags` / `feed.exclude_tags`.

app/models/episode.rb
```diff
@@ -224,15 +222,11 @@ def overrides
   end
 
   def categories
-    self[:categories] ||= []
+    self[:categories] || []
```
cavis, line 225: I opted to leave this postgres array field nil in the database, instead of an empty array.  Assuming it's up to the SQL-er to check for `NOT NULL` when I get around to querying on this.

app/views/podcasts/show.rss.builder
```diff
@@ -82,11 +82,11 @@ xml.rss "xmlns:atom" => "http://www.w3.org/2005/Atom",
       xml.itunes(:summary) { xml.cdata!(itunes_summary(@podcast)) }
     end
 
-    xml.itunes :keywords, @podcast.keywords.join(",") unless @podcast.keywords.blank?
+    # xml.itunes :keywords, @podcast.keywords.join(",") unless @podcast.keywords.blank?
```
cavis, line 85: TODO: do we want to keep rendering these fields - maybe with strict-sanitized (lowercased etc) content?

Or do we just drop support, since they seem deprecated anyways?

cavis: Okay, so my vote is to drop the fields from the RSS.  We can always bring them back later if it turns out folks are using this.

But to date, we haven't been setting `itunes:keywords` from our UI, so we generally don't have data to display here.  Only if it came in on an RSS import.

app/models/concerns/text_sanitizer.rb
kookster: sort?

app/models/episode.rb
```diff
@@ -21,8 +21,6 @@ class Episode < ApplicationRecord
 
   acts_as_paranoid
 
-  serialize :categories, JSON
```
kookster, line 24: will this cause issues to have current data no longer using JSON for serializing?

app/models/episode.rb
```diff
@@ -224,15 +222,11 @@ def overrides
...
+    self[:categories] || []
   end
 
   def categories=(cats)
-    self[:categories] = Array(cats).reject(&:blank?)
-  end
-
-  def keywords
-    self[:keywords] ||= []
+    self[:categories] = sanitize_keywords(cats, false).presence
```
kookster, line 229: is `.presence` returning nil if there is an empty list? is that the idea?

app/models/concerns/text_sanitizer.rb
```diff
@@ -22,4 +22,16 @@ def sanitize_text_only(text)
...
+  end
+
+  def sanitize_keyword(kw, max_length, strict)
```
kookster, line 30: should we rename this since it isn't used for `keywords` anymore?

app/views/podcasts/show.rss.builder
kookster: I vote drop from the RSS, yes, we can bring them back if it turns our to be important

test/models/episode_test.rb
```diff
@@ -166,13 +166,15 @@
...
-    episode.keywords.each { |k| refute_match(/['?!,]/, k) }
+  it "sanitizes categories" do
+    e = build_stubbed(:episode)
+
+    e.categories = []
+    assert_equal [], e.categories
+    assert_nil e[:categories]
+
+    e.categories = ["foo", " Foo ", "BAR  ", "  foo", "!@  $?"]
+    assert_equal ["foo", "Foo", "BAR", "!@  $?"], e.categories
```
kookster, line 177: oh it doesn't do case?

kookster (changes requested):
one possible method name change, and a few questions

test/models/episode_test.rb
cavis: Do we want it to?  That may not be a bad idea.  De-dup based on whichever is first in the array.

app/models/concerns/text_sanitizer.rb
cavis: Hrm... I kind of think we don't want to sort?  If an RSS client is displaying these, that UI places more importance on the first category.  And maybe the user wants to control which category shows up first.

I do agree with your case-insensitive-uniq comment later.

Also fwiw ... I'm considering adding a `episode.strict_sanitized_categories` column at some point.  Which applies and stores that strict sanitization logic, so we can do a more fuzzy SQL query on these.

app/models/concerns/text_sanitizer.rb
cavis: Well... it is still used for `episode.keyword_xid` in addition to categories.  Which... do we still use that at all?  That was an Adzerk thing, right?

kookster: We do not use that for anything - it was supposed to be way to do episode targeting in adzerk/kevel, but guid has always worked better - I'll enter an issue to remove it, actually (doesn't need to be in this PR!)

kookster: https://github.com/PRX/feeder.prx.org/issues/1032

test/models/episode_test.rb
kookster: I feel like there is no good reason to have Foo and foo, or "Arts" and "arts" if that is a better case.

kookster: It would probably be good to prevent this at the time of entry in the UI

cavis: Okay, added this.  And this did trim down on dup categories on a bunch of episodes (where they had both `keywords` and `categories` set ... i made the migration prefer the non-lowercased `categories`).

cavis: Opened #1018 for the UI bit.

app/models/concerns/text_sanitizer.rb
cavis: Cool - renamed the helper method.  So for that ticket, can just nuke the field.

kookster (approved):
awesome

app/models/episode.rb
kookster: makes sense.

app/models/concerns/text_sanitizer.rb
kookster: yeah, that makes more sense to me than forcing sort

PR #1019, opened 2024-05-13
Episode feeds checkboxes

### LINKED ISSUE #632: Convert Feed include/exclude tags to Episode has_and_belongs_to_many Feeds

cavis:
🚧 WIP.  Closes #632.

- [x] Convert feed include/exclude tags into a `has_and_belongs_to_many` relationship
- [x] Wire up checkbox UI on episode and planner forms
- [ ] More better stylez (@brandonhundt)
- [ ] Replace include/exclude tag logic with this new field, when generating feeds
- [ ] Add episodes to new feeds (maybe they just include everything in the default feed???)
- [ ] What about the RSS importer?


app/models/episode.rb
```diff
@@ -48,6 +53,8 @@ class Episode < ApplicationRecord
   validates :segment_count, numericality: {only_integer: true, greater_than: 0, less_than_or_equal_to: MAX_SEGMENT_COUNT}, allow_nil: true
   validate :validate_media_ready, if: :strict_validations
 
+  after_initialize :set_default_feeds, if: :new_record?
+  before_validation :set_default_feeds, if: :new_record?
```
svevang, line 57: Any validation or warning if the episode is not in at least one feed?

db/migrate/20240509194515_create_episode_feeds.rb
```diff
@@ -0,0 +1,41 @@
+class CreateEpisodeFeeds < ActiveRecord::Migration[7.0]
+  include TextSanitizer
+
+  def up
+    create_table :episodes_feeds, primary_key: [:episode_id, :feed_id] do |t|
+      t.belongs_to :episode
+      t.belongs_to :feed
+    end
```
svevang, line 8: Looks like if I insert duplicate records, I get dupes in `Episode#feeds`, which seems like it could cause UI weirdness and other issues.. 

A unique constraint here seems appropriate to eliminate (and find causes) for inserting dupes.

Also, might be worth an index on these two columns -- probably a composite index (and avoid hitting the table at all).

app/models/feed.rb
```diff
@@ -143,17 +183,6 @@ def path
...
-    include_in_feed = []
-    feed_max = display_episodes_count.to_i
-
-    filtered_episodes.each do |ep|
-      include_in_feed << ep if ep.include_in_feed?
-      break if (feed_max > 0) && (include_in_feed.size >= feed_max)
-    end
-    include_in_feed
-  end
-
```
svevang, line 156: Is it possible to run both systems simultaneously? Or is this roll-forward-only?

db/migrate/20240509194515_create_episode_feeds.rb
```diff
@@ -0,0 +1,41 @@
...
+
+    Feed.includes(:podcast).find_each do |f|
+      sql = <<-SQL.squish
+        INSERT INTO episodes_feeds
+        SELECT id AS episode_id, #{f.id} AS feed_id
+        FROM episodes
+        WHERE deleted_at IS NULL AND podcast_id = #{f.podcast_id}
```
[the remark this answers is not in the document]
cavis: My assumption is we don't want deleted episodes in any feeds.  (Both model has_many's have a `dependent: :delete_all` that also enforces that).

app/models/episode.rb
cavis: Yeah, I thought about that too.  I kind of want to ask users for feedback / use-cases on that?

db/migrate/20240509194515_create_episode_feeds.rb
cavis: Hrm... it shouldn't be letting you insert dups, since it has a composite primary key.  Does that not do what I think it does?  (And it's also and index, since it is the primary key).

app/models/feed.rb
cavis: We'll keep the feed include/exclude tags fields for now.  So we could roll the db/code back if something goes wrong.

But functionally, I think we can only go forward.  Or else we could have contradictions.  Example: if your Default Feed excludes `bonus` episodes ... but then you navigate to the episode and click the checkbox to add it to the bonus feed anyways.

db/migrate/20240509194515_create_episode_feeds.rb
svevang: Sorry, I had a stale branch checked out.  I see it now:

```
feeder_development=# \d+ episodes_feeds
                                       Table "public.episodes_feeds"
   Column   |  Type  | Collation | Nullable | Default | Storage | Compression | Stats target | Description 
------------+--------+-----------+----------+---------+---------+-------------+--------------+-------------
 episode_id | bigint |           | not null |         | plain   |             |              | 
 feed_id    | bigint |           | not null |         | plain   |             |              | 
Indexes:
    "episodes_feeds_pkey" PRIMARY KEY, btree (episode_id, feed_id)
    "index_episodes_feeds_on_episode_id" btree (episode_id)
    "index_episodes_feeds_on_feed_id" btree (feed_id)
Access method: heap
```

cavis: Ah, my bad.  Yeah, i started with the rails built-in `has_and_belongs_to_many` table - but ended up switching.  Probably should have just added a 2nd migration.  They're free after all!

svevang (approved):

PR #1021, opened 2024-05-15
Feat/apple feed UI

radical-ube:
Adds the Apple Config UI to an existing `Feeds::AppleSubscription` only, in preparation for another PR to add "Add Apple Subscription Feed" UI

### DIFF
--- a/app/helpers/feeds_helper.rb
+++ b/app/helpers/feeds_helper.rb
@@ -68,4 +68,8 @@ module FeedsHelper
       podcast_feed_path feed.podcast, feed, uploads_retry_params(form)
     end
   end
+
+  def apple_feed?(feed)
+    feed.type == "Feeds::AppleSubscription"
+  end
 end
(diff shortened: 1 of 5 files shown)


app/helpers/feeds_helper.rb
```diff
@@ -68,4 +68,8 @@ def feed_retry_image_path(feed, form)
       podcast_feed_path feed.podcast, feed, uploads_retry_params(form)
     end
   end
+
+  def apple_feed?(feed)
+    feed.type == "Feeds::AppleSubscription"
```
cavis, line 73: Just noting I added a `feed.apple?` method to the model in #1019, to do the same thing.  After these both merge, we should probably consolidate them.

cavis (approved):
One (optional) suggestion on making that `_form.html.erb` block a bit more readable.  But all looks/works good to me!

PR #1027, opened 2024-05-20
API episode feeds

cavis:
After #1019.

Closes https://github.com/PRX/internal/issues/1104.  Should QA that castle API scraping still works.  But I think it will.

- [x] Adds a feed getter/setter to the _authorized_ Episode API.  Default feeds are returned as `default`, and others are their slugs:
    ```
    {
      feedSlugs: ["default", "apple-sub-feed", "some-other-feed"]
    }
    ```
- [x] Change episode API to be the default_feeds

Fallout: apply display_episodes_count to API results.

### DIFF
--- a/app/models/concerns/episode_has_feeds.rb
+++ b/app/models/concerns/episode_has_feeds.rb
@@ -30,11 +30,11 @@ module EpisodeHasFeeds
   end
 
   def in_feed?(feed)
-    episodes_feeds.where(feed: feed).any?
+    published_by?(feed.episode_offset_seconds.to_i) && episodes_feeds.where(feed: feed).any?
   end
 
   def in_default_feed?
-    feeds.where(slug: nil).any?
+    published? && feeds.where(slug: nil).any?
   end
 
   def feed_slugs
--- a/app/models/concerns/episode_has_feeds.rb
+++ b/app/models/concerns/episode_has_feeds.rb
@@ -1,0 +1,49 @@
+require "active_support/concern"
+
+module EpisodeHasFeeds
+  extend ActiveSupport::Concern
+
+  included do
+    has_many :episodes_feeds, dependent: :delete_all
+    has_many :feeds, through: :episodes_feeds
+
+    after_initialize :set_default_feeds, if: :new_record?
+    before_validation :set_default_feeds, if: :new_record?
+
+    # NOTE: this DOES NOT apply the display_episodes_count
+    scope :in_feed, ->(feed) do
+      Episode.joins(:episodes_feeds)
+        .where(episodes_feeds: {feed: feed})
+        .published_by(feed.episode_offset_seconds.to_i)
+    end
+
+    # NOTE: this DOES NOT apply the display_episodes_count
+    scope :in_default_feed, -> { joins(:feeds).where(feeds: {slug: nil}).published }
+  end
+
+  def set_default_feeds
+    if feeds.blank?
+      self.feeds = (podcast&.feeds || []).filter_map do |feed|
+        feed if feed.default? || feed.apple?
+      end
+    end
+  end
+
+  def in_feed?(feed)
+    episodes_feeds.where(feed: feed).any?
+  end
+
+  def in_default_feed?
+    feeds.where(slug: nil).any?
+  end
+
+  def feed_slugs
+    feeds.pluck(:slug).map { |s| s || "default" }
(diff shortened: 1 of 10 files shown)


app/models/concerns/episode_has_feeds.rb
```diff
@@ -0,0 +1,49 @@
...
+
+    # NOTE: this DOES NOT apply the display_episodes_count
+    scope :in_feed, ->(feed) do
+      Episode.joins(:episodes_feeds)
+        .where(episodes_feeds: {feed: feed})
+        .published_by(feed.episode_offset_seconds.to_i)
+    end
+
+    # NOTE: this DOES NOT apply the display_episodes_count
+    scope :in_default_feed, -> { joins(:feeds).where(feeds: {slug: nil}).published }
```
cavis, line 21: It's not trivial to apply `display_episodes_count` to the `/api/v1/episodes` endpoint.  Since it includes multiple podcasts in that single paged response.  It would need to be something like:

```SQL
SELECT e.id, e.published_at
FROM (
  SELECT id, podcast_id, published_at, 
    ROW_NUMBER() OVER (
      PARTITION BY podcast_id ORDER BY published_at DESC
    ) AS row
  FROM episodes
  WHERE published_at <= NOW()
) e 
INNER JOIN feeds f ON (f.slug IS NULL AND f.podcast_id = e.podcast_id)
WHERE (display_episodes_count IS NULL OR row <= display_episodes_count)
```

That isn't exactly quick, when returning all podcasts.

app/models/feed.rb
```diff
@@ -41,11 +41,12 @@ class Feed < ApplicationRecord
 
   validates :slug, uniqueness: {scope: :podcast_id}, if: :podcast_id?
   validates_format_of :slug, allow_nil: true, with: /\A[0-9a-zA-Z_-]+\z/
-  validates_format_of :slug, without: /\A(images|\w{8}-\w{4}-\w{4}-\w{4}-\w{12})\z/
+  validates_format_of :slug, without: /\A(default|images|\w{8}-\w{4}-\w{4}-\w{4}-\w{12})\z/
```
cavis, line 44: Since I'm using `"default"` in that `feedSlugs` API - that's now not allowed to be your explicit slug/path.

app/models/feed.rb
```diff
@@ -41,11 +41,12 @@ class Feed < ApplicationRecord
...
   validates :file_name, presence: true, format: {with: /\A[0-9a-zA-Z_.-]+\z/}
   validates :include_zones, placement_zones: true
   validates :audio_format, audio_format: true
   validates :title, presence: true, unless: :default?
+  validates :episode_offset_seconds, numericality: {equal_to: 0}, allow_nil: true, if: :default?
```
cavis, line 49: To keep things "simple" ... I decided to restrict default_feeds to _not_ have any episode offset.  Episodes must publish at exactly their `published_at` in the default feed.

Which is currently the case in prod, and I can't imagine why you'd want to offset the default feed?  It would be confusing.

kookster (commented):
I have a few questions - like do we need a public /episodes endpoint not limited a podcast/feed, should we allow offset/windowing the default feed?

app/models/feed.rb
cavis: Now that negative offsets work correctly ... I think it makes the most sense for the default feed to always be at 0, and the private feed to be `-604800` (7 days early).

app/models/concerns/episode_has_feeds.rb
cavis: Yeah, I think that's the next thing we should decide after this PR.  What's the utility of our public endpoints?  Just to provide a JSON version of what's in public RSS?  Or something else?

kookster: I would say to provide a json version of the public feed and individual public episodes, and I am fairly convinced it should mirror the public feed in the scope of what data it provides, but as a JSON API, and this gets us much much closer to that

app/models/feed.rb
kookster: I think this is 6 of one, -6 of the other, so I have no strong objection here - it seems weird to me, but let's wait and see if we got a podcaster that wants to do this, and we can see what they think, and this is good until such feedback

kookster (approved):
I'm persuaded. lgtm

PR #1031, opened 2024-05-30
Feat/add apple feed UI

### LINKED ISSUE #796: Implement Apple Subscriptions "Feed type" UI
Producer creates a new "feed" with the Feed type set to Apple Subscriptions

Producer provides their key and key files and they are on their way. We can default a lot of the data to the default feed (or make decisions for them in the backend, like defaulting to FLAC files). meaning there won't be a lot of overrides needed

So while there is a lot going on behind the scenes, to the producer, its straightforward.

Feeder _ Feeds form _ Apple Subscriptions Example

radical-ube:
resolves #796


app/javascript/controllers/apple_key_controller.js
```diff
@@ -0,0 +1,40 @@
...
+      reader.onload = (event) => resolve(event.target.result)
+      reader.onerror = (error) => {
+        console.error("Unable to read text file", file)
+        resolve("")
+      }
+      reader.readAsText(file)
+    })
+  }
+
+  parseKey(text) {
```
radical-ube, line 32: so i'm not actually sure what `.p8` files look like. i'm working off of the sample `.pem` file that Sam put in the test fixtures that the `Apple::Key` tests use. so i'm parsing based on that format.

app/views/feeds/_form_apple_config.html.erb
```diff
@@ -3,28 +3,28 @@
...
+            <%= key_fields.label :provider_id, "Provider ID" %>
           </div>
-
-          <div>
-            <div class="form-floating">
-              <%= key_fields.text_field :key_id, disabled: config_form.object.persisted? %>
-              <%= key_fields.label :key_id %>
-            </div>
+          <div class="form-floating input-group">
+            <%= key_fields.text_field :key_id, disabled: config_form.object.persisted?, data: {apple_key_target: "key", action: "keyup->apple-key#convertKeyToB64"} %>
```
radical-ube, line 18: the `keyup` event trigger is for the case if a user decides to just copy/paste the key into the field, i still want to run the method that converts it to base64 for the (hidden) `key_pem_b64` field.

app/views/feeds/_form_apple_config.html.erb
```diff
@@ -3,28 +3,28 @@
...
+            <%= key_fields.label :key_id, "Apple Key" %>
+            <% unless config_form.object.persisted? %>
```
radical-ube, line 20: hides the file upload button if the feed is already saved.

app/models/apple/config.rb
```diff
@@ -20,6 +20,8 @@ class Config < ApplicationRecord
     delegate :public_feed, to: :podcast, allow_nil: true
     alias_method :private_feed, :feed
 
+    accepts_nested_attributes_for :key, allow_destroy: true, reject_if: :all_blank
```
cavis, line 23: We shouldn't allow deleting these, so I'd remove that bit.

Also - you'll want to remove the `reject_if: :all_blank`.  That way, if you try to create one of these with all fields blank, it'll throw a validation error.

app/views/feeds/_tabs.html.erb
```diff
@@ -14,7 +14,15 @@
   </div>
   <% if policy(podcast).update? %>
     <div class="fixed-bottom col-sm-12 col-md-3 col-xl-2">
-      <%= link_to "Add a Feed", new_podcast_feed_path(podcast), class: "btn btn-success d-flex sticky-md-bottom m-4 shadow" %>
+      <div class="btn-group dropup d-flex sticky-md-bottom m-4 shadow">
```
cavis, line 17: This 2nd Add-Apple-Feed dropup button should show _if_ you don't already have an apple feed.

If you already have one, just show the plain "add a feed" button".

app/views/feeds/_form_apple_config.html.erb
cavis: So slight UI update... re-reading a bunch of threads, it _sounds_ like the filename of the pem that Apple gives you already contains the provider_id and key_id?  In which case, we should try to auto parse that for users.

So complete workflow:

1. Click on the New Apple Feed button, form shows _only_ a file-upload field
2. Upload your PEM with a name like `<some-uuid-4-looking-provider-id>_<10-digit-key-id>.p8`
3. Your stimulus controller parses that filename, fills in 2 hidden fields for `provider_id` and `key_id`
4. Your stimulus controller also _reads_ the file, and fills in the `key_pem_b64` hidden field
5. Click save, browser POSTs those 3 hidden fields
6. If any validation failed, error message is displayed on the form under the file-upload field

Then once you already have a saved Feeds::AppleSubscription, the form shows the (current) readonly `provider_id` and `key_id` fields.

cavis (changes requested):
Structurally I think this is great.  But re-reading slack threads, we may _only_ need the file-upload field?  (And 3 hidden fields).  We should get ahold of an actual apple pem file and try it.

app/javascript/controllers/apple_key_controller.js
```diff
@@ -0,0 +1,57 @@
...
+
+  convertFileName(fileName) {
+    const fields = fileName.split("_")
+    const provider = fields[0]
+    const keyId = fields[1].split(".")[0]
+
+    this.providerTargets.forEach((target) => (target.value = provider))
+    this.providerTarget.disabled = false
+    this.providerTarget.focus()
+    this.providerTarget.disabled = true
```
radical-ube, line 35: this extra step of focusing on L33-35 and L38-40 is just to get the label out of the way of the input box.

app/views/feeds/_form_apple_config.html.erb
```diff
@@ -3,28 +3,31 @@
...
       <%= form.fields_for :apple_config do |config_form| %>
         <%= config_form.fields_for :key do |key_fields| %>
-          <div class="mb-4">
-            <div class="form-floating">
-              <%= key_fields.text_field :provider_id, disabled: config_form.object.persisted? %>
-              <%= key_fields.label :provider_id %>
-            </div>
+          <% unless config_form.object.persisted? %>
+            <%= key_fields.file_field :key_file, class: "mb-4 form-control", accept: ".p8, .pem", data: {action: "apple-key#convertFileToKey"} %>
+          <% end %>
```
radical-ube, line 15: i opted to leave the disabled fields showing at all times. i was having trouble getting them to conditionally pop up for the error and this was the cleanest i could get to work properly.

cavis: +1 works for me.

app/views/feeds/_form_apple_config.html.erb
```diff
@@ -3,28 +3,31 @@
...
-              <%= key_fields.label :key_id %>
-            </div>
+          <div class="form-floating">
+            <%= key_fields.text_field :key_id, disabled: true, data: {apple_key_target: "key"} %>
+            <%= key_fields.label :key_id, "Apple Key" %>
           </div>
+
+          <%= key_fields.hidden_field :provider_id, data: {apple_key_target: "provider"} %>
+          <%= key_fields.hidden_field :key_id, data: {apple_key_target: "key"} %>
+          <%= key_fields.hidden_field :key_pem_b64, data: {apple_key_target: "pem"} %>
```
cavis, line 27: One security thing: can you omit this entirely if the record is persisted?  We don't want to have the base 64'd pem showing up in the html.

app/views/feeds/_tabs.html.erb
```diff
@@ -10,7 +10,17 @@
...
+        <%= link_to "Add a Feed", new_podcast_feed_path(podcast), class: "btn btn-success flex-grow-1" %>
+        <% unless @feeds.select { |f| f.type == "Feeds::AppleSubscription" }.present? %>
```
cavis, line 15: I added a `.apple?` method to Feeds awhile back.  So now could simplify this to `@feeds.none?(&:apple?)`.

app/policies/feed_policy.rb
```diff
@@ -11,11 +11,15 @@ def create?
...
+    create?
+  end
+
   def update?
     PodcastPolicy.new(token, resource.podcast).update?
   end
 
   def destroy?
-    resource.custom? && update?
+    resource.custom? && update? && resource.type != "Feeds::AppleSubscription"
```
cavis, line 23: Or now - `resource.apple?`

cavis (changes requested):
Oookay, just 1 security thing and i think this can go.  (Also noticed you could use the `.apple?` method instead of checking `.type` ... but that's pretty minor).

cavis (approved):
👍 looks good to me!

PR #1035, opened 2024-06-04
Make sure parent/child classes don't all share the same aliases

cavis:
Whoops - every model was sharing the same error message aliases.

### DIFF
--- a/app/models/application_record.rb
+++ b/app/models/application_record.rb
@@ -2,14 +2,22 @@ class ApplicationRecord < ActiveRecord::Base
   primary_abstract_class
   include HalApi::RepresentedModel
 
-  @@error_message_aliases = {}
-
   def self.alias_error_messages(to_field, from_field)
-    @@error_message_aliases[to_field.to_s] = from_field.to_s
+    aliases = self.error_message_aliases
+    aliases[to_field.to_s] = from_field.to_s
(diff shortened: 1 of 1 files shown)


app/models/application_record.rb
```diff
@@ -2,14 +2,22 @@ class ApplicationRecord < ActiveRecord::Base
   primary_abstract_class
   include HalApi::RepresentedModel
 
-  @@error_message_aliases = {}
-
   def self.alias_error_messages(to_field, from_field)
-    @@error_message_aliases[to_field.to_s] = from_field.to_s
+    aliases = error_message_aliases
+    aliases[to_field.to_s] = from_field.to_s
+    class_variable_set(:@@error_message_aliases, aliases)
```
kookster, line 8: Concurrency issue?

kookster (approved):
Lgtm- I'm slightly wondering if there could be a concurrency issue if 2 processes loaded classes simultaneously, but in rails/ruby probably fine

app/models/application_record.rb
cavis: I don't _think_ so, since all the models get preloaded in prod?  And in any case - should be idempotent to keep adding the same keys to the same hash, right?

Prior to this, was seeing some oddness in dev.  Where calling something like `Feed.new.error_message_aliases` would change depending on which models had been loaded already.

PR #1037, opened 2024-06-05
Disable OTEL in dev

cavis:
Check if OTEL envs are set before configuring it.  Gets rid of a bunch of warnings when developing/testing locally.


config/environment.rb
```diff
@@ -3,8 +3,10 @@
 # Load the Rails application.
 require_relative "application"
 
-OpenTelemetry::SDK.configure do |c|
-  c.use_all
+if ENV["OTEL_EXPORTER_OTLP_ENDPOINT"].present? && !Rails.const_defined?(:Console)
```
kookster, line 6: Oh, so don't run it in the console, ever? that's what that `!Rails.const_defined?(:Console)` does?
neat!

kookster (approved):
lgtm

PR #1041, opened 2024-06-12
Allow longer production plans

### LINKED ISSUE #1042: Allow production planning to start longer than a year out
Had a request to plan a new season for a show more than a year out.
Can we allow planning and episodes further out?

kookster:
Fixes #1042 
- [x] Make the number of years for planning a constant
- [x] Increase that constant from 2 to 3 years

### DIFF
--- a/app/views/podcast_planner/_form_calculator_dates.html.erb
+++ b/app/views/podcast_planner/_form_calculator_dates.html.erb
@@ -16,7 +16,7 @@
 
 <div class="col-6 mb-4">
   <div class="form-floating input-group">
-    <%= form.date_field :start_date, min: DateTime.tomorrow, max: DateTime.tomorrow + 1.years, value: DateTime.tomorrow, data: {action: "click#submit"} %>
+    <%= form.date_field :start_date, min: DateTime.tomorrow, max: DateTime.tomorrow + 2.years, value: DateTime.tomorrow, data: {action: "click#submit"} %>
     <%= form.label :start_date, t(".label.start_date") %>
     <%= field_help_text t(".help.start_date") %>
   </div>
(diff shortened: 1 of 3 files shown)


app/views/podcast_planner/_form_draft.html.erb
```diff
@@ -55,7 +55,7 @@
             </button>
           </div>
 
-          <% 24.times do |i| %>
+          <% (12 * PodcastPlanner::MAX_YEARS).times do |i| %>
```
cavis, line 58: Hmm... is this off by 1?  I'm now getting months out to May 2027 here.  But the dropdown field only goes out to May 14th 2026.

kookster: the "start" date should go out 2 years, but then this calendar and the validations allow setting episodes 3 days out.
this mirrors how it worked before, where there was a year differential between the last start date, and the end of where episodes can be created out until.

cavis: Yeah, so `past_max_date?` is going to be May 2027.  Does that mean the `Start Date` field should also go out to 2027?

kookster: ...unless I set it up wrong, but that was what I was going for.

kookster: should the start date and end date match up? they didn't before - I believe they were a year apart before, and that was what I tried to retain here, just shifting both from 1 and 2 years, to 2 and 3 years

cavis: I think the dropdown-calendar and the months at the bottom should be the same.

In other words ... you're allowed to `Start Date` your planned episodes anytime between tomorrow and 3 years from now.  BUT - that `past_max_date?` method enforces a hard limit of 3 years from now.

(If I start planning weekly episodes 2 years and 11 months from now... it'll only give me 4 no matter how many I ask for).

kookster: done

app/views/podcast_planner/_form_calculator_dates.html.erb
```diff
@@ -16,7 +16,7 @@
 
 <div class="col-6 mb-4">
   <div class="form-floating input-group">
-    <%= form.date_field :start_date, min: DateTime.tomorrow, max: DateTime.tomorrow + 1.years, value: DateTime.tomorrow, data: {action: "click#submit"} %>
+    <%= form.date_field :start_date, min: DateTime.tomorrow, max: DateTime.tomorrow + PodcastPlanner::MAX_YEARS.years, value: DateTime.tomorrow, data: {action: "click#submit"} %>
```
kookster, line 19: start date now is also out for the full `MAX_YEARS` @cavis

cavis (approved):
👍

PR #1045, opened 2024-06-13
Strip guids of leading and trailing whitespace

### LINKED ISSUE #1044: Prevent accidental trailing spaces guid values
messes up some publishing integrations when there is a trailing space(s) on explicit guid values set on an episode.

kookster:
fix #1044

### DIFF
--- a/app/models/episode.rb
+++ b/app/models/episode.rb
@@ -312,6 +312,7 @@ class Episode < ApplicationRecord
     self.subtitle = sanitize_text_only(subtitle) if subtitle_changed?
     self.summary = sanitize_links_only(summary) if summary_changed?
     self.title = sanitize_text_only(title) if title_changed?
+    self.original_guid = original_guid.strip if !original_guid.blank? && original_guid_changed?
   end
 
   def description_with_default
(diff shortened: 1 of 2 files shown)


app/models/episode.rb
```diff
@@ -312,6 +312,7 @@ def sanitize_text
     self.subtitle = sanitize_text_only(subtitle) if subtitle_changed?
     self.summary = sanitize_links_only(summary) if summary_changed?
     self.title = sanitize_text_only(title) if title_changed?
+    self.original_guid = original_guid.strip if !original_guid.blank? && original_guid_changed?
```
cavis, line 315: We're a bit inconsistent in this app about using setters vs validation-hooks to mutate data in this app.  But lumping this in with "sanitize stuff" makes sense to me.  Good sense!

cavis (approved):
Works for me 👍

app/models/episode.rb
kookster: yeah, this didn't seem like a validation - we also strip these on import, which made me feel confident this should just get cleaned

PR #1050, opened 2024-06-27
Add table migration and init model

### LINKED ISSUE #1047: Create a new `transcripts` table/model
First step in #969.  Setup the migrations/models for Transcripts.

- [ ] Probably named `Transcript` (a top-level model)
- [ ] Similar to EpisodeImage fields: status, guid, url, original_url, mime_type, file_size, format
- [ ] To start, let's treat it has episode has-one transcript.  We can always change that to many later.
- [ ] Make paranoid.

radical-ube:
resolves #1047

### DIFF
--- a/app/models/episode.rb
+++ b/app/models/episode.rb
@@ -33,6 +33,7 @@ class Episode < ApplicationRecord
 
   has_one :ready_image, -> { complete_or_replaced.order("created_at DESC") }, class_name: "EpisodeImage"
   has_one :uncut, -> { order("created_at DESC") }, autosave: true, dependent: :destroy, inverse_of: :episode
+  has_one :transcript, dependent: :destroy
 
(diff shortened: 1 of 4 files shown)


app/models/episode.rb
```diff
@@ -33,6 +33,7 @@ class Episode < ApplicationRecord
 
   has_one :ready_image, -> { complete_or_replaced.order("created_at DESC") }, class_name: "EpisodeImage"
   has_one :uncut, -> { order("created_at DESC") }, autosave: true, dependent: :destroy, inverse_of: :episode
+  has_one :transcript, dependent: :destroy, inverse_of: :episode
```
cavis, line 36: I _might_ put an order on this, just because the others have one?

cavis (approved):
One minor suggestion, but otherwise looks good to me.

PR #1052, opened 2024-07-09
CopyTranscriptTask

### LINKED ISSUE #1048: Create a new `CopyTranscriptTask` as a subclass of `Task`
Second step of #969.  Setup the async task to copy/process the transcript.

- [ ] Create a new `CopyTranscriptTask` as a subclass of `Task`
- [ ] To start, fire off porter Inspect / Copy tasks and parse the results back into the Transcript
- ~TODO: does Porter need any work to return the correct format/type for a transcript file?  (cc: @farski)~
  - Update: punting on this till later. Detecting different "types" of transcripts is tricky.

radical-ube:
resolves #1048


app/models/tasks/copy_transcript_task.rb
```diff
@@ -0,0 +1,47 @@
...
+
+  def update_transcript
+    transcript.status = status
+
+    if complete?
+      info = porter_callback_inspect
+      transcript.size = porter_callback_size
+
+      # only return for actual images - not detected images in id3 tags
+      if info[:Transcript] && porter_callback_mime&.starts_with?("transcript/")
```
cavis, line 37: Hmm... I'm not sure this is going to exist.  At least no yet - Porter `Inspect` knows things about [audios/videos/images](https://github.com/PRX/Porter/blob/main/src/lambdas/inspect/index.js#L44-L46), but nothing about transcripts.

app/models/tasks/copy_transcript_task.rb
```diff
@@ -0,0 +1,47 @@
...
+      }
+    ]
+  end
+
+  def update_transcript
+    transcript.status = status
+
+    if complete?
+      info = porter_callback_inspect
+      transcript.size = porter_callback_size
```
cavis, line 34: I think this field is named `file_size` on this table, not `size`?

You'll also want to grab the `mime_type` (see `copy_media_task.rb`).

Have you tried actually running one of these through Porter?  Might be a good way to just validate everything is working.

app/models/transcript.rb
```diff
@@ -2,4 +2,96 @@ class Transcript < ApplicationRecord
...
+    super
+    if original_url_changed?
+      reset_transcript_attributes
+    end
+    self[:original_url]
+  end
+
+  def reset_transcript_attributes
+    self.format = nil
+    self.status = :created
```
cavis, line 67: I think you'll want to reset file size and mime type too.

app/models/transcript.rb
```diff
@@ -2,4 +2,96 @@ class Transcript < ApplicationRecord
...
+
+  def retry!
+    if retryable?
+      status_retrying!
+      copy_media(true)
+    end
+  end
+
+  def _retry=(_val)
+    retry!
```
cavis, line 95: Just to note - these "retry" methods are generally for the UI.  So we give users a way to re-kick-off Porter without needing to delete and reupload their file.  (And it's good to have them for Transcripts UI as well!)

test/models/transcript_test.rb
```diff
@@ -0,0 +1,59 @@
...
+    it "skips creating task if complete" do
+      transcript.task = nil
+      assert transcript.status_complete?
+      transcript.copy_media
+
+      assert_nil transcript.task
+    end
+
+    it "skips creating task if one exists" do
+      task = Tasks::CopyImageTask.new
```
cavis, line 51: CopyImageTask?

test/models/transcript_test.rb
```diff
@@ -0,0 +1,59 @@
...
+      transcript.status = "created"
+      transcript.task = task
+      transcript.copy_media
+
+      assert_equal task, transcript.task
+    end
+  end
```
cavis, line 58: I also tend to write the usual I/O tests ... what are we sending to Porter (`#porter_tasks`) and what is Porter updating on callback (`#update_transcript`).  But sounds like you may already be working on those.

cavis (changes requested):
Layout and ported code is looking good.  A few specific things may need tweaking - may be worth trying to get this actually running against Porter to make sure all works.

app/models/transcript.rb
```diff
@@ -2,4 +2,101 @@ class Transcript < ApplicationRecord
...
   belongs_to :episode
+
+  has_one :task, -> { order("id desc") }, as: :owner
+  has_many :tasks, as: :owner
+
+  before_validation :initialize_attributes, on: :create
+
+  validates :original_url, presence: true
+
+  validates :mime_type, inclusion: {in: %w[text/plain application/json application/pdf]}, unless: proc { |t| t.mime_type.blank? }
```
cavis, line 13: Is that proc any different than `allow_blank: true`?

test/fixtures/sampletranscript.pdf
cavis, line 1: Is there a test that uses this fixture?  I see `sampletranscript.pdf` in the factory, but is it actually loading up this file for something?

cavis (approved):
Couple of questions, but this all looks good / works for me. 👍 onwards!

PR #1054, opened 2024-07-11
fix z-index issues

brandonhundt:
See: https://prx.slack.com/archives/CG7G1FPNJ/p1720705946618059

### DIFF
--- a/app/assets/stylesheets/shared/layout.scss
+++ b/app/assets/stylesheets/shared/layout.scss
@@ -25,3 +25,8 @@ h3 {
   position: relative;
   display: block;
 }
+
+.sticky-lg-top,
+.sticky-md-top {
+  z-index: $zindex-dropdown;
+}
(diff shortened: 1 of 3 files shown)


app/assets/stylesheets/shared/layout.scss
```diff
@@ -25,3 +25,8 @@ h3 {
   position: relative;
   display: block;
 }
+
+.sticky-lg-top,
+.sticky-md-top {
+  z-index: $zindex-dropdown;
+}
```
brandonhundt, line 32: This ensures tabs and status boxes stay behind the nav dropdowns.

app/views/episode_media/_form.html.erb
```diff
@@ -14,9 +14,7 @@
           </div>
         </div>
         <div class="col-lg-4 d-grid align-content-start gap-3">
-          <div class="sticky-lg-top">
-            <%= render "form_status", episode: episode, form: form %>
-          </div>
```
brandonhundt, line 19: Since this is the only page that uses off canvas, if we remove the sticky (which isn't needed), it solves our z-index charade here.

cavis (approved):
👍 tested all the pages i could think of, and this looks good!

Issue #1057, opened 2024-07-24
Episode segmenter labels incorrect for shows with an "intro" segment


cavis:
Some shows that have an "Intro" original segment in Augury:

![image](https://github.com/user-attachments/assets/c0e41fdd-0f09-4153-ad54-5ebe6a2ec069)

But when segmenting your audio in Feeder, the labels are a bit misleading:

![image](https://github.com/user-attachments/assets/a89752d9-7c0d-4a34-8e46-cd1697ca5623)

Bad idea: make the labels very generic?  "Break 1/2/3" and "Trim start" / "Trim end"?

brandonhundt:
+1 calling them "Breaks"

PR #1059, opened 2024-07-29
update Segmenter labels

brandonhundt:
Closes: #1057 

Change break language from "Midroll" to "Break"

Screenshot 2024-07-29 at 10 46 17 AM


cavis (approved):
👍 looks good to me!

I do wonder if the `Preroll` label will still cause confusion for these episodes, since prerolls will really be `break 1`.  But this is a good step forward.

PR #1062, opened 2024-08-01
Log info instead of error for 404s

cavis:


app/controllers/errors_controller.rb
```diff
@@ -0,0 +1,10 @@
+class ErrorsController < ActionController::Base
+  include PrxAuth::Rails::Controller
```
cavis, line 2: If the user is authenticated and gets a 404, we'll get the `user_id` in the log.info statement.

config/routes.rb
```diff
@@ -53,4 +53,8 @@
 
   match "/api", via: [:get], to: redirect("/api/v1")
   root "podcasts#index"
+
+  unless Rails.configuration.consider_all_requests_local
+    match "*unmatched", via: :all, to: "errors#not_found"
```
cavis, line 58: For local dev, I left things alone.  You'll still get a log.error and the fancy "route missing" page.  But in prod, they just get the static 404 page.

app/controllers/errors_controller.rb
```diff
@@ -0,0 +1,10 @@
...
+
+  layout false
+  skip_forgery_protection
```
cavis, line 5: Prevents an additional `ActionController::InvalidAuthenticityToken` when someone POSTs to a non-existent route.

PR #1064, opened 2024-08-05
Remove last item added if dupe

### LINKED ISSUE #1033: Prevent case-insensitive duplicate categories in UI
For podcast/episode category fields - after #1018 if you enter `["Art", "art", "ART"]` and save, the server will dedup that down to just `["Art"]`.

We should prevent entering those dup tags into the field in the first place.  (The tag field already prevents same-case dups).

radical-ube:
resolves #1033

### DIFF
--- a/app/javascript/controllers/tag_select_controller.js
+++ b/app/javascript/controllers/tag_select_controller.js
@@ -57,12 +57,11 @@ export default class extends Controller {
   }
 
   dedupeValues(values) {
-    const sanitizedVals = values.map((val) => {
-      return val.trim().toLowerCase()
-    })
-    const dedupedVals = [...new Set(sanitizedVals)]
-    return dedupedVals.filter((val) => {
-      return val !== null || val !== undefined || val !== ""
-    })
+    const sanitizedVals = values
+      .filter((val) => val)
+      .map((val) => {
+        return val.trim().toLowerCase()
+      })
+    return [...new Set(sanitizedVals)]
   }
 }
--- a/app/javascript/controllers/tag_select_controller.js
+++ b/app/javascript/controllers/tag_select_controller.js
@@ -35,8 +35,13 @@ export default class extends Controller {
 
   change() {
     const valueWas = this.element.dataset.valueWas
-    const values = Array.from(this.element.selectedOptions).map((o) => o.value)
+    const initialValues = Array.from(this.element.selectedOptions).map((o) => o.value)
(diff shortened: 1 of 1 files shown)


app/javascript/controllers/tag_select_controller.js
```diff
@@ -35,9 +35,14 @@ export default class extends Controller {
 
   change() {
     const valueWas = this.element.dataset.valueWas
-    const values = Array.from(this.element.selectedOptions).map((o) => o.value)
+    const initialValues = Array.from(this.element.selectedOptions).map((o) => o.value)
+    const values = this.dedupeValues(initialValues)
     const valueIs = JSON.stringify(values)
 
+    if (values.length !== this.element.selectedOptions.length) {
+      Tags.getInstance(this.element).removeLastItem()
```
radical-ube, line 43: the only thing i'm unsure about this is that this assumes that the user is typing in categories one at a time. i'm not sure if there are other ways the user has of bulk "importing" categories or entering a bunch of different selections in one swoop? this method just removes the last thing the user tries to input, if it was a dupe.

cavis: Tried a few things (pasting in multiple lines, tabs, etc)... couldn't get it to add more than one tag at a time.  So this seems safe?

Perusing the docs a bit - there is also a [onCanAdd](https://github.com/lekoala/bootstrap5-tags?tab=readme-ov-file#options) event that would accomplish this (returning false means you can't add the thing).  But this works just as well.

app/javascript/controllers/tag_select_controller.js
```diff
@@ -50,4 +55,14 @@ export default class extends Controller {
       this.tagContainer.classList.add("is-changed")
     }
   }
+
+  dedupeValues(values) {
+    const sanitizedVals = values.map((val) => {
+      return val.trim().toLowerCase()
+    })
+    const dedupedVals = [...new Set(sanitizedVals)]
+    return dedupedVals.filter((val) => {
```
cavis, line 64: If there were null/undefined values in here, wouldn't line 61 (`null.trim().toLowerCase()`) already have errored?  Maybe you should do this first?

A simple falsey check also works, since empty strings are false in javascript `values.filter(val => val)`

cavis (approved):
Couple optional nitpicks, but your solution solves the problem effectively.  And I couldn't get the select to add more than 1 tag at a time.  👍

PR #1066, opened 2024-08-06
Reduce PodcastRSSImport config storage

cavis:
PodcastImports were previously saving the entire original RSS string into the `config` store column in the db.

Turns out, when this is several megabytes, we slow way down as each EpisodeImport completes and calls [status_from_episodes!](https://github.com/PRX/feeder.prx.org/blob/main/app/models/podcast_import.rb#L27) (which has a `with_lock` that does a `SELECT *` from the db).

Minimal solution here - save the parsed `channel` and `first_entry` to the config store, but not the original RSS string.

### DIFF
--- a/app/models/imports/podcast_rss_import.rb
+++ b/app/models/imports/podcast_rss_import.rb
@@ -1,7 +1,7 @@
 require "feedjira"
 
 class PodcastRssImport < PodcastImport
-  store :config, accessors: [:episodes_only, :new_episodes_only, :audio, :feed_rss], coder: JSON
+  store :config, accessors: [:episodes_only, :new_episodes_only, :audio, :channel, :first_entry], coder: JSON
 
   has_many :episode_imports, dependent: :destroy, class_name: "EpisodeRssImport", foreign_key: :podcast_import_id
 
@@ -21,7 +21,7 @@ class PodcastRssImport < PodcastImport
   end
 
   def feed_rss
-    config[:feed_rss] ||= http_get(url)
+    @feed_rss ||= http_get(url)
   end
 
   def feed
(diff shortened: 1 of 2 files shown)


app/models/imports/podcast_rss_import.rb
```diff
@@ -132,22 +144,22 @@ def create_or_update_episode_imports!
     end
   end
 
-  def feed_description(feed)
-    result = [feed.itunes_summary, feed.description].find { |d| !d.blank? }
+  def feed_description
+    result = [channel[:itunes_summary], channel[:description]].find { |d| !d.blank? }
```
cavis, line 148: From here on down, I'm just:

1. Changing accessors with hash `:[]` access instead of `Feedjira::Parser::Podcast` methods.
2. Just referencing the `channel` and `first_entry` stores directly rather than passing around this kind of ambiguous `feed` variable.

app/models/imports/podcast_rss_import.rb
```diff
@@ -21,7 +21,7 @@ def file_name
   end
 
   def feed_rss
-    config[:feed_rss] ||= http_get(url)
+    @feed_rss ||= http_get(url)
```
kookster, line 24: the issue, or not issue, with this is that we will get updated rss if this is run over a prolonged time
maybe that's fine, but it means an inconsistent src of episodes and data

kookster (approved):
looks like it will work
perhaps we could later on also get a capture of the rss as a file on s3?
I worry a bit about not having the rss the import is processing

app/models/imports/podcast_rss_import.rb
cavis: So just thinking this dataflow through...

1. Controller tries to create a new PodcastRssImport
2. [validate_rss](https://github.com/PRX/feeder.prx.org/blob/main/app/models/imports/podcast_rss_import.rb#L66) gets called - which downloads/parses the RSS but _doesn't_ persist any of that
3. Later, in an async PodcastImportJob, we call [import!](https://github.com/PRX/feeder.prx.org/blob/main/app/models/imports/podcast_rss_import.rb#L100)
4. This calls `create_or_update_podcast!`, which again memoizes the RSS.  But also stores the `channel` and `first_entry` to the db.
5. Next during the import, we `create_or_update_episode_imports!` ... this also hits the memoized `feed.entries`, and creates an EpisodeImport with the [entry_hash](https://github.com/PRX/feeder.prx.org/blob/main/app/models/imports/podcast_rss_import.rb#L140) for each of those.
6. EpisodeRssImports run (in their own jobs)... but only reference their already persisted `episode_import.config[:entry]` hash.
7. At the end of each episode import, we call back into [PodcastImport.status_from_episodes!](https://github.com/PRX/feeder.prx.org/pull/1066).  Which is still going to run a `SELECT * ... FOR UPDATE`.  But now, there's much less data in that config field, so it should be quite a bit faster.

So maybe not ideal that we GET the RSS twice now?  But they at least happen pretty close together.  Assuming the worker isn't way behind.

PR #1072, opened 2024-08-14
Remove call w/mutation already present in Apple::Publisher

svevang:


app/models/episode.rb
```diff
@@ -250,7 +250,6 @@ def apple_prepare_for_delivery!
   end
 
   def apple_mark_for_reupload!
-    apple_prepare_for_delivery!
```
svevang, line 253: See #904
`Episode#apple_needs_delivery!` marks the episode as needing delivery. And over in `Apple::Publisher` we're [already calling this same method to do the mutation.](https://github.com/PRX/feeder.prx.org/blob/f7e9cda2c85ae852a18b33987477450c77a0ad78/app/models/apple/publisher.rb#L135)

cavis (approved):
👍 looks good to me

PR #1075, opened 2024-08-16
Feat/transcripts UI

### LINKED ISSUE #1060: Transcript upload UI
3rd step of #969.

Feeder _ Episode _ Media Files

- [ ] Create UI to upload the transcript.  Lot of the javascripty parts can be copied from the image/audio uploaders.  Also some views.
- [ ] On save in the controller, set the original_url/file_size (and randomize the guid) on the Transcript model
- [ ] Kick off a copy transcript task in the episode.copy_media hooks
- [ ] Add Rails validations to the transcript that it has an acceptable mime_type (get a generic list of transcript mimes we'll support).  Similar to image/media.

radical-ube:
resolves #1060 

New UI for transcripts uploading at `episode_media_path`. Code is mostly ported as a mix from media upload and image upload elements; code organization and visual should look most similar to an audio/video upload.

functionality testing:
- does a transcript upload work?
- are allowed file types correctly permitted/blocked from upload?
- upon episode save, is the transcript properly built and saved in the db?
- for a complete upload, is the download button working properly?
- for a complete upload, is the delete button working properly?
- does an error display as intended?

@cavis for functional and code review
@brandonhundt for visual/styling/UI review


brandonhundt (commented):
Great start! Initial UI thoughts:

1. I don't think transcripts need to be in their own card especially if its only a single field. I think in its current form users will get a little confused and put their audio in the transcripts field - they seem a little of the same priority. (Aside- long term I think transcripts will be in their own tab)

2. I think we should write up some help text for Transcript's field noting the accepted file formats. I can provide the language.

brandonhundt:
One more thought. I don't think we should support PDFs. That is outside the spec of supported file types here: https://github.com/Podcastindex-org/podcast-namespace/blob/main/transcripts/transcripts.md

app/views/transcripts/_form.html.erb
```diff
@@ -0,0 +1,13 @@
+<% if upload_new?(transcript) %>
+  <% if policy(transcript).create? %>
+    <%= render "transcripts/new", form: form, transcript: transcript %>
+  <% end %>
+<% elsif upload_stalled?(transcript) %>
+  <%= render "transcripts/error", form: form, transcript: transcript, delete_path: delete_path, retry_path: retry_path %>
```
cavis, line 6: If I upload a non-allowed file (like a PNG) ... I'm getting stuck "processing forever", which eventually lands in this stalled/retry state.

That all is fine, but it looks like the issue is:

1. We set [the mime type](https://github.com/PRX/feeder.prx.org/blob/main/app/models/tasks/copy_transcript_task.rb#L36) to the bad image/png
2. We notice that's invalid, and set [status to invalid](https://github.com/PRX/feeder.prx.org/blob/main/app/models/tasks/copy_transcript_task.rb#L39)
3. BUT - unlike on the [Uncut model](https://github.com/PRX/feeder.prx.org/blob/main/app/models/uncut.rb#L2)... our validations are missing that "if status complete" bit.  So the validation still fails, and we can't save the record.

I think this validation just needs the `if:` status check copied over, so we skip it if status is invalid.

config/locales/en.yml
```diff
@@ -1146,3 +1146,21 @@ en:
     title: Not Authorized
     podcast_policy:
       create?: You are not authorized to create a podcast
+  transcripts:
+    error:
+      processing: There was a problem processing your transcript
+    help:
+      transcript: You can upload your transcript here. Current formats allowed are text, pdf, and json.
```
cavis, line 1153: Is this used anywhere?

cavis (changes requested):
Alright, found one bug when uploading a non-allowed file type.  But otherwise this all looks/works good.

I defer to Brandon on the other UI/layout/allowed-files question.  Some of these things are also real easy to update later on.

app/controllers/episode_transcripts_controller.rb
```diff
@@ -0,0 +1,62 @@
...
+    authorize @episode, :update?
+    @episode.assign_attributes(episode_params)
+
+    respond_to do |format|
+      if @episode.save
+        @episode.copy_media
+        format.html { redirect_to episode_transcripts_path(@episode), notice: t(".notice") }
+      elsif @episode.errors.added?(:base, :media_not_ready)
+
+        # some UI feedback that these files aren't ready
```
radical-ube, line 29: i wasn't really sure how this would translate over to transcript? would it just be a line saying `@episode.transcript.valid?`

app/views/episodes/_tabs.html.erb
```diff
@@ -4,6 +4,7 @@
       <%= tab_link_to @episode.title, edit_episode_path(@episode), class: "text-truncate", active: episode_metadata_active? %>
       <%= render "episode_media/tab", episode: @episode %>
       <%= tab_link_to t(".player"), episode_player_path(@episode) %>
+      <%= render "episode_transcripts/tab", episode: @episode %>
```
brandonhundt, line 7: Can you move the transcripts tab ahead of the Embeddable Player but after Media files (3rd in order)?

app/views/episode_transcripts/_form_transcript.html.erb
```diff
@@ -0,0 +1,20 @@
+<div class="prx-uploads col-12 mb-4">
+  <div class="card shadow border-0">
```
brandonhundt, line 2: Since Transcripts are now in a tab, we don't need the field to be in a separate card.

app/views/episode_transcripts/_form_transcript.html.erb
```diff
@@ -0,0 +1,20 @@
...
+    <div class="card-header card-header-info d-flex justify-content-between">
+      <h2 class="card-title h5 flex-grow-1"><%= t(".title") %></h2>
+      <%= help_text t(".help.transcript") %>
```
brandonhundt, line 5: Move this to be the last icon on the field.

config/locales/en.yml
```diff
@@ -708,6 +709,44 @@ en:
...
+  episode_transcripts:
+    form:
+      confirm: Discard unsaved changes?
+    form_status:
+      <<: *form_status
+      title: Transcript Status
+    form_transcript:
+      title: Transcript
+      help:
+        transcript: You can upload your transcript here. Current formats allowed are text and json.
```
brandonhundt, line 721: "Upload a JSON or TXT file containing the text transcript"

(Note to self, once we have a help doc, we should link to it)

brandonhundt:
@radical-ube Since there's not a design file for the tabbed interface, yet, I've given you some inline feedback. I think this will work well and set some solid UX patterns as we add more features in the future.

Also thanks for removing PDF uploads!

app/controllers/episode_transcripts_controller.rb
radical-ube: looking at `episodes_controller`, i'm inclined to just delete this `contents` code as irrelevant to this controller, but keep this `elsif` branch with the `media_not_ready` error messaging.

app/views/episode_transcripts/_tab.html.erb
```diff
@@ -0,0 +1,17 @@
+<%= turbo_frame_tag "episode-transcript-status", data: {controller: "morphdom"} do %>
+  <% status = episode_transcript_status(episode) %>
+
+  <%= tab_link_to episode_transcripts_path(episode), data: {controller: "active-link", turbo_frame: "_top"} do %>
+    <%= t("episodes.tabs.transcript") %>
+    <% unless status == "incomplete" %>
```
radical-ube, line 6: it didn't look right to have an `incomplete` badge on this tab since transcript uploads are optional, but i still render all the other badges.

brandonhundt:
@radical-ube Thanks for the latest changes, it's looking outstanding and working as intended! One last thing I just noticed...

In the Status box, the updated date should be in a column next to the button. We had to change it in the Media Files status box because of some extra buttons and badges, but because those don't exist for this tab, we should use the other style that's present throughout the app. Very minor. Thanks.

**Existing:**
Screenshot 2024-08-27 at 3 27 08 PM

**Expected:**
Screenshot 2024-08-27 at 3 24 37 PM

app/controllers/episode_transcripts_controller.rb
cavis: Yeah, that "media not ready" is a custom error related to publishing.  (You can't publish an episode to the RSS until we've processed its audio files).

I don't think you need that here - it's fine to publish an episode with the transcript not ready yet.

app/models/transcript.rb
```diff
@@ -10,7 +10,7 @@ class Transcript < ApplicationRecord
 
   validates :original_url, presence: true
 
-  validates :mime_type, inclusion: {in: %w[text/plain application/json application/pdf]}, allow_blank: true
+  validates :mime_type, inclusion: {in: %w[text/plain application/json]}, if: :status_complete?, allow_blank: true
```
cavis, line 13: May be worth uploading the [4 example files](https://github.com/Podcastindex-org/podcast-namespace/tree/main/transcripts) in the podcastindex transcript section, to make sure they pass this.  (And to see what mime-types Porter is returning for them).

app/controllers/episode_transcripts_controller.rb
```diff
@@ -0,0 +1,62 @@
...
+    end
+  rescue ActiveRecord::StaleObjectError
+    render :show, status: :conflict
+  end
+
+  private
+
+  def set_episode
+    @episode = Episode.find_by_guid!(params[:episode_id])
+    @episode.strict_validations = true
```
cavis, line 51: So this would cause [validate_media_ready](https://github.com/PRX/feeder.prx.org/blob/main/app/models/episode.rb#L345) to run (it has an `if: :strict_validations` up top).

For this page, I don't think we need to check if the media is ready.

app/views/episode_transcripts/_tab.html.erb
cavis: Hrm... I guess I don't think this status/badge matters at all for transcripts?  At least not on the tab.

But yes, a transcript isn't required.  So an episode isn't "incomplete" if it doesn't have one.  Vs audio, which _is_ required.

app/helpers/episodes_helper.rb
```diff
@@ -67,6 +67,20 @@ def episode_media_status(episode)
...
 
+  def episode_transcript_status(episode)
+    if upload_problem?(episode.transcript)
+      "error"
+    elsif upload_processing?(episode.transcript)
+      "processing"
+    elsif episode.transcript.present?
+      "complete"
+    elsif episode.published_at.present?
+      "incomplete-published"
```
cavis, line 78: I think we can omit these badges for transcripts - it's optional, unlike the media files.

cavis (changes requested):
I think there are a few things copied from media that can be omitted here (the incomplete errors, status badges, etc).

And since you added a controller, I'd also add a `test/controller` with the usual slate of "updates / validates / authorizes" tests.

app/views/episode_transcripts/transcripts/_new.html.erb
```diff
@@ -0,0 +1,91 @@
...
+
+  <%# step 1: file field %>
+  <div class="input-group">
+    <div class="form-floating <%= "d-none" if upload_uploaded?(transcript) %>" data-upload-target="picker">
+      <% cls = "form-control position-absolute opacity-0 z-1" %>
+      <% data = {action: "change->unsaved#change upload#upload", upload_target: "field"} %>
+      <% data[:unsaved_target] = "changed" if transcript.marked_for_destruction? %>
+
+      <%# NOTE: name is nil, so we don't submit the actual file to Rails %>
+      <%= form.file_field :transcript_file, class: cls, accept: ".html,.json,.srt,.vtt", name: nil, data: data %>
```
radical-ube, line 14: is this the correct assumption? that these are the only file types we should expect based on the podcast index examples? or should i have left it as the broader `text/plain` and `application/json`?

cavis: Looks like `accept` works with both mime types and file extensions.  I think this is a good assumption to start - we'll see if anyone complains!

(Browsers also allow you to override this, and select any file you want.  So there is a workaround.)

cavis (approved):
👍 Woo, this looks good to me.  May want to slightly coordinate when you deploy this, since it probably needs some support docs.

PR #1076, opened 2024-08-21
Lock podcast until timestamp

cavis:
We've been getting some large (multiple thousands) RSS imports lately.

After #1066 and #1067, we seem to get through the initial "import" without clogging the workers.

But now, the thousands of Porter callbacks seem to be clogging the publish queue.  Yesterday, I ended up setting `locked = true` on a Podcast and terminating hundreds of `SELECT pg_advisory_lock(1, 9999)` statements in postgres before the workers unblocked (they were processing < 1 job a minute... and maybe just timing out?)

This works around any contention in that publishing queue advisory lock, by delaying unlocking the podcast for 1 minute per 150 episodes.  The `ReleaseEpisodesJob` cron will pick the Podcast up and call `.publish!` once we get past the `podcast.locked_until` timestamp.

### DIFF
--- a/app/models/podcast.rb
+++ b/app/models/podcast.rb
@@ -165,8 +165,12 @@ class Podcast < ApplicationRecord
     self[:locked_until] = val.present? ? "3000-01-01" : nil
   end
 
-  def locked?
+  def locked
     locked_until.present? && locked_until > Time.now
+  end
+
+  def locked?
+    locked
   end
 
   def publish!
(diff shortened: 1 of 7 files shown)


app/models/podcast.rb
```diff
@@ -161,6 +161,18 @@ def categories=(cats)
     self[:categories] = sanitize_categories(cats, false).presence
   end
 
+  def locked=(val)
```
farski, line 164: Would maybe add a comment about why this works the way it does. Seems like it probably never gets used, but if it's going to exist, someone will come across it someday and probably wonder why we have a way to basically like podcasts permanently.

farski (approved):

PR #1082, opened 2024-09-03
Remove buggy parsing method

radical-ube:
removes the unnecessary parsing of the apple key. also fixes token instantiation for apple feeds.

### DIFF
--- a/app/javascript/controllers/apple_key_controller.js
+++ b/app/javascript/controllers/apple_key_controller.js
@@ -41,17 +41,7 @@ export default class extends Controller {
   }
 
   convertKeyToB64(fileText) {
-    const keyText = this.parseKey(fileText)
-    const encoded = btoa(keyText)
+    const encoded = btoa(fileText)
     this.pemTarget.value = encoded
-  }
-
-  parseKey(text) {
-    const strings = text.split("\n")
-    // remove "begin key" and "end key" text
-    strings.shift()
-    strings.pop()
-
-    return strings.join("")
   }
 }
--- a/app/models/apple/key.rb
+++ b/app/models/apple/key.rb
@@ -7,7 +7,7 @@ module Apple
     validates_presence_of :key_pem_b64
 
     validate :provider_id_is_valid, if: :provider_id?
-    validate :key_pem_format, if: :key_pem_b64?
+    validate :ec_key_format, if: :key_pem_b64?
 
     def provider_id_is_valid
       if provider_id.include?("_")
@@ -19,9 +19,9 @@ module Apple
       Base64.decode64(key_pem_b64)
     end
 
-    def key_pem_format
+    def ec_key_format
       unless passes_openssl_validation?
-        errors.add(:key_pem, "key format did not pass OpenSSL validation")
+        errors.add(:key_pem, "ec key format did not pass OpenSSL validation")
       end
     end
 
--- a/app/models/apple/key.rb
+++ b/app/models/apple/key.rb
@@ -7,7 +7,7 @@ module Apple
     validates_presence_of :key_pem_b64
 
     validate :provider_id_is_valid, if: :provider_id?
-    validate :correct_key_pem_format, if: :key_pem_b64?
+    validate :key_pem_format, if: :key_pem_b64?
 
     def provider_id_is_valid
       if provider_id.include?("_")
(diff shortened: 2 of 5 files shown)


app/javascript/controllers/apple_key_controller.js
```diff
@@ -41,17 +41,7 @@ export default class extends Controller {
   }
 
   convertKeyToB64(fileText) {
-    const keyText = this.parseKey(fileText)
-    const encoded = btoa(keyText)
+    const encoded = btoa(fileText)
```
kookster, line 44: Skips getting rid of the top and tail of the file, looks right

kookster (approved):
lgtm

app/models/feeds/apple_subscription.rb
```diff
@@ -14,14 +13,15 @@ class Feeds::AppleSubscription < Feed
   validate :unchanged_defaults
   validate :only_apple_feed
   validate :must_be_private
+  validate :must_have_token
```
svevang, line 16: This reads really nicely and is a good addition!

A couple of questions / just thinking out loud:

In what context do you expect this validation error/flag to be triggered? Something like a future code refactoring?

Assuming we're in the "reject" state for this validation (somehow the apple subscription no longer has a token), how must we test to ensure we surface the validation message?

app/models/apple/key.rb
```diff
@@ -17,5 +18,14 @@ def provider_id_is_valid
...
     end
+
+    def correct_key_pem_format
+      header = "-----BEGIN"
+      footer = "-----END"
+      lines = key_pem.split("\n")
+      unless lines[0].include?(header) && lines[-1].include?(footer)
+        errors.add(:key_pem, "key pem does not have correct header or footer")
+      end
+    end
```
svevang, line 29: Really cool to see. This validation goes beyond what is necessary to fix the original problem and starts to posit the question: "What if someone uploads a bad/corrupted keyfile?".  From a UI perspective, it's good to guard against bad input!

A couple thoughts:

1) The current approach is to blow up. But it's opaque to the user because it happens in a background job -- not good! The user would upload a bad key, then presumably create an episode with some audio, and a background job is kicked off where we [instantiate the key class](https://github.com/PRX/feeder.prx.org/blob/main/app/models/apple/api.rb#L60) and _then_ it blows up. So this validation better. We don't want the user to upload malformed keys.

2) The OpenSSL class [we are using for the ecdsa key](https://github.com/PRX/feeder.prx.org/blob/main/app/models/apple/api.rb#L60), apparently, does some kind of validation on the key. And this (in an engineering sense) is a _surface_ of potential bugs and problems. We take the (potentially corrupted) key and feed it into this black box.  But it seems reasonable to assume that the key format has to be *at least* consistent in order to build up this ECDSA representation in the openssl class, right?

Re: 2,  would be better to wrap the current approach with this validation? So instance of testing the correctness of the PEM format, we use `OpenSSL::PKey::EC.new(...)` to test the key material and rescue from any exceptions?

app/models/feeds/apple_subscription.rb
radical-ube: q1: yes, refactoring is probably the context i would expect to see this. like if suddenly we allowed users to add tokens to apple feeds, and changed around the UI to be more like current private feeds (where the tokens array can be empty), hopefully this validation would catch that. i don't expect this validation to show up in the current UI since users aren't responsible for adding tokens to apple feeds.

q2: i'm not quite sure i understand the question. is this related to q1, like where does this validation show up? if so, then same answer, doesn't need to be surfaced in the current UI; for developer use only, at the moment.

app/models/apple/key.rb
radical-ube: q2: it seems like that is out of scope for what i'm trying to achieve? if i'm understanding the question, the suggestion is to test that the key that _comes out_ of the "black box" of the OpenSSL class method is formatted correctly. but it feels like that's not what this model is for? this `Apple::Key` model holds the information given to it from the user, which is what the `Apple::API` then uses, right?

but i don't think that's what i'm trying to validate here. i'm only validating that this key that _goes into_ the "black box" has the correct format. anything that happens beyond that point feels like it should belong to either `Apple::API` or wherever else that that OpenSSL method is used to create a key. this validation is only based off of [the PEM documentation as it currently exists](https://en.wikipedia.org/wiki/Privacy-Enhanced_Mail), and that's the only thing i'm trying to match to, which is why it's hard-coded in this way.

kookster:
Is this ready to merge, or more changes requested? @svevang @radical-ube

svevang (changes requested):
We don't need to parse / validate the crypto formats here (pem).

Instead leverage the error handling in the crypto's 'black box' to do that. Something like

```
def is_valid_key?(input_key)
  begin
    OpenSSL::PKey::EC.new(input_key)
    true
  rescue
    false
  end
end
```
And use that in the validation method.

svevang (approved):
Looks good!

Naming minor nit: We're testing the overall ec key format, not just the pem wrapper.

PR #1087, opened 2024-09-12
fix(styles): anchor reboot selector

rpeterman-gp:
Change selector for `<a>` tag reboot styles to apply to those not being styled by a class. This will prevent this reset from bleeding into styled anchors like menu item links.

### DIFF
--- a/app/assets/stylesheets/shared/reboot.scss
+++ b/app/assets/stylesheets/shared/reboot.scss
@@ -2,7 +2,7 @@
 // Global Reboot Styles.
 //
 
-a {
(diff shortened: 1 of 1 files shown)


brandonhundt:
@rpeterman-gp This looks good. Can we merge into #1074? A little hover state work is being done there and I think it'd be good to review this against that branch. Thanks!

brandonhundt (approved):
Thanks @rpeterman-gp this is fine. I do think we should maybe revisit hover styles down the road, but we can do that in a new ticket.

PR #1089, opened 2024-09-13
Turn off delegated delivery UI

svevang:


brandonhundt (approved):
LGTM. thanks @svevang

Issue #1090, opened 2024-09-16
Conditional validations not applied on legacy episode metadata blocks Delegated Delivery


svevang:
- Don't delete the data from the episode
- Truncate the description if the feed is going to be read from apple
- On Creation of the apple feed creation, Generate the RSS first as preamble to the delegated delivery
- Wait for the apple poller
- Run the delegated delivery process.

kookster:
somewhat fixed by #1092

PR #1092, opened 2024-09-17
Truncate description for shows with Apple feeds

kookster:
Fixes most of #1090 

The RSS truncate depends there being an apple_config - not necessarily fully publishing yet, just a config exists.
I think that should allow for these legacy shows to regenerate the public RSS with truncated descriptions for Apple to re-ingest before we try to sync episodes. Luckily, probably not that many feeds like this - just need some way to get those old shows working.

- [x] Truncate the podcast/episode descriptions in the RSS feed
- [x] Truncate the podcast/episode descriptions before sending to Apple API


app/helpers/podcasts_helper.rb
```diff
@@ -5,6 +5,23 @@ module PodcastsHelper
 
   RSS_LANGUAGE_CODES = %w[af sq eu be bg ca zh-cn zh-tw hr cs da nl nl-be nl-nl en en-au en-bz en-ca en-ie en-jm en-nz en-ph en-za en-tt en-gb en-us en-zw et fo fi fr fr-be fr-ca fr-fr fr-lu fr-mc fr-ch gl gd de de-at de-de de-li de-lu de-ch el haw hu is in ga it it-it it-ch ja ko mk no pl pt pt-br pt-pt ro ro-mo ro-ro ru ru-mo ru-ru sr sk sl es es-ar es-bo es-cl es-co es-cr es-do es-ec es-sv es-gt es-hn es-mx es-ni es-pa es-py es-pe es-pr es-es es-uy es-ve sv sv-fi sv-se tr uk]
 
+  def feed_description(feed, podcast)
+    desc = feed.description.present? ? feed.description : podcast.description
+    desc ||= ""
+    if podcast.has_apple_config?
+      desc = desc.truncate_bytes(4000, omission: "")
```
svevang, line 12: We have [`Episode#description_with_default`](https://github.com/PRX/feeder.prx.org/blob/main/app/models/episode.rb#L313-L315) which is doing the work of the "other" side of the conditional branches in our validation. I.e. the part that's not encoded in Rails validation rules, but represents non-validated output that is transformed into validated output .  IIRC, Apple _requires_ a description on the episodes, which is why `description_with_default` was added (we don't have that validation on the episode currently?).

Going by the existing pattern,  I'd expect the fallback to `podcast.description` and the truncation of the bytes here to be in a similar `#description_with_default` type of method.

svevang: Does a blank description seem problematic?  Another aspect of the existing `description_with_default` pattern.

app/helpers/podcasts_helper.rb
```diff
@@ -5,6 +5,23 @@ module PodcastsHelper
...
+    end
+    desc
+  end
+
+  def episode_description(episode)
+    desc = episode.description_with_default
+    if episode.podcast.has_apple_config?
+      desc = episode.description.truncate_bytes(4000, omission: "")
```
svevang, line 20: Per other comment, `description_with_default` is the non-validated validated output (also in the apple api call).

kookster: are you thinking we just add the 4k limit to that method?

app/helpers/podcasts_helper.rb
kookster: RSS spec requires a description tag, but not that there be a value. I see some other hosts require a desc, but generally have not.

Good news is I was able to get rid of truncation here as none of our feeds have a desc > 4k bytes, and instead i just left a helper method to handle choosing between the podcast or feed description override.

app/helpers/podcasts_helper.rb
kookster: oh, I have an error here, meant to use the desc `episode.description_with_default` value below, will fix

svevang: Yeah, that's what I was thinking:  To incorporate the transformations into a single method `description_with_default` (that name is drifting out of relevance. `description_for_integrations` ? naming is hard), so that we can couple these "extra" validations together in one place

kookster: ok, added a `episode#description_safe` method that I think captures the spirit of it!
Also DRY'd up the 4000 value to constant while I was at it

app/models/apple/config.rb
```diff
@@ -65,6 +67,10 @@ def self.setup_delegated_delivery(podcast, key: nil, apple_config: nil, apple_sh
       mark_as_delivered!(pub)
     end
 
+    def publish_feeds
+      podcast&.publish!
+    end
```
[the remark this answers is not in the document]
kookster: ok, I can try moving this over there

app/helpers/podcasts_helper.rb
```diff
@@ -5,6 +5,18 @@ module PodcastsHelper
...
+    [feed.description, podcast.description].detect { |d| d.present? } || ""
+  end
+
+  def episode_description(episode)
+    if episode.podcast.has_apple_feed?
+      episode.description_safe
+    else
+      episode.description_with_default
+    end
+  end
```
kookster, line 18: That look better @svevang ?

If there is an apple feed associated with the podcast
    publishes the "safe" description (truncated with fallbacks),
otherwise
    uses the description (with fallbacks, no truncation)

app/models/apple/episode.rb
```diff
@@ -290,7 +290,7 @@ def episode_create_parameters
             guid: guid,
             title: feeder_episode.title,
             originalReleaseDate: feeder_episode.published_at.utc.iso8601,
-            description: feeder_episode.description_with_default,
+            description: feeder_episode.description_safe,
```
kookster, line 293: naming is hard? safe in this case means "safe to use for Apple integration and the public feed", but maybe just "_safe" is ok?

app/models/episode.rb
```diff
@@ -15,6 +15,7 @@ class Episode < ApplicationRecord
   include ReleaseEpisodes
 
   MAX_SEGMENT_COUNT = 10
+  MAX_DESCRIPTION_BYTES = 4000
```
kookster, line 18: Better to use a descriptive constant!

app/models/feed.rb
```diff
@@ -54,6 +54,7 @@ class Feed < ApplicationRecord
   validates :enclosure_prefix, http_url: true
   validates :display_episodes_count, numericality: {only_integer: true, greater_than: 0}, allow_nil: true
   validates :display_full_episodes_count, numericality: {only_integer: true, greater_than: 0}, allow_nil: true
+  validates :description, bytesize: {maximum: Episode::MAX_DESCRIPTION_BYTES}
```
kookster, line 57: As I said in slack, I checked production data, and there are no feed/podcast descriptions over the 4k limit, so is safe to impose retroactively.

app/models/feeds/apple_subscription.rb
```diff
@@ -31,6 +33,10 @@ def self.model_name
     Feed.model_name
   end
 
+  def publish_feeds
+    podcast&.publish!
```
kookster, line 37: As per suggestion, consolidating the "getting ready to publish to Apple" sort of logic in this class.

I wish I could just trigger a publish of the feed without even trying to publish to apple when this first gets created, and before we want to try and sync with apple...maybe I will add a way to do that, but not sure how to decouple those in the current job?

app/models/podcast.rb
```diff
@@ -73,6 +73,10 @@ def apple_config
     end
   end
 
+  def has_apple_feed?
+    feeds.apple.exists?
```
kookster, line 77: So this condition will see if there's an apple feed at all, not checking for the config or publish status, just if there is a feed at all, then truncate the descriptions in the RSS feed

svevang (approved):

PR #1093, opened 2024-09-17
Remove use of crier and cms

kookster:
Remove crier and cms, including api access, in api responses, and in all tests and fixtures


app/representers/api/auth/feed_representer.rb
```diff
@@ -10,7 +10,6 @@ class Api::Auth::FeedRepresenter < Api::BaseRepresenter
   property :subtitle
   property :description
   property :summary
-  property :summary_preview, exec_context: :decorator, if: ->(_o) { represented.summary.blank? }
```
kookster, line 13: This was used when publish.prx.org would allow editing of the summary

app/representers/api/podcast_representer.rb
```diff
@@ -75,11 +74,7 @@ class Api::PodcastRepresenter < Api::BaseRepresenter
     end
   end
 
-  link :series do
-    URI.join(cms_root, represented.prx_uri).to_s if represented.prx_uri
-  end
-
   link :account do
-    URI.join(cms_root, represented.prx_account_uri).to_s if represented.prx_account_uri
+    URI.join(id_root, represented.prx_account_uri).to_s if represented.prx_account_uri
```
kookster, line 78: the prx auth integration uses access to id for looking up account info, so this is more correct

app/models/imports/podcast_rss_import.rb
```diff
@@ -113,7 +113,7 @@ def import!
     end
   rescue => err
     status_error!
-    unlock_podcast!
+    unlock_podcast_later!
```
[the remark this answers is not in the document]
cavis: Ooh, I remember that.  Good catch!

app/views/podcasts/feed_new.xml
```diff
@@ -1,69 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
```
cavis, line 1: Huh, seems like I unintentionally committed this when working on the importer UI.  Whoops!

cavis (approved):
Begone, old code!

PR #1097, opened 2024-09-18
Fix delegated delivery: Dont update episodes for mixed feed release

### LINKED ISSUE #1080: Handle case where episode has hit public feed prior to delegated delivery upload/upload
We currently have a hybrid API/RSS control of episodes. In the case where an episode as already hit the public feed, the Apple rss poller takes over updates to the episode models, and any subsequent updates to the episode via the API returns a 409.

In the case of episode updates, if an episode is returning 409s, that indicates that the episode is already managed by apple.
 We do have some code for handling 404s, https://github.com/PRX/feeder.prx.org/blob/main/app/models/apple/api.rb#L209-L213
We can extend that to "ignore" the 409s and carry on the publishing routine.

- [ ] ignore episode update 409s for already published episodes
- [ ] log out some warnings

svevang:
Closes #1080

### DIFF
--- a/app/models/apple/api.rb
+++ b/app/models/apple/api.rb
@@ -171,8 +171,9 @@ module Apple
       true
     end
 
-    def ok_code(resp, ignore_not_found: false)
-      return true if ignore_not_found && resp.code.to_i == 404
+    def ok_code(resp, ignore_errors: [])
+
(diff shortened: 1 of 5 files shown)


kookster (approved):
Looks great!

PR #1098, opened 2024-09-18
Fix deprecation warnings

cavis:
Fix some deprecation warnings related to the ruby 3.3.5 upgrade.

- Update `parser/current` so erb-lint doesn't warn
- Add some gems explicitly to the Gemfile, since they'll be remove from ruby core in 3.5.  (Updating Rails may fix a few of these later, but this works for now).

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -47,6 +47,7 @@ gem "paranoia"
 gem "sanitize"
 
 # monitoring/logging
+gem "logger"
 gem "lograge"
 gem "newrelic_rpm"
 gem "ougai"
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -80,6 +82,7 @@ group :development, :test do
   gem "debug", platforms: %i[mri mingw x64_mingw]
   gem "dotenv-rails", "~> 2.0"
   gem "erb_lint", require: false
+  gem "ostruct"
```
kookster, line 85: I thought this was already removed - maybe just deprecated before.
Anyway, great!

kookster (approved):
Seems logical, looks fine

PR #1099, opened 2024-09-24
Re-enable turbo on the submit but direct it to the whole page

### LINKED ISSUE #1095: Disable submit button on episode planner after click
Looks like we don't disable/debounce the "Create 99 Drafts" button on the episode planner.  So it's possible to double-click it (or more).  Which is a problem when planning hundreds of episodes, which might take a bit.

radical-ube:
resolves #1095 

This PR resolves the issue of the submit button on the Planner not disabling fully after submit, causing request issues. The solution re-enables turbo on the submit button, but makes it target the whole page (`_top`) instead of just within the context of its turbo frame.

### DIFF
--- a/app/views/podcast_planner/_form_draft.html.erb
+++ b/app/views/podcast_planner/_form_draft.html.erb
@@ -25,7 +25,7 @@
           <%= t(".title.calendar") %>
         </h5>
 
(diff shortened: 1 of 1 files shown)


cavis (approved):
👍 looks/works good to me!

PR #1100, opened 2024-09-25
Hide episode offset field from default feed view

### LINKED ISSUE #1096: No episode-offset-seconds on default feeds
Things could get weird if we allow non-0 values on `default_feed.episode_offset_seconds`.  It's no longer obvious (to a human or to forecasting) when an episode should have dropped.

image

Now that we support _negative_ episode_offset_seconds, we should just hide this field on the default feed.  I don't even know that we need, want a validation?  Up to the implementer I suppose!

radical-ube:
resolves #1096 

This PR hides the Episode Offset field from the default feed view. This is because we don't want a user to have the option of setting an offset for the default feed. The ticket suggests possibly adding a validation, but [one already exists](https://github.com/PRX/feeder.prx.org/blob/d269a3d51c2c976a61232b1909a22eaa31ede776/app/models/feed.rb#L51) and the [testing for it](https://github.com/PRX/feeder.prx.org/blob/d269a3d51c2c976a61232b1909a22eaa31ede776/test/models/feed_test.rb#L194) seems logical and no additions seem to be needed.

### DIFF
--- a/app/views/feeds/_form_main.html.erb
+++ b/app/views/feeds/_form_main.html.erb
@@ -51,13 +51,15 @@
   </div>
 </div>
 
-<div class="col-6 mb-4">
-  <div class="form-floating input-group">
-    <%= form.select :episode_offset_seconds, episode_offset_options, include_blank: true %>
-    <%= form.label :episode_offset_seconds %>
-    <%= field_help_text t(".help.episode_offset_seconds") %>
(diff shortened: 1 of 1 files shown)


cavis (approved):
👍 didn't realize there was already a validation!  But good to hide this anyways then.

PR #1102, opened 2024-09-26
Render processing view for stalled uploads

### LINKED ISSUE #1094: Media processing errors vs timeouts
Right now, the UI doesn't distinguish between generic processing errors and timeouts.

image

Let's make that error message better (more specific) for timeouts!  Especially because we do periodically see these taking longer than whatever I set for the "stalled" time.  (90 seconds?)

radical-ube:
resolves #1094 

This PR solves a "hidden timeout" issue users were experiencing, by rendering a "still processing" message for "stalled" uploads. The solution i chose to solve for this was to opt for rendering the `_processing` partial instead of the `_error` partial for the `upload_stalled?` user conditional flow. I did this because i saw that the only difference between `upload_processing?` and `upload_stalled?` was the `retryable` condition. Therefore, a "stalled" upload should still be processing. A side benefit to using the `_processing` partial is that it continually polls the server, whereas the `_error` partial does not; if a "stalled" upload eventually completed, the `_error` view wouldn't refresh on its own to a `_complete` view, whereas the `_processing` view does.

Within the `_processing` partial, i added the `retry` button that is conditionally present in the `_error` partial, as well as a small (conditional) message underneath the field that notifies the user if the processing is taking longer than expected. This way, we keep the functionality for a user to choose to retry.

@cavis for functional review
@brandonhundt for language review

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -27,6 +27,8 @@ defaults:
     destroy: &destroy
       success: "%{model} deleted successfully."
       failure: "There was a problem deleting the %{model}."
+  uploads:
+    still_processing: &still_processing Processing is taking longer than expected. Thank you for your patience. You can continue waiting or retry.
 
 en:
   # general config
@@ -649,7 +651,7 @@ en:
         title: MP3 Not Provided
       processing:
         hint: Processing
-        still_processing: Processing is taking longer than expected. Thank you for waiting.
+        still_processing: *still_processing
     segmenter:
       modal:
         label:
@@ -732,6 +734,7 @@ en:
         <<: *transcript_labels
         loading: Loading...
         processing: Processing
+        still_processing: *still_processing
       complete:
         <<: *transcript_labels
       error:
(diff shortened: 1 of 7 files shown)


config/locales/en.yml
```diff
@@ -649,6 +649,7 @@ en:
         title: MP3 Not Provided
       processing:
         hint: Processing
+        still_processing: Processing is taking longer than expected. Thank you for waiting.
```
cavis, line 652: Nice, I do like how this isn't highlighted in red now.   Looks less angry:

image

I might prompt them that it's cool to click the retry button though... "Processing is taking longer than expected - you can continue waiting or retry"

cavis (changes requested):
Makes good sense to me!

Searching through the code, it looks like Transcripts and Images both use basically identical retry logic to this.  Can we port this to those places as well?

config/locales/en.yml
brandonhundt: FWIW: I prefer the term "patience" over "waiting", ie: "thank you for your patience"... Throw in a llittle "Ope" and we might even get the user to chuckle in another wise frustrating moment.

I suggest:
"Ope, file processing is taking longer than expected - you can continue waiting or retry"

cavis (approved):
:+3: nice.  The spinner being active really conveys that processing stalled, and it's not as angry as landing in error state.

PR #1104, opened 2024-09-29
Add podcasts guid, set by uuidv5, add to rss

### LINKED ISSUE #1110: Introduce a guid to use for identifying a podcast
support the podcasting 2.0 `podcasts:guid` in the rss feed

https://github.com/Podcastindex-org/podcast-namespace/blob/main/docs/1.0.md#guid

and this would be a less predicatable/seqeuntual id to use in public - perhaps for use in integrations with other systems

kookster:
fixes #1110

### DIFF
--- a/db/migrate/20240928150204_add_guid_to_podcasts.rb
+++ b/db/migrate/20240928150204_add_guid_to_podcasts.rb
@@ -1,0 +1,6 @@
+class AddGuidToPodcasts < ActiveRecord::Migration[7.2]
+  def change
+    add_column :podcasts, :guid, :string
+    add_index :podcasts, :guid
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -228,12 +228,12 @@ ActiveRecord::Schema[7.2].define(version: 2024_09_28_150204) do
     t.datetime "updated_at", precision: nil, null: false
     t.string "enclosure_prefix"
     t.string "enclosure_template"
-    t.text "exclude_tags"
     t.text "subtitle"
     t.text "description"
     t.text "summary"
     t.boolean "include_podcast_value", default: true
     t.boolean "include_donation_url", default: true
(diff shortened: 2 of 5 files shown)


db/schema.rb
```diff
@@ -443,8 +444,8 @@
   end
 
   create_table "tasks", id: :serial, force: :cascade do |t|
-    t.integer "owner_id"
     t.string "owner_type"
+    t.integer "owner_id"
```
cavis, line 448: Hrm, couple of unrelated changes in here.  Are you working off a prod backup?

db/migrate/20240928150204_add_guid_to_podcasts.rb
```diff
@@ -0,0 +1,6 @@
+class AddGuidToPodcasts < ActiveRecord::Migration[7.2]
+  def change
+    add_column :podcasts, :guid, :string
```
cavis, line 3: Any reason not to just generate these for all existing Podcasts?

cavis (approved):
👍 looks good to me.  I do wonder if we shouldn't just add these to all existing podcasts.

db/schema.rb
kookster: it just changed the order of the columns, not the type or anything - is that a problem?

kookster: (I was not working off a copy of prod when I ran these migrations.)

cavis: Not a problem - it'll just flicker back the next time someone runs a migration locally.

db/migrate/20240928150204_add_guid_to_podcasts.rb
kookster: Yes, and we could do that at the console at any time, I just didn't force it in this PR

cavis (approved):
👍

PR #1112, opened 2024-10-01
Megaphone API publishing integration

kookster:
- [x] Add a new feed type for megaphone
- [x] Save credentials to config
- [x] implement api adapter using faraday for the http library
- [x] megaphone podcast model, retrieving
- [x] megaphone podcast model, persisting
- [x] megaphone episode model, retrieving
- [x] megaphone episode model, persisting
- [x] handle audio changes, with resetting timings
- [x] integrate with callbacks and events to keep in sync?
- [x] Add UI for megaphone feed and config
- [x] Add UI for megaphone episode status
- [x] Update enclosure in feed
- [x] Bug fixes
- [x] More testing of audio replacement and retries


kookster:
Marking this as ready for review, but there is more work to be done, so not quite yet.

app/helpers/feeder_form_builder.rb
```diff
@@ -33,9 +33,19 @@ def text_field(method, options = {})
     add_blank_action(options)
     add_changed(method, options)
     add_disabled(options)
+    redact_value(method, options)
     super(method, options)
   end
 
+  def redact_value(method, options)
```
cavis, line 40: Ooh nice, a form helper!

app/jobs/publish_feed_job.rb
```diff
@@ -19,11 +18,17 @@ def perform(podcast, pub_item)
...
     return :null if null_publishing_item?(podcast, pub_item)
     return :mismatched if mismatched_publishing_item?(podcast, pub_item)
+
     Rails.logger.info("Starting publishing pipeline via PublishFeedJob", {podcast_id: podcast.id, publishing_queue_item_id: pub_item.id})
 
     PublishingPipelineState.start!(podcast)
-    podcast.feeds.each { |feed| publish_apple(podcast, feed) }
+
+    # Publish each integration for each feed (e.g. apple, megaphone)
+    podcast.feeds.each { |feed| publish_integration(podcast, feed) }
```
cavis, line 27: Any dependencies on ordering here?  (Apple before megaphone, or vice versa?)

db/migrate/20240929210828_create_megaphone_configs.rb
```diff
@@ -0,0 +1,13 @@
+class CreateMegaphoneConfigs < ActiveRecord::Migration[7.2]
+  def change
+    create_table :megaphone_configs do |t|
+      t.string :token
+      t.string :network_id
+      t.string :network_name
+      t.boolean "publish_enabled", default: false, null: false
+      t.bigint "feed_id", null: false
```
cavis, line 8: I imagine we might want to consolidate these configs to a AR `store` field(s) at some point?  But... not much cost to having 2 tables for the 2 main integrations at the moment..

env-example
```diff
@@ -80,3 +80,8 @@ APPLE_API_BRIDGE_URL=
 # slack notifications
 SLACK_SNS_TOPIC=
 SLACK_CHANNEL_ID=
+
+# encrypt active record attributes
+ACTIVE_RECORD_ENCRYPTION_PRIMARY_KEY=0123456789abcdefghijklmnopqrstuv
+ACTIVE_RECORD_ENCRYPTION_DETERMINISTIC_KEY=abcdefghijklmnopqrstuv0123456789
+ACTIVE_RECORD_ENCRYPTION_KEY_DERIVATION_SALT=abcdefghijk0123456789lmnopqrstuv
```
cavis, line 87: DB encryption ftw!

app/workers/application_worker.rb
```diff
@@ -1,21 +1,12 @@
...
-  include PrxAccess
   include Shoryuken::Worker
 
-  attr_accessor :message
-
   def self.prefix_name(name)
     [prefix, application, name].join("_")
   end
 
-  def self.announce_queues(model, actions)
```
cavis, line 13: Oh wow, there was still `Announce` stuff around!

app/views/podcast_planner/_form_draft_settings.html.erb
```diff
@@ -34,7 +34,7 @@
   <div class="container">
     <div class="row">
       <% feeds = @podcast.feeds.tab_order %>
-      <% defaults = feeds.filter_map { |f| f.id if f.default? || f.apple? } %>
+      <% defaults = feeds.filter_map { |f| f.id if f.default? || f.integration_type } %>
```
cavis, line 37: Okay, so by default... episodes go to the default feed, and any integration feeds.

app/views/feeds/_tabs.html.erb
```diff
@@ -12,12 +12,17 @@
     <div class="fixed-bottom col-sm-12 col-md-3 col-xl-2">
       <div class="btn-group dropup d-flex sticky-md-bottom m-4 shadow">
         <%= link_to "Add a Feed", new_podcast_feed_path(podcast), class: "btn btn-success flex-grow-1" %>
-        <% if @feeds.none?(&:apple?) %>
+        <% if @feeds.none?(&:integration_type) %>
```
cavis: This will need to change - if you have an apple feed, you can't add a megaphone one etc.

But for now, I think we should just hide the Megaphone button until we're ready to go live.

app/views/episodes/_form_status.html.erb
```diff
@@ -47,6 +47,22 @@
         </p>
       </div>
     <% end %>
+    <% if episode.persisted? && episode.publish_to_integration?(:megaphone) %>
+      <% integration = :megaphone %>
+      <% integration_status = episode_integration_status(integration, episode) %>
+      <div class="col-12 mt-4">
+        <p class="status-text">
+          <strong><%= "#{integration.to_s.titleize} Status" %>:</strong>
```
cavis, line 55: Nitpick: why not just translate the whole thing?  It looks like you were trying to make this generic for any integration, but didn't update the Apple one above this.

app/policies/apple/key_policy.rb
```diff
@@ -0,0 +1,17 @@
+class Apple::KeyPolicy < ApplicationPolicy
+  def new?
```
cavis, line 2: Huh - out of curiousity, why wasn't this needed before?

app/models/prx/api.rb
```diff
@@ -0,0 +1,144 @@
+module Prx
+  module Api
+    class PrxHyperResource < HyperResource
```
cavis, line 3: Also, we never used the prx_access gem in this repo (which I'm pretty sure is where it started life?)  Maybe we'll port it back out at some point.

app/models/feeds/megaphone_feed.rb
```diff
@@ -0,0 +1,61 @@
...
+  after_initialize :set_defaults
+
+  alias_method :config, :megaphone_config
+
+  accepts_nested_attributes_for :megaphone_config, allow_destroy: true, reject_if: :all_blank
+
+  def paranoia_destroy_attributes
+    {
+      deleted_at: current_time_from_proper_timezone,
+      slug: "#{slug}-#{Time.now.to_i}"
```
cavis, line 17: What's this for?  Just to obfuscate any existing copies of the url in the wild?

I wrote an `UnlinkJob` thing to remove stuff from S3 when the slug changes.  Wonder if we should also do that on feed destroy?

cavis: Oh, maybe this is unrelated.  Just doing the same thing as Apple - keep the `slug` available.  The feed base class should probably do that too - someday!

cavis (approved):
Some questions, but i think with (A) removing the "add megaphone" button, and (B) some QA in staging on the apple stuff, I'm a +1.

app/models/megaphone/publisher.rb
```diff
@@ -0,0 +1,133 @@
...
+
+        if !ep.episode_delivery_status(:megaphone).delivered?
+          remaining << ep
+        end
+      end
+    end
+
+    def create_and_update_episodes!
+      megaphone_episodes = []
+      private_feed.episodes.unfinished(:megaphone).each do |ep|
```
svevang, line 87: Do you see the megaphone integration as operating independently from the logic in this [abstract episode set operations concern](https://github.com/PRX/feeder.prx.org/pull/1112/files#diff-92e7117a8391a2f8207842a2c3ac800f2d62db2e82150c9270b95426f5d5da66)?

With the approach here, we have a simpler/compact interface to the feed's episodes, with no shared episode set concerns between the two existing integrations.  But we're lacking visibility into deleted episodes, sorting properties etc. Does that matter?

svevang (approved):

app/jobs/publish_feed_job.rb
kookster: no, that shouldn't matter, but it does do all integrations before RSS publishing, so if it blocks on RSS that is last

app/models/megaphone/publisher.rb
kookster: The episode set operations are used in the base publisher in a few ways - 
`episodes_to_sync` doesn't work for megaphone, as it filters to only published episodes.
```
      def episodes_to_sync
        filter_episodes_to_sync(show.episodes)
      end
```
it is also wrapped with `filter_episodes_to_sync`, which after querying and instantiating all the episodes rejects from the array based on delivery status, whereas the `unfinished` method starts with only episodes in need of delivery based on delivery status, which is more efficient, and includes the draft episodes we need to create/update on megaphone.

The archive episodes logic looks closer to what is needed for megaphone, so that method might be useful when that gets implemented.

app/views/feeds/_tabs.html.erb
kookster: hidden!

app/models/megaphone/publisher.rb
kookster: I'll also mention, the actual filter for deciding which episodes should be included in the sync is not really using the episode set operations - in the `show` it's getting the list of `feed_ready` episodes - so this is not useful for the megaphone integration as a super class method, and the querying to get the episode ids to filter the episodes from the episode set operation does not seem like a better way to do this - I'd rather use scopes and do all the filtering as part of the sql query:

https://github.com/PRX/feeder.prx.org/blob/feat/megaphone/app/models/integrations/base/show.rb#L31

app/policies/apple/key_policy.rb
kookster: yeah, I forget when this popped up, but I believe it was when I was testing the Apple feed config UI...

app/models/feeds/megaphone_feed.rb
kookster: yeah the slug has to be unique, so we have to move it out of the way when logically deleting it

app/models/prx/api.rb
kookster: yeah, this PR has enough refactoring ;)

app/helpers/feeder_form_builder.rb
kookster: seemed like the right place to do this? so far so good.

app/models/prx/api.rb
kookster: added #1207  for this

app/views/episodes/_form_status.html.erb
kookster: yeah, I was trying not to mess too much with Apple code when I didn't need to, but put in what I thought would be a way to make this code reusable for both used only for megaphone.
I guess I could go ahead and refactor....

app/views/podcast_planner/_form_draft_settings.html.erb
kookster: yes, that's how apple worked already, and it also matches the requirements for the megaphone integration

app/workers/application_worker.rb
kookster: not much! but when I ran across this I figured I should cleanse it

db/migrate/20240929210828_create_megaphone_configs.rb
kookster: yeah they were different enough I didn't want to tackle trying to make this polymorphic, but I agree, we could end up there.

env-example
kookster: which is a good reminder I need to deploy values for these to the Infrastructure repo....

app/models/megaphone/episode.rb
```diff
@@ -125,6 +125,22 @@ def update!(episode = nil)
       self
     end
 
+    def delete!
+      self.api_response = api.delete("podcasts/#{podcast.id}/episodes/#{id}")
+      delete_sync_log
+      delete_delivery_status
+      self
+    end
```
kookster, line 133: when deleting from megaphone, also need to delete the sync log (otherwise it holds on to the guid, and tries to update a deleted episode), and the delivery status (so it will stop trying to sync when there is nothing in megaphone to sync to)

app/models/megaphone/episode.rb
```diff
@@ -183,7 +199,7 @@ def cuepoints_batch!(cuepoints)
       # validate all the cuepoints about to be created
       cuepoints.all? { |cp| cp.validate!(:create) }
       body = cuepoints.map { |cp| cp.as_json_for_create }
-      self.api_response = api.put("podcasts/#{podcast.id}/episodes/#{id}/cuepoints_batch", body)
+      self.api_response = api.put_base("episodes/#{id}/cuepoints_batch", body)
```
kookster, line 202: cuepoints for some reason don't include the network in the path, unlike ever other endpoint in the API!

app/models/megaphone/episode.rb
```diff
@@ -303,12 +324,12 @@ def get_media_info(enclosure)
         media_version: nil,
         location: nil,
         size: nil
-      }
+      }.with_indifferent_access
```
kookster, line 327: bug fix!

app/models/megaphone/episode.rb
```diff
@@ -319,13 +340,14 @@ def get_media_info(enclosure)
       raise err
     end
 
-    def arrangement_version_url(location, media_version, count)
+    def arrangement_version_url(location, media_version)
```
kookster, line 343: we don't want the request count in the file name - we really want the media version id to match against

app/models/megaphone/publisher.rb
```diff
@@ -24,7 +26,7 @@ def publish!
     def sync_episodes!
       Rails.logger.tagged("Megaphone::Publisher#sync_episodes!") do
         # delete or unpublish episodes we aren't including in the feed anymore
-        unpublish_and_delete_episodes!
+        delete_episodes!
```
kookster, line 29: megaphone doesn't have an "archived" state - we could unpublish and leave a draft, but that's problematic as well, as those get used for inventory (esp for future episodes).

app/models/megaphone/publisher.rb
```diff
@@ -101,7 +101,16 @@ def create_episode!(megaphone_podcast, ep)
       me.create!
     end
 
-    def unpublish_and_delete_episodes!
+    def delete_episodes!
+      megaphone_episodes = []
+      podcast.episodes.with_deleted.where.not(id: private_feed.episodes).each do |ep|
```
svevang, line 106: It looks like this is skipping the HABTM relation for episodes <-> feeds.  It that right? Is there a integration specific reason?

svevang: Also, in that vein, do we need `private_feed.episodes.with_deleted` here?

kookster: it is looking at episodes in the podcast, not just the feed, to find episodes removed from the Megaphone feed, which indicates they should be deleted from Megaphone. Those episodes aren't deleted from feeder, but should be deleted from megaphone.

I think there is a case I am missing here though, which is episodes still in the Megaphone feed which have been deleted, I'll add that!

cavis (approved):
:+4:

PR #1114, opened 2024-10-02
Deprecate summary and content attributes, and fix description, summary, content text formats in the RSS feeds

### LINKED ISSUE #1030: Problems with `<itunes:summary\>`, `<content:encoded\>`, `<description\>` and apple podcasts preview pages
Apple renders the podcast preview pages using a different set of rules for the description text/html than in the apps.

The native apps now seem to only look at `<description/>`, and allow some html including `<p>` tags for new lines.
https://help.apple.com/itc/podcasts_connect/#/itcb54353390

It says just these tags, but it also seems to respect `<br>` tags:
```
description is text containing one or more sentences describing your episode to potential listeners. You can specify up to 4000 bytes. You can use rich text formatting and some HTML (<p>, <ol>, <ul>, <li>, <a>) if wrapped in the <CDATA> tag.
```

We also have fields in the episodes db table for `episodes.content` (`<content:encoded \>`) and `episodes.summary` (`<itunes:summary \>`), but I don't think the new feeder UI sets those.
However, we have old values in there, from when publish would set them and they still come in on rss import.

I also see in the ui, that there are hidden fields that will set episode `content` and `summary` to nil whenever an episode is saved.
This seems to indicate we want to move away from those, and it will cause any `<content:encoded />`  to stop being included once that is set to `nil`, but the rss builder will still set an `<itunes:summary/>` based on a sanitized `description`.

This further is an issue because the sanitization procedure replaces `<br\>` and `<p\>` tags but doesn't convert them to spaces or newlines, so it causes words to get concatenated that were previously separated by the whitespace form those tags.

So there are a few things we could do - 

1. get rid of displaying `<content:encoded>` and `<itunes:summary>` entirely - and just rely on `<description>`.

2. add back `<content:encoded>` but have it always identical to description, this will then take precedence over summary if that value remains 

3. if we keep summary, we need to add content encoded, and we need to fix the white space issue with the sanitizer

kookster:
fixes #1030 using option 3 from that ticket

### DIFF
--- a/app/models/concerns/import_utils.rb
+++ b/app/models/concerns/import_utils.rb
@@ -2,8 +2,10 @@ require "active_support/concern"
 require "prx_access"
 require "net/http"
 require "uri"
+require "text_sanitizer"
 
 module ImportUtils
+  include TextSanitizer
   extend ActiveSupport::Concern
 
   class HttpError < StandardError
@@ -110,9 +112,7 @@ module ImportUtils
   end
 
   def sanitize_html(text)
-    return nil if text.blank?
-    sanitizer = Rails::Html::WhiteListSanitizer.new
-    sanitizer.sanitize(Loofah.fragment(text).scrub!(:prune).to_s).strip
+    sanitize_white_list(text)
   end
 
   def files_match?(file, url)
--- a/app/models/concerns/text_sanitizer.rb
+++ b/app/models/concerns/text_sanitizer.rb
@@ -8,7 +8,7 @@ module TextSanitizer
     return nil if text.blank?
     sanitizer = Rails::Html::WhiteListSanitizer.new
     text = sanitizer.sanitize(Loofah.fragment(text).scrub!(:prune).to_s)
-    cleanup_text(text)
+    clean_whitespace(text)
   end
 
   def sanitize_links_only(text)
@@ -18,14 +18,14 @@ module TextSanitizer
     scrubber.attributes = %w[href target nofollow]
     text = add_newlines_to_tags(text)
     text = Loofah.fragment(text).scrub!(:prune).scrub!(scrubber).to_s
-    cleanup_text(text)
+    clean_whitespace(text)
   end
 
   def sanitize_text_only(text)
     return nil if text.blank?
     text = add_newlines_to_tags(text)
     text = Loofah.fragment(text).scrub!(:prune).text(encode_special_chars: false)
-    cleanup_text(text)
+    clean_whitespace(text)
   end
 
   def sanitize_categories(kws, strict)
@@ -52,7 +52,7 @@ module TextSanitizer
       .gsub(/<br\s*\/>/i, "\n<br>")
   end
 
-  def cleanup_text(text)
+  def clean_whitespace(text)
     text.tr("\r", "\n")
       .gsub(/\s*\n\s*/, "\n")
       .squeeze(" ")
--- a/app/models/concerns/text_sanitizer.rb
+++ b/app/models/concerns/text_sanitizer.rb
@@ -53,9 +53,10 @@ module TextSanitizer
   end
 
   def clean_whitespace(text)
-    text.tr("\r", "\n")
-      .gsub(/\s*\n\s*/, "\n")
-      .squeeze(" ")
-      .strip
+    text.gsub(/\R/, "\n") # if there is any kind of newline, make it \n
+      .gsub(/([ \t]*)\n([ \t]*)/, "\n") # get rid of whitespace on eother side of a newline
+      .gsub(/\n\n\n/, "\n\n") # no more than 2 newlines together
+      .squeeze(" ") # make all groups of spaces into a siingle space
+      .strip # get rid of leading or trailing spaces
   end
 end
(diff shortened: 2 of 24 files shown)


app/models/concerns/import_utils.rb
```diff
@@ -110,9 +112,7 @@ def remove_podcastchoices_link(str)
   end
 
   def sanitize_html(text)
-    return nil if text.blank?
-    sanitizer = Rails::Html::WhiteListSanitizer.new
-    sanitizer.sanitize(Loofah.fragment(text).scrub!(:prune).to_s).strip
+    sanitize_white_list(text)
```
kookster, line 115: might as well just use this so it is consistent

app/models/concerns/text_sanitizer.rb
```diff
@@ -37,4 +42,20 @@ def sanitize_category(kw, max_length, strict)
...
+      .gsub(/<\/p>/i, "</p>\n")
+      .gsub(/<div>/i, "\n<div>")
+      .gsub(/<\/div>/i, "</div>\n")
+      .gsub(/<br>/i, "\n<br>")
+      .gsub(/<br\s*\/>/i, "\n<br>")
+  end
+
+  def clean_whitespace(text)
+    text.tr("\r", "\n")
+      .gsub(/\s*\n\s*/, "\n")
```
kookster, line 57: the goal here is to end up with a single newline, even if there were multiple p or br tags stacked
arguably, we could support multiple newlines...that would be another way to go, but I don't see a good reason for it in the text versions of these descriptions

app/models/imports/episode_rss_import.rb
```diff
@@ -110,7 +107,7 @@ def entry_description(entry)
   end
 
   def entry_description_attribute(entry)
-    [:content, :itunes_summary, :description, :title].find { |d| !entry[d].blank? }
+    [:content, :description, :itunes_summary, :itunes_subtitle, :title].find { |d| !entry[d].blank? }
```
kookster, line 110: content and description consistently have the full html version much more often than summary

app/views/podcasts/show.rss.builder
```diff
@@ -155,7 +155,7 @@ xml.rss "xmlns:atom" => "http://www.w3.org/2005/Atom",
               url: ep.enclosure_url(@feed))
           end
 
-          xml.content(:encoded) { xml.cdata!(ep.content) } unless ep.content.blank?
+          xml.content(:encoded) { xml.cdata!(episode_description(ep)) }
```
cavis, line 158: So now `<content:encoded>` will be identical to `<description>` - do we really need both?

app/models/concerns/text_sanitizer.rb
```diff
@@ -37,4 +42,21 @@ def sanitize_category(kw, max_length, strict)
...
     end
   end
+
+  def add_newlines_to_tags(text)
+    text.gsub(/<p>/i, "\n<p>")
+      .gsub(/<\/p>/i, "</p>\n")
+      .gsub(/<div>/i, "\n<div>")
+      .gsub(/<\/div>/i, "</div>\n")
+      .gsub(/<br>/i, "\n<br>")
+      .gsub(/<br\s*\/>/i, "\n<br>")
```
cavis, line 52: Should these check if the tags already have newlines?  Or do we scrub those out before calling this?

app/models/concerns/text_sanitizer.rb
```diff
@@ -37,4 +42,21 @@ def sanitize_category(kw, max_length, strict)
...
+  end
+
+  def clean_whitespace(text)
+    text.gsub(/\R/, "\n") # if there is any kind of newline, make it \n
+      .gsub(/([ \t]*)\n([ \t]*)/, "\n") # get rid of whitespace on eother side of a newline
+      .gsub(/\n\n\n/, "\n\n") # no more than 2 newlines together
```
cavis, line 58: Ah, maybe this covers my previous questions... even if we had too many newlines, this should remove some of them.

Are these rules from some real-world cases you've seen, or just from the test?

I also wonder if the last steps (trimming/combining spaces) should happen before the multiple-newline detection.  But this is probably fine.

cavis (approved):
Couple questions, but this looks good to me!

:+4:

app/models/concerns/text_sanitizer.rb
kookster: it dedupes the newlines anyway, so adding another is always fine.

app/models/concerns/text_sanitizer.rb
kookster: the `strip` could be before or after I guess, doesn't seem to matter much.
the prior regex gets rid of spaces near newlines the final squeeze is any place with multiple spaces, so again different, and could be any order

app/views/podcasts/show.rss.builder
kookster: I see multiple rss hosts that do provide both

app/models/concerns/text_sanitizer.rb
cavis: Makes good sense!

app/models/concerns/text_sanitizer.rb
cavis: +1

cavis (commented):
:+onwards!:

PR #1115, opened 2024-10-03
Feat/transcripts rss link

### LINKED ISSUE #1061: Transcript link in RSS
Fourth step of #969.

Add transcript link to the RSS.  I forget exactly what tag it is right now... but it's a thing!

radical-ube:
resolves #1061 

This PR adds the `<podcast:transcript>` tag to the episode RSS builder. To be within the PodcastIndex standards, `url` and `type` are required attributes and are set in the RSS builder. `url` is based on the `transcript.url` and `type` is mapped from the `transcript.format` to one of the valid `mime_type`s. To allow a user to set a `transcript.format`, a new dropdown was added into the transcript UI to select one of the four transcript types. There is potentially user conflict here: currently, the format selection doesn't have to match the upload format. ideally, there would be some sort of matching of the file's detected mime_type to auto-select a format, but that's not what Porter currently returns to us. could possibly implement javascript that selects a format based on file extension? but i don't know how critical or important it is to do this at this time.


app/models/transcript.rb
```diff
@@ -10,9 +17,12 @@ class Transcript < ApplicationRecord
 
   validates :original_url, presence: true
 
-  validates :mime_type, inclusion: {in: %w[text/plain application/json]}, if: :status_complete?, allow_blank: true
+  validates :mime_type, inclusion: {in: MIME_TYPES.values}, if: :status_complete?, allow_blank: true
+
+  validates :format, presence: true
 
   enum :status, [:started, :created, :processing, :complete, :error, :retrying, :cancelled, :invalid], prefix: true
+  enum :format, MIME_TYPES, prefix: true, default: :json
```
cavis, line 25: Hmm... I see where you're coming from here (this is nice and DRY code).

But what Rails is actually doing is storing the string `text/html` in the database.  And translating that to `html` in the model.

I can imagine that causing problems down the road (imagine we add a 2nd `format` that _just happens_ to also have the mime-type `text/html`).

So I think we want to store the literal string `html` in the database here?  So the hash we pass in should be `{html: "html"}` ... not `{html: "text/html"}` here.

app/views/episode_transcripts/_form_main.html.erb
```diff
@@ -0,0 +1,9 @@
+<%= form.fields_for :transcript, episode.transcript, include_id: false do |fields| %>
+  <div class="col-12 mb-4">
+    <div class="form-floating input-group">
+      <%= fields.select :format, episode_transcript_options, {} %>
```
cavis, line 4: So for an episode

cavis (changes requested):
One gotcha on enums, and a thing about changed-fields.  But this is looking/working good!

app/views/episode_transcripts/_form_main.html.erb
radical-ube: resolved in `_form_transcript` below, since that is where a Transcript is first built (if new).

cavis (approved):
:+3: onwards!

Issue #1016, opened 2024-05-06
Escalate publish pipeline failures


cavis:
We've hit cases where failed publish attempts aren't loud enough.  Or are too loud (we expect _some_ failures/retries, uploading to Apple).

We should escalate errors based on how many / how long we've been failing.

- [ ] For <= 5 or something, just ops-warn
- [ ] For > 5, ops-error
- [ ] For > some number (hours?) ... ops-fatal

PR #1116, opened 2024-10-03
Asset processing count

svevang:
Re: https://github.com/PRX/feeder.prx.org/issues/1016

This sets up a count on the episodes that are waiting for asset state == 'PENDING'. For each round trip of the publish feed job, this increments a count on the apple delivery status model


app/controllers/episodes_controller.rb
```diff
@@ -92,7 +92,7 @@ def destroy
 
     respond_to do |format|
       if @episode.destroy
-        @episode.publish!
+        @episode.podcast.publish!
```
cavis, line 95: Are you trying to skip the `apple_mark_for_reupload!` when the episode is deleted?  It might be better to add a `unless deleted?` check to that episode method then, in case someone calls `episode.publish!` on a deleted episode from somewhere else in the codebase.

app/models/apple/episode.rb
```diff
@@ -491,7 +491,7 @@ def audio_asset_state_success?
     end
 
     def has_media_version?
-      return false unless delivery_status.present? && delivery_status.source_media_version_id.present?
+      return false unless delivery_status.source_media_version_id.present?
```
cavis, line 494: Hrm... could `delivery_status` be nil?  Looks like that aliases to a has_many.first on the Episode.  Maybe a `&.` safe nav?

app/models/apple/episode_delivery_status.rb
```diff
@@ -1,5 +1,21 @@
...
+    def self.update_status(episode, attrs)
+      new_status = (episode.apple_episode_delivery_status&.dup || new(episode: episode))
+      new_status.assign_attributes(attrs)
+      new_status.save!
+      episode.apple_episode_delivery_statuses.reset
+      new_status
+    end
+
+    def increment_asset_wait
+      self.class.update_status(episode, asset_processing_attempts: (asset_processing_attempts || 0) + 1)
```
cavis, line 14: Ah, so the "asset wait" is the number of processing attempts we've made on the episode thus far.  Or ... the latest version of this episode's media, i suppose.

test/models/apple/publisher_test.rb
```diff
@@ -369,6 +405,70 @@
...
+          error = assert_raises(RuntimeError) do
+            publisher.wait_for_asset_state(eps)
+          end
+          assert_equal "Timed out waiting for asset state", error.message
+        end
+      end
+
+      log = logs[0]
+      assert_equal "Timed out waiting for asset state", log[:msg]
+      assert_equal 30, log[:level]
```
cavis, line 466: Oh, so this PR is already downgrading these to log.infos?  Where is that happening?

Hrm... not sure I'm seeing it.  Is it a side effect of a change in here?

cavis (commented):
Couple of questions, but overall this is looking good to me!

Not sure I'm completely following how the logs were downgraded to infos already (in the tests).  I may just be misunderstanding how those are handled in the outer worker tasks.

app/models/apple/episode.rb
svevang: This was one of the changes in this PR, to set up a default status.

So when querying the status, the default is set up:
https://github.com/PRX/feeder.prx.org/pull/1116/files#diff-4ced10ac3e0d2a50f776b961b3f61a72a3ba92bf4dd67664942cdefe0b72f2f4R28-R33

and when updating the status, the default is provided:
https://github.com/PRX/feeder.prx.org/pull/1116/files#diff-932681036f4ca05556eefe956f92cff8ad8bc433cda926d65d22a701b8c5f050R6

The idea is that we can move methods from the episode model over onto this status model (provided we initialize a default).  Ultimately, this lets us use instance methods over on the status model (like `def increment_asset_wait`).

app/models/apple/episode_delivery_status.rb
svevang: The media version increments are tracked in another column `source_media_version_id`. e.g. if you group on the episode id you can track the media versions that the episode have gone through:

```
SELECT count(source_media_version_id) count, episode_id
FROM (
        SELECT DISTINCT source_media_version_id,
                        episode_id
        FROM apple_episode_delivery_statuses
        WHERE source_media_version_id IS NOT NULL
) q
GROUP BY episode_id
HAVING count(source_media_version_id) > 1
ORDER BY count DESC
```

app/controllers/episodes_controller.rb
svevang: Patched that in ccc9c62d7eb0f286b0896a34d2c992381c01071d

app/models/apple/publisher.rb
```diff
@@ -226,12 +229,26 @@ def wait_for_upload_processing(eps)
...
     def wait_for_asset_state(eps)
       Rails.logger.tagged("##{__method__}") do
         eps = eps.filter { |e| e.podcast_delivery_files.any?(&:api_marked_as_uploaded?) }
 
         (waiting_timed_out, _) = Apple::Episode.wait_for_asset_state(api, eps)
-        raise "Timed out waiting for asset state" if waiting_timed_out
+        if waiting_timed_out
+          attempts = eps.map { |ep| ep.apple_episode_delivery_status.asset_processing_attempts }.max
+          max_duration = eps.map { |ep| ep.feeder_episode.measure_asset_processing_duration }.compact.max
+          Rails.logger.info("Timed out waiting for asset state", {attempts: attempts, duration: max_duration, episode_count: eps.length})
```
cavis, line 249: Ooh, duration is nice to know too.  Good call.

PR #1119, opened 2024-10-08
Add default values to the audio format selects

### LINKED ISSUE #1117: Better default feed.audio_formats
When you setup a feed audio format, it seems like we're picking the lowest quality from all the dropdowns?

We should pick some better defaults.  128/2/44100 for mp3, etc.

radical-ube:
resolves #1117 

This PR sets a default value for the audio format dropdowns. The issue was that with no default, the first choice in options is selected and it is the lowest quality option (sorting low to high). Adding a default value with a higher quality intends to reduce user error unintentionally selecting low quality options.

### DIFF
--- a/app/views/feeds/_form_audio_format.html.erb
+++ b/app/views/feeds/_form_audio_format.html.erb
@@ -23,7 +23,7 @@
 
         <div class="col-3 <%= display_bitdepth(feed) %>" data-toggle-field-target="field" field="wav flac">
           <div class="form-floating input-group">
-            <%= form.select :audio_bitdepth, options_for_select(audio_bitdepth_options, 32) %>
+            <%= form.select :audio_bitdepth, options_for_select(audio_bitdepth_options, 16) %>
             <%= form.label :audio_bitdepth %>
           </div>
         </div>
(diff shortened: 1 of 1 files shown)


app/views/feeds/_form_audio_format.html.erb
```diff
@@ -16,28 +16,28 @@
...
-            <%= form.select :audio_bitrate, audio_bitrate_options %>
+            <%= form.select :audio_bitrate, options_for_select(audio_bitrate_options, 128) %>
             <%= form.label :audio_bitrate %>
           </div>
         </div>
 
         <div class="col-3 <%= display_bitdepth(feed) %>" data-toggle-field-target="field" field="wav flac">
           <div class="form-floating input-group">
-            <%= form.select :audio_bitdepth, audio_bitdepth_options %>
+            <%= form.select :audio_bitdepth, options_for_select(audio_bitdepth_options, 32) %>
```
kookster, line 26: I'm not sure we want to set this - I think we only want default options for mp3, since that is the default format, and this is a field only for "wav flac"?

cavis: This is the initially selected default when you select "flac" or "wav" from the leftmost dropdown.  So I do think we want it.

But I think 16 (or 24?) is more common than 32 for this.

PR #1120, opened 2024-10-08
Wrap subnav with authorization policy

### LINKED ISSUE #1103: Leaking unauthorized info
When looking at an episode for a podcast you don't have access to, the page still display's the podcast's title.

Look for any other cases where info may be leaking, and fix the leaks. 

image

radical-ube:
resolves #1103 

This PR addresses the issue of leaked authorized info of a podcast. Previously, a user could still see the navbar of a podcast that they did not have authorization for, and therefore could view the title of the podcast seen in the podcast switcher partial. This PR wraps the subnav in a policy in order prevent access to users who do not have view access to the selected podcast. Because the subnav is used across many other views, this should be consistent any layout that utilizes the subnav, including the `episodes` and `feeds` views.


kookster (approved):
lgtm!

PR #1122, opened 2024-10-09
Limit error interface

svevang:
For https://github.com/PRX/feeder.prx.org/issues/1016

Review #1116 before reviewing this PR

- Limits the number and places where errors are logged.
- Adds testing for error logging.
- Breaks out the error logging into help methods.


app/jobs/publish_feed_job.rb
```diff
@@ -18,30 +18,59 @@ def perform(podcast, pub_item)
...
-    res = PublishingPipelineState.error_apple!(podcast)
-    raise e if feed.apple_config.sync_blocks_rss?
-    res
+
+    begin
+      res = PublishAppleJob.do_perform(podcast.apple_config)
+      PublishingPipelineState.publish_apple!(podcast)
+      res
+    rescue => e
+      handle_apple_error(podcast, e)
```
cavis, line 52: So the overall flow through these 3 rescue blocks is something like:

1. We call `publish_apple` for the feeds
2. If PublishAppleJob errors, we rescue here.  This sets an `error_apple!` pipeline state, alerts NR, and raises.
3. We rescue again up in `def perform` - which either `handle_apple_timeout_error` or `handle_error`s it
4. That log.errors and sets a 2nd `error!` pipeline state

So we're setting 2 pipeline states per failure, NewRelic-noticing in the inner-error-handler, and rails.logger-ing in the outer-error handler.  Did I get that right?

db/migrate/20241002215949_add_asset_processing_count_to_episode_delivery_status.rb
```diff
@@ -0,0 +1,5 @@
+class AddAssetProcessingCountToEpisodeDeliveryStatus < ActiveRecord::Migration[7.2]
+  def change
+    add_column :apple_episode_delivery_statuses, :asset_processing_attempts, :integer, default: 0, null: false
```
cavis, line 3: Huh, for some reason I thought we were using the Pipeline states to track how many times we failed to process.  Is this just another way to track it?

test/models/episode_test.rb
```diff
@@ -493,4 +493,48 @@
       refute_nil episode.ready_transcript
     end
   end
+
+  describe "#increment_asset_wait" do
```
cavis, line 497: Hrm - why is this test on the `Episode`?  Shouldn't this be in one of the Apple modules?

cavis (approved):
Couple questions, making sure I understand what the nested rescues are doing.  But looks good to me 👍

db/migrate/20241002215949_add_asset_processing_count_to_episode_delivery_status.rb
svevang: Sorry, this looks like a rebasing artifact. This landed in the PR https://github.com/PRX/feeder.prx.org/pull/1116 that you reviewed earlier.

app/jobs/publish_feed_job.rb
svevang: Yup, the gist is that there are two layers of error handling. So that we can set two errors. The outer error handler represents the terminal error state. And the inner errors (error_rss, error_apple) represent transient moves through the publishing pipeline state.

svevang: Good catch, though to bring up the NR notice -- we're missing the `sync_blocks_rss` handling here

db/migrate/20241002215949_add_asset_processing_count_to_episode_delivery_status.rb
svevang: https://github.com/PRX/feeder.prx.org/pull/1116 is the start of an approach to count the timeouts, per discussion in slack etc counting the error states swallows other non-timeout errors.  Is this what you wanted to do instead?

cavis: Ah, I thought past-Ryan had seen a migration for something like this before!

This approach works - just wanted to make sure we weren't counting it twice.

cavis (approved):
Okay, making more sense after that rebase!

Looks good to me 👍 and good catch on the sync-blocks-rss.

PR #1124, opened 2024-10-10
Use mp3 instead of flac for apple feeds

### LINKED ISSUE #1121: Editable display_episodes_count on Feeds::AppleSubscription
Right now, we're hiding the `display_episodes_count` for apple feeds.  And `||=` defaulting it to the same as the public feed on initialize.

Turns out, we want to let users set this field themselves.  So they can give more episodes to their paid subscribers than the public gets.  So:

- [ ] For brand new apple feeds, default the `display_episodes_count` to the same as the public feed.
- [ ] Add the episodes count field back into the UI for apple feeds

cavis:
Turns out, Apple relaxed their requirements for paid subscribers.

This adds a utility `.guess_audio_format` to check the default_feed and episodes for the highest quality mp3 format they use.

Also closes #1121 .

### DIFF
--- a/app/models/feeds/apple_subscription.rb
+++ b/app/models/feeds/apple_subscription.rb
@@ -111,7 +111,7 @@ class Feeds::AppleSubscription < Feed
       f: "mp3",
       b: [MIN_MP3_BITRATE, af[:b]].compact.max,
       c: MP3_CHANNELS.include?(af[:c]) ? af[:c] : MP3_CHANNELS.min,
-      s: [MIN_MP3_SAMPLERATE, af[:s]].compact.max,
+      s: [MIN_MP3_SAMPLERATE, af[:s]].compact.max
     }.with_indifferent_access
   end
 end
--- a/app/models/feeds/apple_subscription.rb
+++ b/app/models/feeds/apple_subscription.rb
@@ -1,8 +1,14 @@
 class Feeds::AppleSubscription < Feed
   DEFAULT_FEED_SLUG = "apple-delegated-delivery-subscriptions"
   DEFAULT_TITLE = "Apple Delegated Delivery Subscriptions"
-  DEFAULT_AUDIO_FORMAT = {"f" => "flac", "b" => 16, "c" => 2, "s" => 44100}.freeze
+  DEFAULT_AUDIO_FORMAT = {f: "mp3", b: 128, c: 2, s: 44100}.with_indifferent_access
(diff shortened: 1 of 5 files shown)


app/models/feeds/apple_subscription.rb
```diff
@@ -77,4 +87,31 @@ def apple?
...
+
+  def episode_audio_format
+    episodes = podcast&.episodes&.published&.includes(:contents)&.limit(10)
+    contents = (episodes || []).map { |e| e.contents.first }.compact
+    mp3_contents = contents.select { |c| c.audio? && c.mime_type == "audio/mpeg" }
+
+    if mp3_contents.any?
+      max_bitrate = mp3_contents.map { |c| c.bit_rate.to_i }.max
+      max_channels = mp3_contents.map { |c| c.channels.to_i }.max
+      max_sample = mp3_contents.map { |c| c.sample_rate.to_i }.max
```
kookster, line 104: this seems right, better to use the max, i.e. better to upscale than downscale quality

app/models/feeds/apple_subscription.rb
```diff
@@ -77,4 +87,31 @@ def apple?
...
     !!apple_config&.publish_to_apple?
   end
+
+  def default_feed_audio_format
+    af = podcast&.default_feed&.audio_format
+    standardize_audio_format(af) if af && af[:f] == "mp3"
+  end
+
+  def episode_audio_format
+    episodes = podcast&.episodes&.published&.includes(:contents)&.limit(10)
```
kookster, line 97: I'd probably put the `|| []` on line 97 instead of 98

app/models/feeds/apple_subscription.rb
```diff
@@ -20,7 +26,7 @@ class Feeds::AppleSubscription < Feed
   def set_defaults
     self.slug ||= DEFAULT_FEED_SLUG
     self.title ||= DEFAULT_TITLE
-    self.audio_format ||= DEFAULT_AUDIO_FORMAT
+    self.audio_format ||= guess_audio_format
```
kookster, line 29: Now that this is variable, should we also open it up in the UI to change?

kookster (approved):
I have a couple of semi-related questions, but this looks good to me, approved!

app/models/feeds/apple_subscription.rb
cavis: Done!

PR #1131, opened 2024-10-16
Extract functional interface

svevang:
Switches some helper methods to class methods, so it's easier to interact with the Apple::Episode episode state transitions.

### DIFF
--- a/app/models/apple/episode.rb
+++ b/app/models/apple/episode.rb
@@ -384,16 +384,20 @@ module Apple
     end
 
     def publishing_state_bridge_params(state)
+      self.class.publishing_state_bridge_params(apple_id, state)
+    end
+
+    def self.publishing_state_bridge_params(apple_id, state)
       {
         request_metadata: {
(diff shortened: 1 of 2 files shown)


app/models/apple/episode.rb
```diff
@@ -384,16 +384,20 @@ def remove_episode_audio_container_parameters
...
+
+    def self.publishing_state_bridge_params(apple_id, state)
       {
         request_metadata: {
           apple_episode_id: apple_id
         },
-        api_url: api.join_url("episodePublishingRequests").to_s,
-        api_parameters: publishing_state_parameters(state)
+        api_url: Apple::Api.join_url("episodePublishingRequests").to_s,
+        api_parameters: publishing_state_params(apple_id, state)
```
kookster, line 396: yeah, that makes sense, pass through the apple_id instead of from the instance state/methods.

kookster (approved):
Seems like a straightforward improvement, lgtm

PR #1133, opened 2024-10-18
Quiet job continuation

svevang:


kookster (commented):
had a few questions!

kookster (approved):
yeah, those changes to raise logic make sense to me (and removing some code too!)

PR #1136, opened 2024-10-21
Integrations interface

svevang:
Sets up a new interface that hopes to bring coupling across integration interfaces.  This focuses on two major points of contact:

1) The set operations on the episodes, addresses episodes that are rolling through their CRUD lifecycle
2) The relationship between feeds and the integration. The approach here is to use the same pattern as the Delegated Delivery integration and focus on the private+default feeds

### DIFF
--- a/app/models/integrations/base/episode_set_operations.rb
+++ b/app/models/integrations/base/episode_set_operations.rb
@@ -1,0 +1,56 @@
+module Integrations
+  module Base
+    module EpisodeSetOperations
+      # In the case where there are duplicate guids in the feeds, we want to make
+      # sure that the most "current" episode is the one that maps to the remote state.
+      def sort_by_episode_properties(eps)
+        # Sort the episodes by:
+        # 1. Non-deleted episodes first
+        # 2. Published episodes first
+        # 3. Published date most recent first
+        # 4. Created date most recent first
+        eps =
+          eps.sort_by do |e|
+            [
+              e.deleted_at.nil? ? 1 : -1,
+              e.published_at.present? ? 1 : -1,
+              e.published_at || e.created_at,
+              e.created_at
+            ]
+          end
+
+        # return sorted list, reversed
+        # modeling a priority queue -- most important first
+        eps.reverse
+      end
+
+      def filter_episodes_to_sync(eps)
+        # Reject episodes if the audio is marked as uploaded/complete
+        # or if the episode is a video
+        eps
+          .reject(&:synced_with_integration?)
+          .reject(&:video_content_type?)
+      end
+
+      def filter_episodes_to_archive(eps, eps_in_feed)
+        # Episodes to archive can include:
+        # - episodes that are now excluded from the feed
+        # - episodes that are deleted or unpublished
+        # - episodes that have fallen off the end of the feed (Feed#display_episodes_count)
+        eps
+          .reject { |ep| eps_in_feed.include?(ep) }
+          .reject(&:integration_new?)
+          .reject(&:archived?)
+      end
+
+      def filter_episodes_to_unarchive(eps)
+        eps.filter(&:archived?)
+      end
+
+      # Only select episodes that have an remote integration state
+      def only_episodes_with_integration_state(eps)
+        eps.reject(&:integration_new?)
+      end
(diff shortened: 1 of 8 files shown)


app/models/integrations/base/episode_set_operations.rb
```diff
@@ -0,0 +1,56 @@
+module Integrations
+  module Base
+    module EpisodeSetOperations
```
svevang, line 3: The root of the episode set manipulation.  The consumption side: Deriving the sets of episodes and filtering them for CRUD sync

app/models/integrations/base/show.rb
```diff
@@ -0,0 +1,50 @@
+module Integrations
+  module Base
+    class Show
```
svevang, line 3: The show gives the handle on the episodes. The production side: baseline unfiltered, sorted set of episodes.

svevang (commented):
The overall code layout here is not super important -- i.e. using inheritance versus modules etc. The way this interface was extracted was basically to make a 1:1 with the existing DD code.  I tried to make it plausible, but the organization is not super critical.

What is important stuff is over here (bag o' methods):

1) the Show's handle on the root of the episode sets https://github.com/PRX/feeder.prx.org/pull/1136/files#diff-7e63a5638b6bcfc6976d59f81c02fd900a2f4d47e105962571a44f5cf89fc576
2) the filtering and sorting of episodes in the Episode class: https://github.com/PRX/feeder.prx.org/pull/1136/files#diff-92e7117a8391a2f8207842a2c3ac800f2d62db2e82150c9270b95426f5d5da66

app/models/integrations/base/publisher.rb
```diff
@@ -0,0 +1,29 @@
...
+
+      def episodes_to_archive
+        filter_episodes_to_archive(show.podcast_episodes, Set.new(show.episodes))
+      end
+
+      def episodes_to_unarchive
+        filter_episodes_to_unarchive(show.episodes)
+      end
+
+      def publish!
```
kookster, line 24: This makes sense as it is the only part of this that I see called by another class/module, this seems like the interface of the publisher

app/models/integrations/base/show.rb
```diff
@@ -0,0 +1,50 @@
...
+      include EpisodeSetOperations
+
+      attr_reader :feed
+
+      def initialize(public_feed:, private_feed:)
+        @public_feed = public_feed
```
kookster, line 9: public feed not used in this

kookster (commented):
I'm merging this into the megaphone branch and trying to make use of some of this.
There are some apple-ish ideas in here still, and names of things, which do not match megaphone, or likely other integrations, "show" which is only an Apple concept.

I think the set operations on episodes and comparing all episodes in the podcast to what belongs in the feed make sense, as does a publisher object responsible for making changes to individual episodes to match those sets.

The way the api integration is built to apple is different form how I already built the megahone api entities, so those integrations.base.show and episode may not be as useful.

What I am wishing I had was a common way to maintain state, like the episode status, that says if an episode is sync'd or not and to what version of audio. I think I can refactor the apple episode statuses model for this, and make it more generic.

I also wish there was a common way to record external ids related to an integration.

The sync log is close, but linking the data to the parent feed instead of the private/integration feed is an odd choice, and it lacks awareness of being an apple specific log at the moment (though adding a column could help with that).

I'm not sure why the sync log records all the data based on the public feed, when the private feed has all the other configuration info and data related to apple?
It also means a podcast cannot sync to more than one show on apple.

idk if the refactoring necessary to make these work for another (megaphone) integration will be too time consuming and extensive, I'll have to get into it and see. If not, I'll have to add different tables to maintain state.

What makes more sense to me is to consider the feed as the integration, and so storing external id of the podcast to the integration can be on the feed itself, and then the episode_feeds join table or an entity related to it makes sense to me as a place to store the external_id for the episode in that integration.

The sync log is sorta close to that, but normalized to store a multitude of entities, and currently is associated with the wrong feed id (the public one, not the integration/private feed).

svevang:
closing per https://github.com/PRX/feeder.prx.org/pull/1112

PR #1139, opened 2024-10-23
Apple Delegated Delivery UI updates

kookster:
- [x] Add block rss, and enabled check boxes
- [x] Get the Apple show ID, select from the API if available

also supports locking feeds down, so this info can't be changed via the UI:


app/controllers/feeds_controller.rb
```diff
@@ -114,10 +134,6 @@ def feed_params
     params.fetch(:feed, {}).permit(:slug).merge(nilified_feed_params)
   end
 
-  def feed_params_with_apple
```
kookster, line 117: This isn't needed as we're going to allow updating

app/controllers/feeds_controller.rb
```diff
@@ -22,19 +23,38 @@ def new
     @feed.clear_attribute_changes(%i[file_name podcast_id private slug])
   end
 
+  def get_apple_show_options(feed)
```
kookster, line 26: If there is no apple show id set, and the API key is set, then this will call the API to get the list automatically so the user can pick.

app/policies/apple/config_policy.rb
```diff
@@ -12,6 +12,6 @@ def create?
   end
 
   def update?
-    false
+    FeedPolicy.new(token, resource.feed).update?
```
kookster, line 15: updates! so we can set the block rss and enable publishing attributes

config/locales/en.yml
```diff
@@ -330,6 +330,9 @@ en:
     application:
       field_copy_tooltip: Copied!
     label:
+      apple/config:
+        publish_enabled: Enable publishing to Apple Podcasts Connect
+        sync_blocks_rss: Episodes must publish to Apple before the public feed
```
kookster, line 335: It took me an embarrassing amount of time to figure out how to set these

db/migrate/20241023012600_add_apple_show_id_to_feeds.rb
```diff
@@ -0,0 +1,6 @@
+class AddAppleShowIdToFeeds < ActiveRecord::Migration[7.2]
+  def change
+    add_column :feeds, :apple_show_id, :string
```
kookster, line 3: This is just an idea - I want to discuss and perhaps change this

svevang: From what I gathered in discussion, this is such a commonly used external id that it makes sense to pull this up onto the feed model itself so that it can be used as an XID across many different services. +1

svevang: We just need a way to override or replace the existing mechanism for linking a show to the external Apple resource.  The way it's done currently is with the SyncLog model. https://github.com/PRX/feeder.prx.org/blob/main/app/models/apple/show.rb#L108-L114

In the old init, the linking happened like so:
https://github.com/PRX/feeder.prx.org/blob/e35f2c7d29ae2266cdadd37ec74de2c18fad6a04/app/models/apple/config.rb#L93

In the current system, that initialization routine is set up under a rails hook:
https://github.com/PRX/feeder.prx.org/blob/main/app/models/feeds/apple_subscription.rb#L26-L36

`Apple::Show.connect_existing` could be a good interface to use? That, plus the associated changes in the `Show#apple_id`, could get you something.

app/controllers/feeds_controller.rb
kookster: good point!

db/migrate/20241023012600_add_apple_show_id_to_feeds.rb
kookster: I tried that, but it always tries to insert, and when the value emptied, or updated, it throws an error.
I think I need to make it a bit more complex to handle those cases

kookster: ok, I updated the `connect_existing` method to handle when the apple_show_id changes or is removed.

app/models/apple/show.rb
```diff
@@ -8,18 +8,29 @@ class Show
...
 
     def self.connect_existing(apple_show_id, apple_config)
-      api = Apple::Api.from_apple_config(apple_config)
+      if (sl = SyncLog.find_by(feeder_id: apple_config.public_feed.id, feeder_type: :feeds))
+        if apple_show_id.blank?
+          sl.destroy!
+        elsif sl.external_id != apple_show_id
+          sl.update(external_id: apple_show_id)
+        end
+      end
```
kookster, line 26: @svevang now it handles updated and deletes

app/controllers/feeds_controller.rb
kookster: done!

svevang (commented):
A couple minor notes and question.

app/controllers/feeds_controller.rb
```diff
@@ -144,20 +147,13 @@ def nilified_feed_params
...
-  end
-
-  def apple_params
-    nilify params.fetch(:feed, {}).permit(
-      apple_config_attributes: {
-        id: :id,
-        key_attributes: %i[id provider_id key_id key_pem_b64]
-      }
+      itunes_images_attributes: %i[id original_url size alt_text caption credit _destroy _retry],
+      apple_config_attributes: [:id, :publish_enabled, :sync_blocks_rss, {key_attributes: %i[id provider_id key_id key_pem_b64]}]
```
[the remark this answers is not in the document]
kookster: We have other credentials we set in other apps where we do not immediately try them, but yes, we could add a validation.

It seems like this is a new requirement, as setting the keys in the UI has been a feature for some time, so perhaps a new issue for this.

app/models/apple/show.rb
kookster: 1. this is being called by the feed being updated with the apple id, so that would be circular if we also set the feed from this.

2. agreed, good point, I can add setting this on existing feeds from the sync log in the migration (or a new migration)

app/models/feeds/apple_subscription.rb
```diff
@@ -35,6 +37,25 @@ def set_defaults
...
+  def update_apple_show
+    if previous_changes[:apple_show_id]
+      Apple::Show.connect_existing(apple_show_id, apple_config)
+    end
+  end
+
+  def apple_show_options
+    used_ids = Feed.apple.distinct.where("id != ?", id).pluck(:apple_show_id).compact
+    api = Apple::Api.from_apple_config(apple_config)
+    shows_json = Apple::Show.apple_shows_json(api) || []
```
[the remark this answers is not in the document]
kookster: oh good call, I'll add more filtering on there for ARCHIVED for this PR.
Not sure we need to replicate all Podcasts Connect functionality, like archiving and unarchiving a show, in our UI, but I agree we shouldn't allow linking our feeds to an already archived show, good catch.

app/views/feeds/_form_apple_config.html.erb
```diff
@@ -27,9 +27,34 @@
...
+      <div class="col-12 mt-4">
+        <div class="form-floating">
+          <% if !@apple_show_options.blank? %>
+            <%= form.select :apple_show_id, @apple_show_options, {include_blank: true} %>
+          <% else %>
+            <%= form.text_field :apple_show_id %>
+          <% end %>
+          <%= form.label :apple_show_id, "Apple Show ID" %>
+        </div>
+      </div>
```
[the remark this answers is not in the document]
kookster: Yeah, I thought about that, and I was torn on whether or not it was worth it - in the case where a user already knows the apple id, we don't need to make it two steps, and since almost all shows will have an apple id whether they have a subscription channel or not, I didn't want to prevent using a known value until it could be looked up.

I'll try making this conditional/two-step, and see how it goes!

app/models/apple/show.rb
kookster: added a routine to the migration!

app/controllers/feeds_controller.rb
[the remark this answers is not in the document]
svevang: Makes sense. Would be a good chore (among others) for keys.

app/views/feeds/_form_apple_config.html.erb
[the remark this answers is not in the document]
kookster: Another problem with this design is you can't delete or replace the API keys.

app/controllers/feeds_controller.rb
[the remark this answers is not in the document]
kookster: ok, added a validation to the key - disabled in test env, but otherwise checks that a call works.

app/models/feeds/apple_subscription.rb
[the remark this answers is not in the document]
kookster: filtered these out, and made it so the select is always used to update the apple show id

app/javascript/controllers/apple_key_controller.js
```diff
@@ -30,14 +30,7 @@ export default class extends Controller {
...
 
     this.providerTargets.forEach((target) => (target.value = provider))
-    this.providerTarget.disabled = false
-    this.providerTarget.focus()
-    this.providerTarget.disabled = true
-
     this.keyTargets.forEach((target) => (target.value = keyId))
-    this.keyTarget.disabled = false
-    this.keyTarget.focus()
-    this.keyTarget.disabled = true
```
kookster, line 40: I can't see that these are needed - in my testing they are not.

app/models/apple/key.rb
```diff
@@ -8,6 +8,16 @@ class Key < ApplicationRecord
 
     validate :provider_id_is_valid, if: :provider_id?
     validate :ec_key_format, if: :key_pem_b64?
+    validate :must_have_working_key
```
kookster, line 11: @svevang here's the validation for the key - it tries getting a list of shows, and fails the validation on any kind of error i.e. fails on 401s which raise an error

app/models/feeds/apple_subscription.rb
```diff
@@ -23,6 +25,14 @@ class Feeds::AppleSubscription < Feed
   validate :must_be_private
   validate :must_have_token
 
+  # for soft delete, need a unique slug to be able to make another
+  def paranoia_destroy_attributes
+    {
+      deleted_at: current_time_from_proper_timezone,
+      slug: "#{slug} - #{Time.now.to_i}"
```
kookster, line 32: Enabled deleting apple feeds, but to make a new one, the old one has to be deleted and have a distinct slug, or there is a postgres uniq index that will prevent inserting another apple feed

app/models/feeds/apple_subscription.rb
```diff
@@ -35,6 +45,25 @@ def set_defaults
...
+      Apple::Show.connect_existing(apple_show_id, apple_config)
+    end
+  end
+
+  def apple_show_options
+    used_ids = Feed.apple.distinct.where("id != ?", id).pluck(:apple_show_id).compact
+    api = Apple::Api.from_apple_config(apple_config)
+    shows_json = Apple::Show.apple_shows_json(api) || []
+    shows_json
+      .filter { |sj| sj["attributes"]["publishingState"] != "ARCHIVED" }
```
kookster, line 59: filters ARCHIVED out!

app/policies/feed_policy.rb
```diff
@@ -12,14 +12,14 @@ def create?
   end
 
   def new_apple?
-    create?
+    update?
   end
 
   def update?
-    PodcastPolicy.new(token, resource.podcast).update?
+    PodcastPolicy.new(token, resource.podcast).update? && !resource.edit_locked?
```
kookster, line 19: now allowing deletes - we can `edit_lock` feeds we do. not want touched, but this way self service folks can not only disable, but delete the integration with Apple

app/policies/feed_policy.rb
```diff
@@ -12,14 +12,14 @@ def create?
...
   end
 
   def destroy?
-    resource.custom? && update? && !resource.apple?
+    resource.custom? && update?
```
kookster, line 23: now allowing deletes - we can edit_lock feeds we do. not want touched, but this way self service folks can not only disable, but delete the integration with Apple

app/views/feeds/_form.html.erb
```diff
@@ -6,14 +6,24 @@
...
+    <% end %>
       <div class="row" data-controller="feed-link">
         <% if feed.default? %>
           <%= render "form_main", podcast: podcast, feed: feed, form: form %>
           <%= render "form_audio_format", podcast: podcast, feed: feed, form: form %>
         <% elsif apple_feed?(feed) %>
           <%= render "form_apple_config", podcast: podcast, feed: feed, form: form %>
-          <%= render "form_audio_format", podcast: podcast, feed: feed, form: form %>
-          <%= render "form_ad_zones", podcast: podcast, feed: feed, form: form %>
+          <% if feed.persisted? %>
```
kookster, line 23: now it hides the rest of the form if the apple config with key isn't saved yet

app/views/feeds/_form_apple_config.html.erb
```diff
@@ -6,38 +6,66 @@
     <div class="card-body" data-controller="apple-key">
       <div class="mb-4">
         <p><%= t(".description") %></p>
-        <a href="https://help.prx.org/hc/en-us/articles/26602803552155-Setting-up-Apple-Podcast-Subscriptions-with-Apple-Delegated-Delivery"><%= t(".guide_link") %></a>
+        <%= t(".guide_link").html_safe %>
```
kookster, line 9: this is a cosmetic change- help urls should not be embedded in the code, moved it to the translations, and allowed a link with `html_safe`

app/views/feeds/_form_apple_config.html.erb
```diff
@@ -6,38 +6,66 @@
...
       </div>
       <%= form.fields_for :apple_config do |config_form| %>
         <%= config_form.fields_for :key do |key_fields| %>
           <% unless config_form.object.persisted? %>
-            <%= key_fields.file_field :key_file, class: "mb-4 form-control", accept: ".p8, .pem", data: {action: "apple-key#convertFileToKey"} %>
+            <%= key_fields.file_field :key_file, class: "mb-4 form-control", accept: ".p8, .pem", data: {action: "apple-key#convertFileToKey"}, required: true %>
```
kookster, line 14: now this is required so you can't create the apple feed without adding a file

app/views/feeds/_form_apple_config.html.erb
```diff
@@ -6,38 +6,66 @@
...
+              <%= key_fields.label :key_id, "Apple Key" %>
+            </div>
           <% end %>
-          <div class="mb-4 form-floating">
-            <%= key_fields.text_field :provider_id, disabled: true, data: {apple_key_target: "provider"} %>
-            <%= key_fields.label :provider_id, "Provider ID" %>
+          <%= key_fields.hidden_field :provider_id, data: {apple_key_target: "provider"} %>
+          <%= key_fields.hidden_field :key_id, data: {apple_key_target: "key"} %>
+        <% end %>
+        <% if config_form.object.persisted? %>
```
kookster, line 31: within this form, also hides all the other fields until a valid key is uploaded and saved

app/views/feeds/_form_apple_config.html.erb
```diff
@@ -6,38 +6,66 @@
...
+            </div>
           </div>
+        <% end %>
+      <% end %>
+      <% if form.object.apple_config&.persisted? %>
+        <div class="col-12 mt-4">
           <div class="form-floating">
-            <%= key_fields.text_field :key_id, disabled: true, data: {apple_key_target: "key"} %>
-            <%= key_fields.label :key_id, "Apple Key" %>
+            <%= form.select :apple_show_id, @apple_show_options, {include_blank: true} %>
```
kookster, line 53: now always uses a select, with values from the api, for selecting a show

config/locales/en.yml
```diff
@@ -779,10 +782,13 @@ en:
         ad_zones: Control the types of ads that should be stitched into episodes in this feed.
     form_apple_config:
       title: Apple Subscriptions
-      description: In order to publish episodes through Apple's Podcast Connect API, ensure you've delegated access to the PRX account in Podcast Connect.
```
kookster, line 782: this isn't true - you need an API key from whatever apple account, doesn't have to be the PRX account

svevang (approved):
great! 🤩

app/views/feeds/_form.html.erb
kookster: ok, I alerted paloma to have a look at that help doc, thanks for the reminder!

Issue #1134, opened 2024-10-18
Retry delegated delivery publishing errors when `sync_blocks_rss == false`


svevang:
- Adding Retry in pipeline error handler
 - Add some test coverage

PR #1148, opened 2024-10-30
Retry on error_apple non-terminal error

svevang:
Closes https://github.com/PRX/feeder.prx.org/issues/1134

This includes the intermediate `error_apple` in the list of retryable error states. Prior to this, we had only considered _terminal_ error states. With this PR, we will retry publishing pipelines having intermediate `error_apple` states and terminal `complete` (success) states.

### DIFF
--- a/app/models/publishing_pipeline_state.rb
+++ b/app/models/publishing_pipeline_state.rb
@@ -19,7 +19,7 @@ class PublishingPipelineState < ApplicationRecord
   scope :latest_failed_pipelines, -> {
                                     # Grab the latest attempted Publishing Item AND the latest failed Pub Item.
                                     # If that is a non-null intersection, then we have a current/latest+failed pipeline.
-                                    where(publishing_queue_item_id: PublishingQueueItem.latest_attempted.latest_failed.order(id: :asc).select(:id))
+                                    where(publishing_queue_item_id: PublishingQueueItem.latest_attempted.latest_failed.select(:id))
                                   }
 
   scope :latest_by_queue_item, -> {
--- a/app/models/publishing_pipeline_state.rb
+++ b/app/models/publishing_pipeline_state.rb
@@ -19,7 +19,7 @@ class PublishingPipelineState < ApplicationRecord
   scope :latest_failed_pipelines, -> {
                                     # Grab the latest attempted Publishing Item AND the latest failed Pub Item.
                                     # If that is a non-null intersection, then we have a current/latest+failed pipeline.
-                                    where(publishing_queue_item_id: PublishingQueueItem.latest_attempted.latest_failed.select(:id))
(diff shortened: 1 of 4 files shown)


app/models/publishing_pipeline_state.rb
```diff
@@ -173,8 +173,16 @@ def self.expire_pipelines!
...
+    PublishingQueueItem.where(id: latest_failed_pipelines.select(:publishing_queue_item_id).distinct)
+  end
+
+  def self.latest_failed_podcasts
+    Podcast.where(id: latest_failed_publishing_queue_items.select(:podcast_id).distinct)
+  end
+
   def self.retry_failed_pipelines!
-    Podcast.where(id: latest_failed_pipelines.select(:podcast_id).distinct).each do |podcast|
+    latest_failed_podcasts.each do |podcast|
```
kookster, line 185: yeah, nicer to have a method for this.

app/models/publishing_pipeline_state.rb
```diff
@@ -1,6 +1,6 @@
 class PublishingPipelineState < ApplicationRecord
   TERMINAL_STATUSES = [:complete, :error, :expired].freeze
-  TERMINAL_FAILURE_STATUSES = [:error, :expired].freeze
+  FAILURE_STATUSES = [:error, :expired, :error_apple].freeze
```
kookster, line 3: nice, makes sense to have this grouping

kookster (commented):
A couple of questions, but looks right

app/models/publishing_pipeline_state.rb
```diff
@@ -19,7 +19,7 @@ class PublishingPipelineState < ApplicationRecord
   scope :latest_failed_pipelines, -> {
                                     # Grab the latest attempted Publishing Item AND the latest failed Pub Item.
                                     # If that is a non-null intersection, then we have a current/latest+failed pipeline.
-                                    where(publishing_queue_item_id: PublishingQueueItem.latest_attempted.latest_failed.select(:id))
+                                    where(publishing_queue_item_id: PublishingQueueItem.latest_attempted.latest_failed.order(id: :asc).select(:id))
```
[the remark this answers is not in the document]
svevang: Good point. The intent here was to order the output of the scope. But this attempt doesn't map cleanly between the two models (Queue and PublishingState). Removed!

Early on, I opted to avoid a base scope that defines ordering (IIRC my attempt caused some SQL slowness). Lately, I've been wishing that the output of these scopes follows a natural ordering (like :created, :rss, :completed) instead of a random ordering (as can happen with joins and aggregation).

In the console and in tests, its much easier to read the state transitions in the order they occurred.

kookster (approved):
looks good! I probably will have a little fun merging this with the megaphone changes, but I think it should be fine - maybe error_integration ends up one of those non-terminal failure states... I'll figure it out, this looks good!

Issue #1055, opened 2024-07-16
Implement social links


cavis:
Add "social links" for a Podcast (or maybe at the Feed level instead?)

[Figma design](https://www.figma.com/proto/7XYAGyTiv1FvpbbbJRyaub/Dovetail?node-id=2450-25654&t=BLIULCY9X4UhhU8w-1&scaling=min-zoom&content-scaling=fixed&page-id=80%3A2947&starting-point-node-id=2450%3A25654)

![image](https://github.com/user-attachments/assets/f1ba7092-756b-4fa5-8f8c-e27916bc2e0e)

Maybe think of a better model name for these...

 - (Podcast/Feed) has_many social links
 - Apparently we'll have to write these to an external JSON file in `f.prxu.org` ... and link to that?  Should find and confirm the spec for this.
 - Add the link out to that file to the RSS

This ticket probably needs more thought, given the almost STI-ish interface in the design.

brandonhundt:
Another pass at the UI for this, especially concerning Feeds...

- Overwrite at a feed level
- Turn off for private feeds / Apple

rpeterman-gp:
@cavis Here is the proposed spec for the `<podcast:follow>` JSON file: https://github.com/Podcastindex-org/podcast-namespace/discussions/609?ref=nathangathright.com

PR #1153, opened 2024-11-16
Episode description footer

kookster:
- [x] Add footer to the feed model and db table
- [x] Include this value when determining RSS description
- [x] Add to the feed form to edit

### DIFF
--- a/app/views/feeds/_form_main.html.erb
+++ b/app/views/feeds/_form_main.html.erb
@@ -76,3 +76,12 @@
     <%= field_help_text t(".help.display_full_episodes_count") %>
   </div>
 </div>
+
+<div class="col-12 mb-4">
+  <div class="form-floating input-group">
+    <%= form.text_field :footer %>
+    <%= form.label :footer %>
+    <%= field_help_text t(".help.footer") %>
+  </div>
+</div>
+
--- a/app/views/feeds/_form_main.html.erb
+++ b/app/views/feeds/_form_main.html.erb
@@ -79,8 +79,8 @@
 
 <div class="col-12 mb-4">
   <div class="form-floating input-group">
-    <%= form.text_field :footer %>
-    <%= form.label :footer %>
-    <%= field_help_text t(".help.footer") %>
+    <%= form.text_field :episode_footer %>
+    <%= form.label :episode_footer %>
+    <%= field_help_text t(".help.episode_footer") %>
   </div>
 </div>
--- a/app/views/feeds/_form_main.html.erb
+++ b/app/views/feeds/_form_main.html.erb
@@ -84,4 +84,3 @@
     <%= field_help_text t(".help.footer") %>
   </div>
 </div>
-
--- a/db/migrate/20241116011452_add_footer_to_feeds.rb
+++ b/db/migrate/20241116011452_add_footer_to_feeds.rb
@@ -1,5 +1,5 @@
 class AddFooterToFeeds < ActiveRecord::Migration[7.2]
   def change
-    add_column :feeds, :footer, :string
+    add_column :feeds, :episode_footer, :string
   end
 end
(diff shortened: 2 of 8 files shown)


db/migrate/20241116011452_add_footer_to_feeds.rb
```diff
@@ -0,0 +1,5 @@
+class AddFooterToFeeds < ActiveRecord::Migration[7.2]
+  def change
+    add_column :feeds, :footer, :string
```
cavis, line 3: `description_footer` or `episode_footer` might be a bit more precise?  We do have some `episode_***` fields on the feeds already, for stuff that applies at the ep-level rather than podcast/feed.

But I'm fine with this too.

app/views/feeds/_form_main.html.erb
```diff
@@ -76,3 +76,11 @@
     <%= field_help_text t(".help.display_full_episodes_count") %>
   </div>
 </div>
+
+<div class="col-12 mb-4">
+  <div class="form-floating input-group">
+    <%= form.text_field :footer %>
+    <%= form.label :footer %>
+    <%= field_help_text t(".help.footer") %>
```
cavis, line 84: There are some designs (like the screenshot on #1055) that imply this is a wysiwyg editor on the Podcast -> Engagement Settings.  And (I think) some sort of checkbox on each feed, whether or not to include the "dynamic show notes".

But - it seems like you're thinking this will be customizable per-feed, and just a single line of text.

This is a simpler solution, so I'm fine with starting here.  And we'll get some feedback!

cavis (commented):
Generally this looks good.  Some questions on naming (is hard), and the previous designs.

brandonhundt:
My thought is that if you are using programmatic, the verbiage for the
Opt-out only needs to be set once on the Podcast -> Engagement settings,
and will appear on all episodes going forward - without the producer
needing to add it to every episode. This could also be useful for other
things and will get us closer to dynamic show notes.

If it is set episode by episode, it will get missed.

brandonhundt:
Oh, if its set on the feed, that's a good place for it too!!!

db/migrate/20241116011452_add_footer_to_feeds.rb
kookster: it only makes sense to add a footer to episodes (if it was in the podcast/feed, it would just be at the bottom of the single description)

kookster: I could rename `episode_footer`, I thought about that as an option too

app/views/feeds/_form_main.html.erb
kookster: yeah this is not the dynamic show notes - that I think will make sense as a different thing, which could include dynamic elements and more of a template even

kookster:
> My thought is that if you are using programmatic, the verbiage for the Opt-out only needs to be set once on the Podcast -> Engagement settings, and will appear on all episodes going forward - without the producer needing to add it to every episode. This could also be useful for other things and will get us closer to dynamic show notes. If it is set episode by episode, it will get missed.

It makes sense in the feed to me as ad-free or other feeds won't have programmatic ads and won't require an opt out.

app/views/feeds/_form_main.html.erb
kookster: this opt-out sort of footer should be more of a set and forget single value, on other platforms it is not editable by the publishers, and customization and control require help tickets

db/migrate/20241116011452_add_footer_to_feeds.rb
cavis: I'm pro `episode_footer` - just something to indicate it's not the footor-for-this-feed's-description or something.  Since feeds can inherit a description from the default feed.

app/views/feeds/_form_main.html.erb
cavis: +1

db/migrate/20241116011452_add_footer_to_feeds.rb
kookster: done!

PR #1161, opened 2024-11-26
Add class to remove chartable prefix

### LINKED ISSUE #1160: Script to get rid of chartable prefix
- [x] Remove the chartable prefix anywhere it is in a feed enclosure prefix or template
- [x] Replace with the swap.fm prefix, if appropriate

kookster:
for review!
I ran it against prod data in debug mode, and it looks like it works on all the prefix values.
fixed #1160


kookster:
it worked!

PR #1163, opened 2024-12-08
Override enclosure url and/or prefix

kookster:
implements #1172 

- [x] Add db columns for enclosure url and prefix overrides
- [x] Use overrides in building the enclosure urls for podcast episodes
- [x] Analyze external media to get size, duration, type for RSS values
- [x] Update the episodes UI to display and set the overrides
- [x] Add "override only" media type


app/models/concerns/episode_media.rb
```diff
@@ -4,14 +4,99 @@ module EpisodeMedia
...
+
+  def destroy_out_of_range_contents
+    if segment_count.present? && segment_count.positive?
+      contents.where.not(position: segment_range.to_a).destroy_all
+    end
   end
 
   def feed_ready?
-    !media? || complete_media?
+    !media? || complete_media? || override_ready?
```
cavis, line 99: Feels like `def complete_media?` should also be checking the override?

app/models/concerns/episode_media.rb
```diff
@@ -131,19 +226,48 @@ def media_ready?(must_be_complete = true)
...
+
+  def override?
+    medium_override? || !enclosure_override_url.blank?
+  end
+
+  def external_media_ready?
+    external_media_resource&.status_complete?
+  end
+
+  def analyze_external_media
```
cavis, line 265: Feels like this could be wrapped into the existing `copy_media` (which yeah, is a misnomer... really means any async process for media).  That would keep things more consistent, around when async jobs get fired.  Thus far, we always do that in `copy_media` called from usually the controller.  Rather than activerecord hooks.

app/models/episode.rb
```diff
@@ -23,22 +23,15 @@ class Episode < ApplicationRecord
 
   acts_as_paranoid
 
-  serialize :overrides, coder: HashSerializer
```
cavis, line 26: Did this just go away?

app/models/external_media_resource.rb
```diff
@@ -0,0 +1,34 @@
...
+  end
+
+  def media_url
+    original_url
+  end
+
+  def copy_media(force = false)
+  end
+
+  def analyze_media(force = false)
```
cavis, line 20: Can we keep using copy_media (even if the name is misleading).  We could pretty easily rename across the board to `.process_media` if we wanted.

cavis (commented):
It begins with the comments!

So I think my main asks are:

- Move overrides into the existing `media_ready?` logic, to avoid all the `|| override?` conditionals
- Reuse the existing `copy_media` stuff, and maybe we just open a ticket to rename that to `process_media` across the board at some point?  Or just `process`?  Naming is hard.

I'm still mulling over having both new Episode fields and a new MediaResource subclass.  It seems a bit redundant?  But there may be a reason to keep both.

Also note to self... need to QA some of the episode publishing flows/validations, to make sure this does all the usual stuff (external must be ready on initial publish, can be replaced without being ready yet, etc).  I guess one issue there is that this is a `has_one`, so we don't have the old override in-hand to serve until the new one is ready.  Edge cases!

app/models/episode.rb
kookster: oh, we don't use it, so I removed it, but yeah, I can move it to another PR if that's surprising ;)

app/models/external_media_resource.rb
kookster: yeah, I changed this to be consistent

app/models/concerns/episode_media.rb
kookster: yes, updated!

app/models/concerns/episode_media.rb
kookster: `complete_media?` is also used for determining if media should be included in the representer / API response, in which case it would incorrectly add the media collection to the response when there was an override and no actual media to include

app/models/concerns/episode_media.rb
```diff
@@ -4,14 +4,99 @@ module EpisodeMedia
...
+  end
+
+  def validate_media_ready
+    return unless published_at.present? && media?
+
+    # media must be complete on _initial_ publish
+    # otherwise - having files in any status is good enough
+    is_ready =
+      if published_at_was.blank?
+        media_ready?(true) || override_ready?
```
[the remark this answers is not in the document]
kookster: > Why make a separate method, instead of just including the override logic in `def media_ready?`?

b/c that method is used in 2 places, it is also in the representer to check for including a media resource collection in the episode api response, and that would be off if it checked for overrides

app/models/concerns/episode_media.rb
```diff
@@ -4,14 +4,99 @@ module EpisodeMedia
...
+      elsif medium_uncut?
+        uncut.present? && !uncut.marked_for_destruction?
+      elsif override?
+        external_media_ready?
```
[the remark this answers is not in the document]
kookster: see above on why I'm not into combining these methods.
you're right, I should allow saving a published episode even when the updated external isn't processed yet.

app/models/concerns/episode_media.rb
```diff
@@ -4,14 +4,99 @@ module EpisodeMedia
...
 
   included do
+    enum :medium, [:audio, :uncut, :video, :override], prefix: true
+
     # NOTE: this just-in-time creates new media versions
     # TODO: convert to sql, so we don't have to load/check every episode?
     # TODO: stop loading non-latest media versions
     scope :feed_ready, -> { includes(media_versions: :media_resources).select { |e| e.feed_ready? } }
+
+    has_many :media_versions, -> { order("created_at DESC") }, dependent: :destroy
```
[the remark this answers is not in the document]
kookster: the question is what you want to have scattered, or split between classes and concerns??

with this PR, everything to do with episode media is in this one concern - not splitting between the model and the concern, but now some associations are split between them.

I think this is less scattered, as I was bouncing back and forth between the model and concern while trying to figure out what was going on and make changes with media.

Compromise? I add a comment with the associations listed in the model, so you don't have to go into the concern to find them?

config/locales/en.yml
```diff
@@ -381,6 +381,7 @@ en:
           audio: Audio - upload multiple audio files
           uncut: Audio - upload a single file and then segment
           video: Video
+          override: Override - specify external media url
```
brandonhundt, line 384: Recommend alphabetizing this list.

app/views/episode_media/_enclosure_override.html.erb
```diff
@@ -0,0 +1,21 @@
+<div class="col-12 mb-4">
```
brandonhundt, line 1: I created this partial file.

app/views/episode_media/_enclosure_override.html.erb
```diff
@@ -0,0 +1,21 @@
...
+  <div class="card shadow border-0">
+    <div class="card-header card-header-light d-flex justify-content-between">
+      <h2 class="card-title h5"><%= t(".title.enclosure") %></h2>
+    </div>
+    <div class="card-body">
+      <div class="form-floating input-group mb-2">
+        <%= form.text_field :enclosure_override_url %>
+        <%= form.label :enclosure_override_url, required: episode.medium_override? %>
```
brandonhundt, line 9: The fieldname should be `Enclosure Override URL`

app/views/episode_media/_enclosure_override.html.erb
```diff
@@ -0,0 +1,21 @@
...
+        <%= field_help_text t(".help.enclosure_override_url") %>
+      </div>
+      <div class="form-check">
+        <%= form.check_box :enclosure_override_prefix %>
```
brandonhundt, line 13: form name: `Enclosure Override Prefix`

config/locales/en.yml
```diff
@@ -629,6 +630,11 @@ en:
       media_not_ready: Audio not ready
       notice: Episode updated
   episode_media:
+      title:
+        enclosure_override: Enclosure Override
+      help:
+        enclosure_override_url: UNIQUE HELP TEXT
+        enclosure_override_prefix: UNIQUE HELP TEXT
```
brandonhundt, line 637: @kookster I wasn't sure how to describe these fields in a helpful way, so I added a placeholder for now.

app/helpers/episodes_helper.rb
```diff
@@ -52,16 +52,13 @@ def episode_status_class(episode)
   end
 
   def episode_media_status(episode)
-    if episode_all_media(episode).any? { |m| upload_problem?(m) }
```
cavis, line 55: Looks like `upload_problem?` is unused now.

app/models/concerns/episode_media.rb
```diff
@@ -4,14 +4,93 @@ module EpisodeMedia
...
+    return unless published_at.present? && media?
+
+    # media must be complete on _initial_ publish
+    # otherwise - having files in any status is good enough
+    unless enclosure_ready?(published_at_was.blank?)
+      errors.add(:base, :media_not_ready, message: "media not ready")
+    end
+  end
+
+  def enclosure_ready?(must_be_complete = true)
```
cavis, line 30: Wooo combined method!

app/models/concerns/episode_media.rb
```diff
@@ -4,14 +4,93 @@ module EpisodeMedia
...
+            u.duration = contents.first.duration
+
+            # use the feeder cdn url for older completed files
+            is_old = (Time.now - c.created_at) > 24.hours
+            u.original_url = (c.status_complete? && is_old) ? c.url : c.original_url
+          end
+        end
+        contents.each(&:mark_for_destruction)
+      else
+        contents.each(&:mark_for_destruction)
```
cavis, line 57: What happens when you switch from (or to) medium=override?

Should this be deleting the override you changed from override->contents or something?

app/models/concerns/episode_media.rb
```diff
@@ -20,7 +99,7 @@ def cut_media_version!
     latest_ids = latest_media.map(&:id)
 
     # backfill media_versions for newly completed media
-    if media_ready? && latest_ids != media_ids
+    if media_ready?(true) && latest_ids != media_ids
```
cavis, line 102: Okay, and here's the possibly-a-bug where it wasn't making sure the latest content was actually completed.

app/models/concerns/episode_media.rb
```diff
@@ -111,39 +192,119 @@ def video_content_type?(*args)
...
+      media.inject(0) { |s, c| s + c.file_size.to_i }
+    end
   end
 
+  # 1) must_be_complete or not
+  # that is called with "true" in 2 places
+  # - in checking valid on 1st time publish
+  # - displaying "complete" for status
+  # So really this is about 1st time, must be fully ready,
+  # as there is no prior published media version to fall back on?
```
cavis, line 215: Did you mean to leave all these notes/questions in here?

(I'm not against a more verbose comment on this method, but maybe a bit shorter than this?)

app/models/concerns/episode_media.rb
```diff
@@ -111,39 +192,119 @@ def video_content_type?(*args)
...
+  # If this is not must_be_complete, then
+  # If a single file upload, is there file uploaded, that isn't marked deleted
+  #
+  # 5) TODO: invalid or error status?
+  # should counts of files include invalid/error status files?
+  # should it be ready if any files is in error or invalid state?
   def media_ready?(must_be_complete = true)
-    if media.empty?
+    if !must_be_complete && medium_uncut?
+      uncut.present? && !uncut.marked_for_destruction?
```
cavis, line 236: Seems like this was another bug, right?  Maybe related to #1164?

app/models/enclosure_url_builder.rb
```diff
@@ -21,8 +21,17 @@ def podcast_episode_url(podcast, episode, feed = nil)
     feed ||= podcast.default_feed
     prefix = feed.try(:enclosure_prefix)
 
-    url = base_enclosure_url(podcast, episode, feed)
-    enclosure_prefix_url(url, prefix)
+    base_url = if !episode.enclosure_override_url.blank?
+      episode.enclosure_override_url
```
cavis, line 25: So this will return the override url, regardless of its status.  Are there protections against serving a processing/invalid one?  Or TBD?

config/locales/en.yml
cavis: Just noting... sorting these in en.yml won't sort them in the UI.  Since it's just pulling the order of the rails enum.  We'd have add a sort-by-transated-value to `def episode_medium_options`.

app/models/concerns/episode_media.rb
```diff
@@ -111,39 +192,119 @@ def video_content_type?(*args)
...
+        external_media_ready?(must_be_complete)
+      else
+        !enclosure_override_url.blank?
+      end
+    else
+      false
+    end
+  end
+
+  def override_processing?
```
cavis, line 286: Looks unused now.

app/models/concerns/episode_media.rb
```diff
@@ -111,39 +192,119 @@ def video_content_type?(*args)
...
+    !enclosure_override_url.blank? && !external_media_ready?
+  end
+
+  def override?
+    medium_override? || !enclosure_override_url.blank?
+  end
+
+  def external_media_ready?(must_be_complete = true)
```
cavis, line 294: Is this still necessary, with `def override_ready?` being the main place we check?  I'm not sure I follow the distinction between the override being ready and the external media.

app/models/concerns/episode_media.rb
```diff
@@ -111,39 +192,119 @@ def video_content_type?(*args)
...
     else
-      media.size >= segment_count
+      true
     end
   end
 
   def media_status
-    states = media.map(&:status).uniq
+    states = if override?
+      [external_media_resource&.status]
```
cavis, line 252: So if this is audio/video/uncut... and you overrode the enclosure_url... this will show the external media status, regardless of the contents status.  Probably fine, since these overrides get set programmatically for now.

app/models/concerns/episode_media.rb
```diff
@@ -4,14 +4,93 @@ module EpisodeMedia
...
+
     # NOTE: this just-in-time creates new media versions
     # TODO: convert to sql, so we don't have to load/check every episode?
     # TODO: stop loading non-latest media versions
     scope :feed_ready, -> { includes(media_versions: :media_resources).select { |e| e.feed_ready? } }
+
+    validate :validate_media_ready, if: :strict_validations
+
+    after_save :destroy_out_of_range_contents, if: ->(e) { e.segment_count_previously_changed? }
+    after_save :analyze_external_media
```
cavis, line 17: But this isn't kicking off any analyze tasks right?  Maybe "create external media" would be more accurate?

cavis (changes requested):
Okay, this is looking pretty dang close now.  And probably squashes some existing bugs with swapping-published-files / changing-mediums.

Couple questions, unused methods, and maybe extra comments. But that's all I got!

app/models/concerns/episode_media.rb
cavis: I guess if it just left the override there... users would at least see it on the page.  And realize they should delete that override_url.  Not terrible.

app/helpers/episodes_helper.rb
kookster: deleted!

app/models/concerns/episode_media.rb
kookster: it does fire off an analyze task for the external media url, so we can get the duration, media type, and file size t include in the enclosure tag attributes

app/models/concerns/episode_media.rb
kookster: since an override url can exist without a type, changing the type should not delete it - I think this logical?

app/models/concerns/episode_media.rb
kookster: I did want to have all that in there, as I found it useful for understanding what the method needed to do, but I can remove if you like

app/models/concerns/episode_media.rb
kookster: idk that I was thinking about that bug, but I was incorporating the uncut logic that was in the validate method into this media_ready? method, but I think it might fix that issue - I need to test it more to be sure.

app/models/concerns/episode_media.rb
kookster: yes, once there is an override url, then that is the url that will be used in the enclosure, so the displayed status needs to reflect that override's status. For the API, the complete/ready media would still be returned, but for the user, the status of the override is all they will get.

app/models/concerns/episode_media.rb
cavis: It doesn't fire a task though, it just calls `create_external_media_resource`.  And later, `copy_media` fires the async job.  (That's the general pattern in Feeder - no async stuff fired unless you explicitly ask for it via `copy_media` or other means).

app/models/concerns/episode_media.rb
kookster: yeah, I can remove it

app/models/concerns/episode_media.rb
kookster: a little duplicative, I'll combine it

app/models/enclosure_url_builder.rb
kookster: no protection -  if you override, it's overidden - when we are talking about this being used for megaphone, there is no condition where it would be better to serve the underlying media instead (unless the override was removed).

kookster: protection for this will come from the publishing pipeline, which won't publish the episode until the override is processed

app/models/concerns/episode_media.rb
cavis: +1

app/models/enclosure_url_builder.rb
cavis: I was thinking more about what would happen if `medium = override`, the episode was published, and you replaced/changed the enclosure.

I don't want to get too wrapped up in that edge case though, so this seems a good place to start!

app/models/concerns/episode_media.rb
cavis: It's just a bit outdated - has some questions, and mentions "overrides" (which aren't in this method now) and has some TODOs.

Re-reading this, and trying to state it back, maybe something like: ```
# must_be_complete=true, is used for:
#  A) validating the episode Contents are ready-to-go on initial publish
#  B) helping DTR decide which (old/replaced or new) to serve, via the API "ready_media"
#  C) cut new media_versions, when all Contents are completed
#
# otherwise, just check that this episodes has _enough_ media to stay published.
# and hopefully/eventually we'll finish processing it, and it will all be valid:
#  1) medium=audio ... must have enough files (handling segment_count=nil episodes)
#  2) medium=video ... same, but we enforce segment_count=1 elsewhere
#  3) medium=uncut ... must have a non-deleted Uncut, which we'll process/slice later
```

cavis: Or another way to state the 2nd part (and address some of your TODOs)...

`ep.media_ready?(false)` is really about that `validate_media_ready` method.  Should you be able to save this episode or not?  Having errored/invalid Contents is fine - we'll display a warning and convince you to fix it.  But it shouldn't prevent you from saving other changes.

config/locales/en.yml
kookster: looks like they are sorted based on the order in the `mediums` enum in the `episode_media.rb`
https://github.com/PRX/feeder.prx.org/blob/9dc036c81f162ae2f579a1e9f50d5f97b1c9cd0b/app/helpers/episodes_helper.rb#L116-L118

and that sort order is:
https://github.com/PRX/feeder.prx.org/pull/1163/files#diff-bda765f89a866286ad1f1b6a4cf2ed1eed4bb2b6859d7e354da995bfaf139f32R7

cavis (approved):
:+3: huzzah!  Onwards!

config/locales/en.yml
kookster: example order:
image

app/models/concerns/episode_media.rb
kookster: re-commented, thanks!

Issue #1040, opened 2024-06-06
Nested image validation errors not in UI


cavis:
I imported the TTD RSS into staging long ago.  And some of the images in there were invalid (non-square).

So now, you can't save the episode in the UI, but the nested-error message shows up nowhere.

- [ ] Confirm the current RSS importer will mark bad images as invalid (after they process)
- [ ] Make nested image errors show up in the UI

cavis:
Just tested out importing an RSS with bad image sizes on both the channel and item.  I did indeed get error messages.

![Image](https://github.com/user-attachments/assets/ef37c476-a9cd-4adb-82e6-2c7cbb386f79)

So the porter callbacks set the image status to invalid, so we're not actually triggering rails validations, and you can click save successfully.

PR #1170, opened 2024-12-21
Ignore images unless changed, default segment count when nil

kookster:
fix issues with update errors for older Mortified issues -
https://prx.zendesk.com/agent/tickets/204696

### DIFF
--- a/app/models/episode.rb
+++ b/app/models/episode.rb
@@ -149,7 +149,11 @@ class Episode < ApplicationRecord
 
   def set_defaults
     guid
-    self.segment_count ||= 1 if new_record? && strict_validations
+    if new_record? && strict_validations
+      self.segment_count ||= 1
+    elsif segment_count.blank? && contents.any?
+      self.segment_count = contents.map(&:position).max || contents.size
+    end
   end
 
   def guid
--- a/app/models/episode_image.rb
+++ b/app/models/episode_image.rb
@@ -3,8 +3,12 @@ class EpisodeImage < ApplicationRecord
 
   belongs_to :episode, touch: true, optional: true
 
-  validates :height, :width, numericality: {greater_than_or_equal_to: 1400, less_than_or_equal_to: 3000}, if: :status_complete?
-  validates :width, comparison: {equal_to: :height}, if: :status_complete?
+  validates :height, :width, numericality: {greater_than_or_equal_to: 1400, less_than_or_equal_to: 3000}, if: :should_validate?
+  validates :width, comparison: {equal_to: :height}, if: :should_validate?
+
+  def should_validate?
+    status_complete? && (height_changed? || width_changed?)
+  end
 
   def destination_path
     "#{episode.path}/#{image_path}"


app/models/episode_image.rb
```diff
@@ -3,8 +3,12 @@ class EpisodeImage < ApplicationRecord
 
   belongs_to :episode, touch: true, optional: true
 
-  validates :height, :width, numericality: {greater_than_or_equal_to: 1400, less_than_or_equal_to: 3000}, if: :status_complete?
-  validates :width, comparison: {equal_to: :height}, if: :status_complete?
+  validates :height, :width, numericality: {greater_than_or_equal_to: 1400, less_than_or_equal_to: 3000}, if: :should_validate?
+  validates :width, comparison: {equal_to: :height}, if: :should_validate?
+
+  def should_validate?
+    status_complete? && (height_changed? || width_changed?)
```
cavis, line 10: Any case for moving this to the shared `ImageFile`?  `ITunesImage` has similar validations, and there's a `:format` validation as well.

I see 1829 invalid episode images in prod ... as recently created as Feb 2023.  Bit surprised by that.

app/models/episode.rb
```diff
@@ -149,7 +149,11 @@ def image=(file)
 
   def set_defaults
     guid
-    self.segment_count ||= 1 if new_record? && strict_validations
+    if new_record? && strict_validations
+      self.segment_count ||= 1
+    elsif segment_count.blank? && contents.any?
+      self.segment_count = contents.map(&:position).max || contents.size
```
cavis, line 155: This makes me a bit more nervous.  There's already logic in the episode media to set this field, and to handle nil segment_counts (I've definitely seen those in the wild).

Was this also an issue with the ep referenced in the ticket?

kookster: yes, as I recall it had a 0 segment count, even though it had audio

app/models/episode_image.rb
kookster: image format on this one wasn't an issue, just the dimensions; probably we enforced a more lax value in the past.

cavis:
I'm also wondering if this relates to #1040.  Where an import creates an invalid image, but the UI wasn't telling me it was invalid.

cavis:
Retesting some things - I'm not sure this would ever be an issue with recent data?  (Since the introduction of `image.status == "invalid").  These days, the porter callbacks would set the image.status to invalid, so the validation wouldn't run in the first place.

app/models/episode.rb
cavis: Remembering now - nil segment counts either come in from RSS imports (items with no media), or API users.

We were keeping these nil for API users.  So when they PUT an update, they don't need to send both `media: [...three urls...]` and a matching `segment_count: 3`.

We could change `media=` to also set the segment count?  That setter is used in 3 places:

1. The API (we'd just have to decide `segment_count` related validations can't be used by API users, which is maybe fine)
2. The `EpisodeRssImport`, but it's immediately setting the segment count after that.  So changing would be fine.
3. `Uncut.slice_contents`, which i can't imagine the `segmentation` not matching the segment count anyways right?  Or we could build contents in a different way there, which would maybe be better.

app/models/episode_image.rb
cavis: Just checked prod, and I see 1829 invalid EpisodeImages.  But 0 ITunesImages.  Since this isn't a problem there, probably don't need the validation in both models.

cavis (changes requested):
Need to figure out what to do about the segment_count.  But otherwise this looks good!

kookster:
looks like this might cause problems with API calls (spoolr) that are not setting segments - do not merge quite yet!

kookster:
seems to be a not great idea - might take up the image part in a new PR, but this can close.

PR #1177, opened 2025-01-17
Export episodes to CSV

### LINKED ISSUE #1175: Get all episode player urls
Couple of requests came in to get _all_ episode embed player urls for podcasts.  So folks can update their CMS data etc.

Maybe we add a button to the episode index?  Or somewhere else?  Also - maybe we should just build out the CSV export that was in the original feeder designs.  And include a bunch of other stuff:

- PRX id/guid
- RSS `<guid>`
- Title
- Publish date
- Enclosure/mp3 url
- Embed player url
- Timings???
- Other???

cavis:
Fixes #1175.

Simple export-episodes-to-CSV button/route.  Threw a few maybe-useful things in there.  We should get some feedback and see what folks actually want (there are many many fields we could add).

image

image

### DIFF
--- a/app/controllers/episodes_controller.rb
+++ b/app/controllers/episodes_controller.rb
@@ -1,4 +1,6 @@
 class EpisodesController < ApplicationController
+  include EmbedPlayerHelper
+
   before_action :set_episode, only: %i[show edit update destroy]
   before_action :set_podcast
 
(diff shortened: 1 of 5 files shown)


brandonhundt (approved):
👍 looks good to me. Thanks @cavis

PR #1180, opened 2025-01-23
Update standard, lint

svevang:
Fixes a problem I had in Visual Studio Code where standard was not able run. This PR:

- Updates the standard gem to latest 1.44.0
- Pulls in new linting behavior via `./bin/rails lint:fix`


cavis (approved):
👍

PR #1186, opened 2025-02-07
Copy codec when slicing

cavis:


app/models/tasks/copy_media_task.rb
```diff
@@ -116,7 +116,7 @@ def porter_copy_task
   end
 
   def porter_slice_task
-    output_opts = ["-map_metadata 0"]
+    output_opts = ["-map_metadata 0", "-c copy"]
```
kookster, line 119: it slices, it dices, it can even cut through this aluminum can!

kookster (approved):
nice! looks good locally, let's make sure porter behaves!

PR #1187, opened 2025-02-11
Don't include prefixes in enclosure preview urls

### LINKED ISSUE #1178: Remove prefixes when previewing MP3s
"Preview" urls (before the ep is published) should never go through enclosure prefixes.

cavis:
Fixes #1178

### DIFF
--- a/app/helpers/embed_player_helper.rb
+++ b/app/helpers/embed_player_helper.rb
@@ -93,6 +93,12 @@ module EmbedPlayerHelper
     options_for_select(opts, selected)
   end
 
+  def enclosure_with_token(ep)
+    url = EnclosureUrlBuilder.new.base_enclosure_url(ep.podcast, ep, ep.podcast.default_feed)
+    sep = url.include?("?") ? "&" : "?"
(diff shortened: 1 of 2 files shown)


app/helpers/embed_player_helper.rb
```diff
@@ -93,6 +93,12 @@ def embed_player_category_options(podcast, selected)
     options_for_select(opts, selected)
   end
 
+  def enclosure_with_token(ep)
```
cavis, line 96: Moved this from private->public, so I could test it.  (This helper had no tests, but this seemed worth testing).

app/helpers/embed_player_helper.rb
```diff
@@ -93,6 +93,12 @@ def embed_player_category_options(podcast, selected)
...
+    url = EnclosureUrlBuilder.new.base_enclosure_url(ep.podcast, ep, ep.podcast.default_feed)
```
kookster, line 97: that's right, `base_enclosure_url` expands the template, but doesn't add in the prefix values, exactly.

kookster (approved):
lgtm

PR #1190, opened 2025-02-20
Feat/subscribe social links

radical-ube:
This PR addresses the UI portion for #1055 . The feature is added into the Podcast Engagement view, and adds a migration for the new `SubscribeLink` model used for the links. There is a dropdown from which the user can select one or multiple links to add, fields to fill in the ID information needed to create the link, and deletion. Saving will persist (or destroy) the links if they are valid.


app/helpers/podcasts_helper.rb
```diff
@@ -101,4 +101,58 @@ def disable_non_compliant_language(language)
       language
     end
   end
+
+  def subscribe_link_platform_label(link)
+    I18n.t("helpers.label.podcast.subscribe_link.#{link.platform}")
+  end
+
+  def subscribe_link_id_label(link)
```
radical-ube, line 109: difference between this and the above method is that this is used for the field label, and the above is used as the label next to the icon and in the dropdown.

app/helpers/podcasts_helper.rb
```diff
@@ -101,4 +101,54 @@ def disable_non_compliant_language(language)
...
+    end
+  end
+
+  def subscribe_link_options(podcast)
+    selected = podcast.subscribe_links.select(&:persisted?).map { |sl| sl.platform }
+
+    SubscribeLink::PLATFORMS.filter { |p| !selected.include?(p) }.map { |p| {platform: p, icon: SubscribeLink::PLATFORM_ICONS[p]} }
+  end
+
+  def subscribe_link_external_id(podcast, platform)
```
radical-ube, line 143: this is the helper that pre-populates the field. apple id gets pre-filled only if there is already an apple id to pull from. feed url uses the `public_url` and guid uses `guid`, but if these need to use different attributes let me know.

app/javascript/controllers/subscribe_links_controller.js
```diff
@@ -0,0 +1,33 @@
...
+    const now = new Date().getTime()
+    const platform = event.target.dataset.platform
+
+    const template = this.templateTargets.filter((target) => {
+      return target.dataset.platform === platform
+    })
+    const html = template[0].innerHTML.replaceAll("NEW_RECORD", now).replaceAll("NEW_EXTERNAL_ID", "")
+    this.templateTarget.insertAdjacentHTML("beforeBegin", html)
+    this.unsavedOutlet.change()
+    event.target.classList.add("d-none")
```
radical-ube, line 17: largely follows other such patterns like `FeedTokenController`, with this last addition to hide the option once it's selected.

app/models/podcast.rb
```diff
@@ -273,4 +278,42 @@ def respond_to_missing?(method, *args)
       super
     end
   end
+
+  def only_one_of_each_link_platform
```
radical-ube, line 282: i actually don't know if this is technically required? but i added this validation to prevent users from adding multiple of the same platform (i.e. no adding multiple spotifys, etc.)

app/models/podcast.rb
```diff
@@ -273,4 +278,42 @@ def respond_to_missing?(method, *args)
...
+    end
+
+    if dupes.any?
+      dupes.each do |p|
+        errors.add(:podcast, "cannot have more than one #{I18n.t("helpers.label.podcast.subscribe_link.#{p}")} link")
+      end
+    end
+  end
+
+  def build_subscribe_links_json
```
radical-ube, line 300: preparation for the next PR in building/uploading json.

app/models/subscribe_link.rb
```diff
@@ -0,0 +1,180 @@
+class SubscribeLink < ApplicationRecord
+  PLATFORMS = %w[apple spotify overcast pocketcasts youtube youtube_feed amazon antenna anytime apollo breez castamatic castbox castro curiocaster deezer fountain global goodpods gpodder hark iheart lnbeats luminary moon pandora player podbean podaddict podapp podguru_apple podguru_feed podrepublic podfriend podlp_apple podlp_guid podstation podurama podverse snipd sonnet steno truefans tunein].freeze
```
radical-ube, line 2: i started with the platforms shown in the original design, which is why they are at the beginning, but the rest are in alphabetical order. if there are any changes to be made to this order or we want to define the most commonly chosen ones, let me know.

app/models/subscribe_link.rb
```diff
@@ -0,0 +1,180 @@
...
+
+  PLATFORM_HREFS = {
```
radical-ube, line 4: pulled directly from the [Podcast Index proposal](https://github.com/nathangathright/podcast-platform-links?tab=readme-ov-file#show-links)

app/models/subscribe_link.rb
```diff
@@ -0,0 +1,180 @@
...
+    "podurama" => "https://podurama.com/podcast/${slug}-i${appleID}",
+    "podverse" => "https://api.podverse.fm/api/v1/podcast/podcastindex/${podcastIndexShowID}",
+    "snipd" => "https://share.snipd.com/show/${uniquePlatformID}",
+    "sonnet" => "https://sonnet.fm/p/${appleID}",
+    "steno" => "https://steno.fm/show/${podcastGUID}",
+    "truefans" => "https://truefans.fm/${podcastGUID}",
+    "tunein" => "https://tunein.com/podcasts/${uniquePlatformID}"
+  }
+
+  PLATFORM_ICONS = {
```
radical-ube, line 51: the reason this is a thing is that some platforms have 2 options, and i thought it would be easier to map them in this hash, rather than adding a duplicate `svg`. this mapping is then used to access the proper partial in `views/icons/_#{platform}`

app/models/subscribe_link.rb
```diff
@@ -0,0 +1,180 @@
...
+
+  def uses_slug?
+    SLUG_PLATFORMS.include?(platform)
+  end
+
+  def icon
+    PLATFORM_ICONS[platform]
+  end
+
+  def href
```
radical-ube, line 145: more preparation for the json PR to follow this one, but essentially i take the href mapping above and substitute in the `external_id` where applicable, depending on what that href uses.

app/views/icons/_amazon.html.erb
```diff
@@ -0,0 +1 @@
....19 1.37.58 1.78c.38.41.94.61 1.67.61.4 0 .75-.06 1.06-.19.07-.03.12-.06.15-.1.03-.05.04-.11.04-.17v-.24c0-.12-.04-.18-.12-.18M8.65 13.1h.69c.07 0 .13-.06.13-.12v-1.87c0-.41-.02-.97.48-.97s.42.58.42.97v1.87c0 .07.06.12.12.12h.69c.07 0 .13-.06.13-.12v-1.87c0-.2 0-.5.06-.67.07-.17.23-.28.42-.29.2 0 .36.07.41.31.03.14.02.51.02.66v1.87c0 .07.06.12.12.12h.69c.08 0 .14-.06.14-.12v-2.23c0-.38.04-.81-.18-1.11-.19-.25-.49-.39-.8-.39-.41 0-.79.21-.95.66-.19-.45-.47-.66-.9-.66-.43 0-.74.21-.91.66h-.02v-.46c0-.06-.06-.12-.12-.12h-.64c-.07 0-.13.06-.13.13V13c0 .06.06.11.12.12m7.17-1.72c0 .26 0 .48-.12.71-.11.19-.28.31-.48.31-.26 0-.42-.2-.42-.48 0-.57.52-.67 1.02-.67v.14-.01Zm0-.68c-.4.04-.91.08-1.28.23-.45.17-.75.61-.73 1.09 0 .69.45 1.04 1.02 1.04.49 0 .75-.11 1.12-.49.12.17.17.26.39.45.05.03.12.02.16-.02l.52-.44c.06-.04.05-.12 0-.18-.12-.17-.26-.3-.26-.62v-1.04c0-.44.03-.84-.3-1.15-.26-.25-.7-.33-1.03-.33-.65 0-1.37.24-1.52 1.02-.02.08.05.12.1.14l.66.07c.06 0 .11-.06.12-.12.06-.27.29-.4.55-.4.14 0 .3.05.38.17.09.14.08.33.08.48v.08l.02.02Zm-7.82 2.96a8.785 8.785 0 0 0 5.79 2.12c1.6 0 3.46-.48 4.74-1.39.21-.15.03-.37-.19-.29a11.892 11.892 0 0 1-10.2-.61c-.14-.08-.25.06-.13.17" /></g></g></svg>
```
radical-ube, line 1: i put all of the `svg`s into their own `.erb` partials in order to render them. only thing of note is that i added a `width` and a `height` variable into each one, so that the size could be controlled and passed in.

app/views/podcast_engagement/_form_subscribe_links.html.erb
```diff
@@ -0,0 +1,51 @@
...
+
+      <% if policy(podcast).update? %>
+        <div class="mb-3">
+          <button class="btn btn-primary dropdown-toggle" type="button" data-bs-toggle="collapse" data-bs-target="#dropdownOptions" aria-expanded="false" aria-controls="dropdownOptions">
+            <%= t(".add") %>
+          </button>
+        </div>
+        <div id="dropdownOptions" class=" collapse" aria-labelledby="optionsHeader">
+          <div class="card card-body p-0">
+            <ul class="list-group list-group-flush">
```
radical-ube, line 36: i turned the dropdown into a `list-group` that can `collapse`. i did this because the dropdown would always close after selecting one platform, and also it rendered a bit weirdly and felt awkward to scroll through. i also tried an accordion approach but that felt a little too extra, so this is solution is kind of between those two approaches. the dropdown can expand or close, and stays open after a selection is made. and there is stimulus to hide the option after being chosen, to give the appearance of the options updating in real time. (user would have to refresh to bring them back if they made a mistake)

app/views/podcasts/show.rss.builder
```diff
@@ -106,6 +106,10 @@ xml.rss "xmlns:atom" => "http://www.w3.org/2005/Atom",
       xml.podcast :guid, @podcast.guid
     end
 
+    if @podcast.subscribe_links.present?
```
radical-ube, line 109: placeholder material in preparation for the json PR. but this will be where the RSS gets updated after the json of links is uploaded.

config/locales/en.yml
```diff
@@ -492,6 +492,58 @@ en:
         serial_orders:
           false: Episodic
           true: Serial
+        subscribe_link:
+          apple: Apple Podcasts
+          spotify: Spotify
+          overcast: Overcast
+          pocketcasts: Pocket Casts
+          youtube: Youtube Music
+          youtube_feed: Youtube Music (feed)
```
radical-ube, line 501: is there a different way to label this to make it known to the user that this option uses a `feedURL` instead of a `uniquePlatformID`? only other platforms that do this below are Podcast Guru and PodLP.

config/locales/en.yml
```diff
@@ -1142,6 +1194,16 @@ en:
         payment_pointer: The PRX Embeddable player support micropayments. We've created a guide to walk you through the process of collecting payments as folks listen. <a href="https://help.prx.org/hc/en-us/articles/9901810244251" target="_blank">Learn more about micropayments.</a>
     form_status:
       <<: *form_status
+    form_subscribe_links:
+      title: Subscription Links
+      description: Please select which platforms you would like to display subscription links for users using the PRX Embeddable Player. Platforms will require either your Apple ID, a unique platform ID, a feed URL, a podcast GUID, or a Podcast Index ID.
```
radical-ube, line 1199: verbiage check?

cavis (changes requested):
One minor bug, and a suggestion/question (suquestion) on where to put validation/svgs.  But I like how the UI works - nice solution to the "so many different flavors of these things".  And the model mappings to generate the hrefs looks good.

app/javascript/controllers/subscribe_links_controller.js
```diff
@@ -0,0 +1,33 @@
...
+  }
+
+  removeLink(event) {
```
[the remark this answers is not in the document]
radical-ube: bug was triggered from the `nukeLink()` method below. i resolved the bug in the `subscribe_link` partial. the `hidden_field` for `external_id` was sitting outside of the removed element so it still kept submitting.

also added a bit more to `nukeLink()` to reappear the option in the dropdown.

app/models/subscribe_link.rb
radical-ube: that would certainly save a bunch of lines. will push up a change.

app/views/icons/_amazon.html.erb
radical-ube: moved!

brandonhundt:
Thanks @radical-ube! The UX is ready to ship. Yay! 

Over to @cavis for a last code review.

cavis (approved):
👍 looks good to me - onwards!

PR #1196, opened 2025-02-27
Fix release memoized values

svevang:
Fixes the case where the Apple badges can fall out of sync indicating a mismatch between remote and local state.

The Apple::Show does not release some memoized attributes causing the sync routine to ignore newly created episodes and to not poll the newly updated remote state.


cavis (approved):
👍

PR #1200, opened 2025-03-01
Update episode image attributes without replacing

### LINKED ISSUE #1198: Can not update episode image props via the API
Can't seem to send updated values for episode image's caption, credit, or alt text.

If I send an image object without an href, the image gets removed. 

if sent with a href that matches the current original URL, none of the new prop values are applied to the image.

Can only get props to update if I send a new href URL, which can be the original URL with cache buster params, but this causes the image to be processed again when the image file isn't what needs changed. This seems expensive from a compute and storage perspective since we are duplicating the same image just to change attribute values.

Need to be able to send updated image attributes from WP plugin without having to process a "new" image.

kookster:
Closes #1198

### DIFF
--- a/app/models/concerns/image_file.rb
+++ b/app/models/concerns/image_file.rb
@@ -110,14 +110,18 @@ module ImageFile
     self.status = :created
   end
 
+  # only replace if there is a new image url, different from current
   def replace?(img)
-    original_url != img.try(:original_url)
+    original_url && (original_url != img.try(:original_url))
   end
 
   def update_image(img)
     %i[alt_text caption credit].each do |key|
       img[key] = self[key]
     end
+
+    # autosave won't work; explictly save, as img is image is detached from the association
+    img.save
   end
 
   def retryable?
--- a/app/models/concerns/image_file.rb
+++ b/app/models/concerns/image_file.rb
@@ -119,9 +119,6 @@ module ImageFile
     %i[alt_text caption credit].each do |key|
       img[key] = self[key]
     end
-
-    # autosave won't work; explictly save, as img is image is detached from the association
-    img.save
   end
 
   def retryable?
--- a/app/models/episode.rb
+++ b/app/models/episode.rb
@@ -132,8 +132,9 @@ class Episode < ApplicationRecord
     images.first
   end
 
-  def image=(file)
-    img = EpisodeImage.build(file)
+  def image=(img)
+    img = EpisodeImage.new(original_url: img) if img.is_a?(String)
+    img = EpisodeImage.new(img) if img.is_a?(Hash)
 
     if !img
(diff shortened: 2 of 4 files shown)


app/models/concerns/image_file.rb
```diff
@@ -110,14 +110,18 @@ def reset_image_attributes
...
+    original_url && (original_url != img.try(:original_url))
   end
 
   def update_image(img)
     %i[alt_text caption credit].each do |key|
       img[key] = self[key]
     end
+
+    # autosave won't work; explictly save, as img is image is detached from the association
+    img.save
```
cavis, line 124: Hrm... why does autosave not happen?  It's doing a `img.update_image(image)`, and that image is `images.first` ... and `has_many :images` has autosave on it.

I wonder if this could cause issues with the RSS importer code (it also uses this setter in several places).

app/models/episode.rb
```diff
@@ -132,8 +132,9 @@ def image
     images.first
   end
 
-  def image=(file)
-    img = EpisodeImage.build(file)
+  def image=(img)
+    img = EpisodeImage.new(original_url: img) if img.is_a?(String)
+    img = EpisodeImage.new(img) if img.is_a?(Hash)
```
cavis, line 137: So when this comes from an API PUT, what is the `img` param?  I would have guessed a hash, but sounds like maybe you saw something else?

app/models/concerns/image_file.rb
kookster: rails AR autosave does insert and delete, not update

cavis: Hrm, this appears to do the update for me.  In the console:

```ruby
e = EpisodeImage.last.episode
e.images[0].caption = "whatev"
e.save!
```

kookster: you can see it's still an issue https://github.com/rails/rails/issues/17466

kookster: I believe the call to `images.first` gets the object, which is not connected to the association anymore

kookster: for example, this does not work:
```
e = EpisodeImage.last.episode
i = e.image
i.alt_text = "er"
e.save!
```

app/models/episode.rb
kookster: It came in as an `EpisodeImage` instance already

app/models/concerns/image_file.rb
cavis: Oh, I see.  Should we be doing `images[0]` then?  It seems like in our code, there's never going to be more than 1 (due to the `replace_resources!` hook deleting the others).

PR #1202, opened 2025-03-07
Feat/subscribe links json

### LINKED ISSUE #1055: Implement social links

radical-ube:
resolves #1055 

This PR resolves the Subscribe Links feature by adding the backend components. A podcast's Subscribe Links are converted into a JSON string, uploaded, and a link to that file is saved to the podcast's RSS.

### DIFF
--- a/app/controllers/podcast_engagement_controller.rb
+++ b/app/controllers/podcast_engagement_controller.rb
@@ -11,6 +11,7 @@ class PodcastEngagementController < ApplicationController
     respond_to do |format|
       if @podcast.save
         @podcast.publish!
+        @podcast.copy_subscribe_links
         format.html { redirect_to podcast_engagement_path(@podcast), notice: t(".notice") }
       else
         flash.now[:error] = t(".error")
--- a/app/models/concerns/podcast_subscribe_links.rb
+++ b/app/models/concerns/podcast_subscribe_links.rb
@@ -1,0 +1,42 @@
+require "active_support/concern"
+
+module PodcastSubscribeLinks
+  extend ActiveSupport::Concern
+
+  def build_subscribe_links_json
+    if subscribe_links.present?
+      {
+        version: "1.0.0",
+        links: subscribe_links&.map(&:as_json)
+      }.to_json
+    end
+  end
+
+  def subscribe_links_path
+    "#{base_published_url}/subscribelinks.json"
+  end
+
+  def copy_subscribe_links(options = {})
+    opts = default_options.merge(options)
+    opts[:body] = build_subscribe_links_json
+    opts[:bucket] = ENV["FEEDER_STORAGE_BUCKET"]
+    opts[:key] = "#{path}/subscribelinks.json"
+
+    @put_object = s3_client.put_object(opts)
+  end
+
+  def default_options
+    {
+      content_type: "application/rss+xml; charset=UTF-8",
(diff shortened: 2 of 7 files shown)


test/models/concerns/podcast_subscribe_links_test.rb
```diff
@@ -0,0 +1,25 @@
...
+class PodcastSubscribeLinksTest < ActiveSupport::TestCase
+  let(:podcast) { create(:podcast) }
+  let(:podcast_2) { create(:podcast) }
+  let(:apple_link) { SubscribeLink.create(platform: "apple", podcast: podcast, external_id: "12345") }
+
+  before do
+    assert apple_link
+  end
+
+  describe ".build_subscribe_links_json" do
```
radical-ube, line 12: i wasn't sure if these tests would be helpful or if there is something different i should be testing for with these methods. like formatting?

app/models/concerns/podcast_subscribe_links.rb
```diff
@@ -0,0 +1,44 @@
...
+      opts[:bucket] = ENV["FEEDER_STORAGE_BUCKET"]
+      opts[:key] = "#{path}/subscribelinks.json"
+
+      @put_object = s3_client.put_object(opts)
+    end
+  end
+
+  def default_options
+    {
+      content_type: "application/rss+xml; charset=UTF-8",
```
cavis, line 32: Should be `application/json; charset=UTF-8` instead.

test/models/concerns/podcast_subscribe_links_test.rb
cavis: I think these are about right.  In this case, the "formatting" done in the method is pretty trivial.  So testing just the business logic (returns nil when absent) makes sense.

cavis: Nitpick: the standard when writing tests is to use `.class_method_name` or `#instance_method_name`.  Since these are instance methods, they should start with a `#`.  (Little counter-intuitive yeah, but that's the standard!)

app/controllers/podcast_engagement_controller.rb
```diff
@@ -11,6 +11,7 @@ def update
     respond_to do |format|
       if @podcast.save
         @podcast.publish!
+        @podcast.copy_subscribe_links
```
cavis, line 14: We may eventually want to throw this in a background job, since writing to S3 can be a _little_ bit slow.  But it's just on this 1 form, so could be fine.  Let's find out!

cavis (approved):
One change to make on the content type, but otherwise this looks good!

PR #1205, opened 2025-03-14
Feat/exclude episodes on new feed

### LINKED ISSUE #1171: Feeds: Add a toggle to include or exclude existing episodes from the default feed
Use case: An existing show needs to add a feed including only a subset of episodes.

- Reveal needs a subfeed that only includes 4 episodes.
- When the subfeed is created, it copies all episodes into the new feed.
- Someone would need to remove the new feed from all the existing episodes. A tedious thing to do with a large backlog.

See: https://prx.slack.com/archives/C07L9VB7XSM/p1736527698377859

Proposal: Add a toggle to the feed form to include or exclude all existing episodes. It could be defaulted to include all.

radical-ube:
resolves #1171 

This PR adds a checkbox to exclude all default feed episodes upon creating a new feed. The solution in this PR is to remove the `:after_create` hook that automatically adds all the default feed episodes to a new feed, and instead call that method from the controller upon successful creation.


test/factories/feed_factory.rb
```diff
@@ -14,6 +14,12 @@
     enclosure_template { Feed.enclosure_template_default }
     enclosure_prefix { "" }
 
+    after(:create) do |feed, evaluator|
```
radical-ube, line 17: a handful of tests in a couple different models started failing after removing the after create hook in the Feed model; i think because feeds created in this way don't pass through the controller, but they assume episodes are included in non-default feeds. i added this hook into the factory as one solution to that, though upon reflection it might be better to just go into those individual suites and call `feed.set_default_episodes` in those files specifically?

app/views/feeds/_form_distribution.html.erb
```diff
@@ -61,6 +61,17 @@
         </div>
       <% end %>
 
+      <% unless feed.persisted? %>
+        <div class="col-12 mb-4">
```
brandonhundt, line 65: Can you change the `mb-4` to `mb-2`? Would have to do it on the prior checkbox as well. This would improve the spacing.

config/locales/en.yml
```diff
@@ -883,6 +883,7 @@ en:
         new_feed_url: If your podcast feed is moving, use this field to point to the new URL where your podcast is located. The current feed should be maintained until all of your subscribers have migrated.
         private: Your feed can be public or private. Leave unchecked to keep it a Public Feed. If you check this box this feed becomes Private and you'll need to generate an authorization token. Be careful who you share your private URL with - <strong>anyone who has access to the URL will be able to use it</strong>.
         url: If you already have a public URL for your podcast feed (e.g., <a href="https://support.google.com/feedburner/answer/78475?hl=en" target="_blank">feedburner</a>), enter it here. It should point to your <a href="%{published_url}" target="_blank">private feed URL</a>.
+        exclude_episodes: Exclude all episodes from the default feed from being automatically added to this feed.
```
brandonhundt, line 886: @radical-ube I'm finding the label "Exclude Default Episodes" a little confusing. The help text is clear. So maybe we can change the label to:

"Exclude Default Feed Episodes" or "Skip Default Feed Episodes"

app/controllers/feeds_controller.rb
```diff
@@ -157,4 +158,8 @@ def nilified_feed_params
       apple_config_attributes: [:id, :publish_enabled, :sync_blocks_rss, {key_attributes: %i[id provider_id key_id key_pem_b64]}]
     )
   end
+
+  def exclude_default_episodes?
+    params[:feed][:exclude_default_episodes] == "1"
```
cavis, line 163: [ActiveModel::Type::Boolean.new.cast](https://api.rubyonrails.org/classes/ActiveModel/Type/Boolean.html) is also handy for this.

cavis (commented):
:+3: looks and works good!

cavis (approved):
(Meant to click approve there, not comment)...

Issue #1154, opened 2024-11-18
Record when an episode is first written to default feed


cavis:
An important metric in our system is when an episode appears in the public (default_feed) RSS the first time.

- [x] Maybe just a timestamp, we set when the RSS is written?
- [ ] Also, on the podping side, maybe we want to actually fire a job to go request the public-rss URL, and record the first time we see it there?  (The DTR feed-scaler sort of does this)

cavis:
Closing in favor of #1275.

PR #1211, opened 2025-03-19
Feat/add first rss publish to episodes

radical-ube:
This PR addresses the first task on #1154 , which is to record a timestamp when an episode is published to the default feed rss for the first time.

### DIFF
--- a/db/migrate/20250318202734_add_first_rss_published_at_to_episodes.rb
+++ b/db/migrate/20250318202734_add_first_rss_published_at_to_episodes.rb
@@ -1,0 +1,5 @@
+class AddFirstRssPublishedAtToEpisodes < ActiveRecord::Migration[7.2]
+  def change
+    add_column :episodes, :first_rss_published_at, :timestamp
+  end
+end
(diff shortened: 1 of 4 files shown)


cavis (changes requested):
One more ask!  but this looks good so far.

db/migrate/20250318202734_add_first_rss_published_at_to_episodes.rb
```diff
@@ -0,0 +1,9 @@
+class AddFirstRssPublishedAtToEpisodes < ActiveRecord::Migration[7.2]
+  def change
+    add_column :episodes, :first_rss_published_at, :timestamp
+
+    default_feeds = EpisodesFeed.where(feed_id: Feed.default)
+    episodes_in_default_feeds = Episode.published.where(id: default_feeds.pluck(:episode_id))
+    episodes_in_default_feeds.update_all("first_rss_published_at = published_at")
```
cavis, line 7: It looks like right now, the SQL this is running is querying for 1000s of episode_ids, then running an update with them as params.  Which isn't the most efficient, when updating 100K rows.

To do this entirely in the db, I think you just need to change `.pluck` to a `.select`.  Then the whole thing runs in a single query.

cavis (approved):
:+4: one suggestion to speed up the SQL a bit more!  (Diff between `.pluck` and `.select`) ... but otherwise looks/works good to me!

PR #1213, opened 2025-03-19
Fixes from testing megaphone in staging

kookster:
- [x] Support serveDrafts on auth feeds API
- [x] Podcast title for Megaphone podcast (not feed title)
- [x] Override attributes on update, prevent summary from repeating ad choices
- [x] Don't set pre_offset post_offset, sending as nil throws a 500 from MP API
- [x] Faraday responses have recursive attributes, only log json for headers

### DIFF
--- a/app/models/feed.rb
+++ b/app/models/feed.rb
@@ -85,6 +85,10 @@ class Feed < ApplicationRecord
     false
   end
 
+  def serve_drafts
+    false
+  end
+
   def publish_integration!
   end
 
--- a/app/models/feeds/megaphone_feed.rb
+++ b/app/models/feeds/megaphone_feed.rb
@@ -41,6 +41,10 @@ class Feeds::MegaphoneFeed < Feed
     super
   end
 
+  def serve_drafts
+    publish_integration?
+  end
+
   def publish_integration?
     publish_to_megaphone?
   end
--- a/app/models/megaphone/episode.rb
+++ b/app/models/megaphone/episode.rb
@@ -16,7 +16,7 @@ module Megaphone
 
     CREATE_ATTRIBUTES = CREATE_REQUIRED + %i[pubdate pubdate_timezone author link explicit draft
       subtitle summary background_image_file_url background_audio_file_url pre_count post_count
-      insertion_points guid pre_offset post_offset expected_adhash original_filename original_url
+      insertion_points guid expected_adhash original_filename original_url
       episode_number season_number retain_ad_locations advertising_tags ad_free]
 
     UPDATE_ATTRIBUTES = CREATE_ATTRIBUTES
@@ -24,7 +24,7 @@ module Megaphone
     # All other attributes we might expect back from the Megaphone API
     # (some documented, others not so much)
     OTHER_ATTRIBUTES = %i[id podcast_id created_at updated_at status
-      download_url audio_file_processing audio_file_status audio_file_updated_at]
+      download_url audio_file_processing audio_file_status audio_file_updated_at pre_offset post_offset]
 
     DEPRECATED = %i[]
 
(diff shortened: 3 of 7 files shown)


app/models/feed.rb
```diff
@@ -85,6 +85,10 @@ def publish_integration?
     false
   end
 
+  def serve_drafts
+    false
+  end
```
kookster, line 90: by default, feeds don't allow serving draft audio

app/models/feeds/megaphone_feed.rb
```diff
@@ -41,6 +41,10 @@ def set_defaults
     super
   end
 
+  def serve_drafts
+    publish_integration?
+  end
```
kookster, line 46: megaphone feeds will allow draft audio, but only if it is a valid, saved, enabled feed

app/models/megaphone/episode.rb
```diff
@@ -16,15 +16,15 @@ class Episode < Integrations::Base::Episode
 
     CREATE_ATTRIBUTES = CREATE_REQUIRED + %i[pubdate pubdate_timezone author link explicit draft
       subtitle summary background_image_file_url background_audio_file_url pre_count post_count
-      insertion_points guid pre_offset post_offset expected_adhash original_filename original_url
+      insertion_points guid expected_adhash original_filename original_url
```
kookster, line 19: we never use these old pre_offset post_offset attributes, as we set cerpoints, but, sending them as nil was causing 500 errors, so just leave them out.

app/models/megaphone/episode.rb
```diff
@@ -145,8 +145,9 @@ def delete_delivery_status
     # like when dtr wasn't ready at first
     # so we can make that update and then mark uploaded
     def upload_audio!
-      set_audio_attributes
-      update!
+      # basically just do an update, passing in the episode
+      # which will override the audio attributes
+      update!(feeder_episode)
```
kookster, line 150: ok, when this is called, it does a retrieve of the current attributes on megaphone, then writes those back - but, for the summary, each cycle of doing that adds the ad choices line at the end, so after a few audio saves, it ends up repeated a bunch of times. Instead, go with the current attributes for audio and everything else when updating.

app/models/megaphone/episode.rb
```diff
@@ -331,8 +332,8 @@ def get_media_info(enclosure)
         info[:location] = resp.env.response_headers["location"]
         info[:size] = resp.env.response_headers["content-length"].to_i
       else
-        logger.error("DTR media redirect not returned: #{resp.status}", enclosure: enclosure, resp: resp)
-        raise("DTR media redirect not returned: #{resp.status}")
+        logger.error("DTR media redirect not returned: #{resp&.status}", enclosure: enclosure, headers: resp&.env&.response_headers)
```
kookster, line 335: the Faraday response includes a reference back to the Faraday response, i.e. `response == response.env.response`, which was causing an infinite loop / stack error when serializing to json.

app/models/megaphone/podcast.rb
```diff
@@ -54,7 +54,7 @@ def self.attributes_from_feed(feed)
       podcast = feed.podcast
       itunes_categories = feed.itunes_categories.present? ? feed.itunes_categories : podcast.itunes_categories
       {
-        title: feed.title || podcast.title,
+        title: podcast.title,
```
kookster, line 57: the feed title is always "Megaphone Feed", so use the podcast's

app/models/megaphone/episode.rb
cavis: Ooh, fun!

app/models/megaphone/podcast.rb
cavis: I do wonder if we should have distinguished "feed title override" vs "feed label".  Where the label is just for the Feeder UI, so you can tell them apart.

Ah well, problem for our future selves.

cavis (approved):
👍

PR #1214, opened 2025-03-20
Fix/mega fixes 2

kookster:
- [x] source atributes weren't getting saved correctly to the delivery status
- [x] source file name can't be used for comparison b/c of how megaphone saves original filename via api
- [x] bad log statement on error in asset checking

### DIFF
--- a/app/models/megaphone/episode.rb
+++ b/app/models/megaphone/episode.rb
@@ -4,7 +4,7 @@ module Megaphone
     attr_accessor :podcast
 
     # track upload source data
-    SOURCE_ATTRIBUTES = [:source_media_version_id, :source_size, :source_fetch_count, :source_url, :source_filename]
+    SOURCE_ATTRIBUTES = %i[source_media_version_id source_size source_fetch_count source_url source_filename]
     attr_accessor(*SOURCE_ATTRIBUTES)
 
     # Used to form the adhash value
@@ -161,7 +161,7 @@ module Megaphone
       set_audio_attributes
 
       # check to see if the audio on mp matches
-      if original_filename == source_filename
+      if audio_is_latest?
         # if processing done, set the cuepoints and the enclosure
         if !audio_file_processing && audio_file_status == "success"
           set_enclosure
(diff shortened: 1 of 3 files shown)


app/models/megaphone/publisher.rb
```diff
@@ -53,7 +53,7 @@ def check_status_episodes!
       # if after all those checks, still incomplete? throw an error
       if episodes.size > 0
         msg = "Megaphone::Publisher.check_status_episodes! timed out on: #{episodes.map(&:id)}"
-        Logger.error(msg)
+        Rails.logger.error(msg)
```
kookster, line 56: just bad

app/models/megaphone/episode.rb
```diff
@@ -178,6 +178,12 @@ def check_audio!
       end
     end
 
+    def audio_is_latest?
+      feeder_audio_updated_at = feeder_episode.media_versions.last&.updated_at
+      megaphone_updated_at = DateTime.parse(audio_file_updated_at) if audio_file_updated_at
+      megaphone_updated_at && feeder_audio_updated_at && feeder_audio_updated_at <= megaphone_updated_at
+    end
```
kookster, line 185: so the `original_filename` is getting set to a guid, not the value send over by the API.
I believe this is how it is when uploading via the API, when uploading via the site, I believe this is the correct filename.

So, we can't rely on that - instead, use the updated_at of the media version, and expect that the audio on megaphone has to be updatd after that, or it is out of sync.

Not perfect, but better than nothing to check to see the audio updated - if this proves to be unreliable, could start tracking the timestamp of the audio even more.

cavis: Assuming _some_ lag in megaphone processing, this seems like it would be good enough.

cavis (approved):
👍 looks good to me!

app/models/megaphone/episode.rb
kookster: I'm assuming lag in our processing, between when we cut a version, and when that is available on DTR. and when we then send it megaphone as a new audio file

PR #1223, opened 2025-03-25
Manage bad delegated delivery api key warnings

### LINKED ISSUE #1193: Delegated delivery key authorization errors prevent updates
In the case where the apple delegated delivery key has a lesser set of permissions, the delivery error + retry semantics don't distinguish between temporary and permanent error states. Because you can attach "roles" to the keys, it's possible to craft an api key that does is read-only.

The resulting errors look like:

```
"data": {
          "body": {
            "errors": [
              {
                "id": "88258fa2-8d8c-46ff-b3a6-80a650ef9ed4",
                "status": "403",
                "code": "FORBIDDEN_ERROR",
                "title": "This request is forbidden for security reasons",
                "detail": "The API key in use does not allow this request"
              }
            ]
          },
          "status": 403,
          "context": "response"
        }
```

- distinguish this permanent error of insufficient permissions, from the more general transitory errors
- add a circuit breaker pattern to ensure that we don't endlessly retry. For this authorization error, either fail immediately or use a very low retry count. Disable the apple uploads or publishing (via sync_blocks_rss semantics)
- Set up alerts to 'round trip' the circuit breaker action to the user and support staff (slack?)

svevang:
closes #1193 

This PR handles a new type of error in the publishing flow described in #1193. This authorization error cannot be recovered from automatically, so for now the approach is use the existing ops-errors as a message queue to have support folks help address it with the user. But, we don't want this to swamp the ops-error channel with a never-ending deluge of errors, so a backoff / short-circuit seems appropriate: release a burst of errors and then level off at an error every 8 hours.

The job error semantics follow the AssetTimeoutError. We repeatedly try the publishing routine, but unlike the AssetTimeoutError that counts up and then starts erroring, this will start erroring and exponentially back off.

This also tries to flatten the error handling in the PublishFeedJob, combining and removing duped calls and handlers.

### DIFF
--- a/app/models/apple/api_error.rb
+++ b/app/models/apple/api_error.rb
@@ -22,5 +22,28 @@ module Apple
 
       resp
     end
+
+    def self.for_response(message, response)
+      begin
+        body = JSON.parse(response.body) if response.try(:body).present?
+
+        # Check API key permission error pattern
+        if matches_permission_error_pattern?(body)
+          return ApiPermissionError.new("Apple API permission error - delegated delivery key lacks required permissions", response)
+        end
+      rescue JSON::ParserError
+        # Fall through to default
+      end
+
+      new(message, response)
+    end
+
+    def self.matches_permission_error_pattern?(body)
+      return false unless body.is_a?(Hash)
+
+      first_error = body&.dig("errors", 0)
+      first_error&.dig("code") == "FORBIDDEN_ERROR" &&
+        first_error&.dig("detail")&.include?("API key in use does not allow this request")
+    end
   end
 end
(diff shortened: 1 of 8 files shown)


test/jobs/publish_feed_job_test.rb
```diff
@@ -172,7 +229,7 @@
             [3, 40],
             [4, 40],
             [5, 50],
-            [6, 60]
+            [6, 50]
```
cavis, line 232: So this was fatal, but now just an error?

app/models/apple/api_error.rb
```diff
@@ -22,5 +22,28 @@ def format_message(message, response)
...
+      end
+
+      new(message, response)
+    end
+
+    def self.matches_permission_error_pattern?(body)
+      return false unless body.is_a?(Hash)
+
+      first_error = body&.dig("errors", 0)
+      first_error&.dig("code") == "FORBIDDEN_ERROR" &&
```
cavis, line 45: Have you seen forbidden statuses for other problems?  Just wondering if matching on the error details could break if they ever reworded it slightly.

cavis (approved):
One question on matching on error details, but that may be necessary to match/isolate these.  👍 looks good to me!

Issue #1212, opened 2025-03-19
Integration sync log should touch updated_at for each publishing success / sync


svevang:
- touch the episode sync log's update_at

PR #1225, opened 2025-03-26
Touches the updated_at timestamp with each sync

svevang:
Closes [#1212](https://github.com/PRX/feeder.prx.org/issues/1212)

Makes sure that the sync log's timestamp is updated with each "sync" of the remote state. Helpful for debugging.


kookster (approved):
Looks like rails has no better way to do this
(this PR might, but not getting approved https://github.com/rails/rails/pull/51455)
Looks good, I'm going to steal this for the megaphone code too!

PR #1228, opened 2025-03-27
Fix megaphone audio format

kookster:


app/models/feeds/megaphone_feed.rb
```diff
@@ -33,10 +42,11 @@ def integration_type
...
 
   def set_defaults
-    self.slug = "prx-#{id}"
-    self.title = DEFAULT_TITLE
-    self.tokens = [FeedToken.new(label: DEFAULT_TITLE)] if tokens.empty?
+    self.audio_format ||= DEFAULT_AUDIO_FORMAT
     self.private = true
+    self.slug ||= "prx-#{id}"
+    self.title ||= DEFAULT_TITLE
+    self.tokens = [FeedToken.new(label: DEFAULT_TITLE)] if tokens.empty?
```
cavis, line 49: Just to note, this `tokens.empty?` (and the one in the apple feed) do run a SQL query.  I ran into this the other day trying to do an `Episodes.includes(:feeds)` ... I would get an N+1 just from trying to initialize the feeds.

Not right now, but we should probably pick a different hook to init these.

app/views/feeds/_form.html.erb
```diff
@@ -26,6 +26,9 @@
           <% end %>
         <% elsif megaphone_feed?(feed) %>
           <%= render "form_megaphone_config", podcast: podcast, feed: feed, form: form %>
+          <% if feed.persisted? %>
+            <%= render "form_audio_format", podcast: podcast, feed: feed, form: form, lock_format: true %>
```
cavis, line 30: Ah, so they can see what format it'll be but not change it - +1.

app/models/feeds/megaphone_feed.rb
kookster: oh good point
https://github.com/PRX/feeder.prx.org/issues/1229

app/views/feeds/_form.html.erb
kookster: yeah, I thought this was a fairly decent way to do that, by passing in a value to the partial?
it works!

PR #1231, opened 2025-03-29
Megaphone external resource handling and fixes

kookster:


app/controllers/episode_media_controller.rb
```diff
@@ -32,6 +32,7 @@ def update
       if @episode.save
         @episode.uncut&.slice_contents!
         @episode.copy_media
+        @episode.publish!
```
kookster, line 35: we need to also update integrations like megaphone when the media controller has updates

app/models/megaphone/episode.rb
```diff
@@ -7,6 +7,9 @@ class Episode < Integrations::Base::Episode
     SOURCE_ATTRIBUTES = %i[source_media_version_id source_size source_fetch_count source_url source_filename]
     attr_accessor(*SOURCE_ATTRIBUTES)
 
+    # Default duration to publish early on megaphone
+    PUBLISH_EARLY_DURATION = 30.minutes.freeze
```
kookster, line 11: kind of like our feed offset, but publish in megaphone a bit early

app/models/megaphone/episode.rb
```diff
@@ -16,15 +19,19 @@ class Episode < Integrations::Base::Episode
 
     CREATE_ATTRIBUTES = CREATE_REQUIRED + %i[pubdate pubdate_timezone author link explicit draft
       subtitle summary background_image_file_url background_audio_file_url pre_count post_count
-      insertion_points guid expected_adhash original_filename original_url
-      episode_number season_number retain_ad_locations advertising_tags ad_free]
+      insertion_points guid expected_adhash original_filename episode_number season_number
+      retain_ad_locations ad_free episode_type clean_title]
```
kookster, line 23: `original_url` is for imports, don't update it
`episode_type` was missing
`clean_title` was missing

`advertising_tags` will be a different PR, don't update for now

app/models/megaphone/episode.rb
```diff
@@ -16,15 +19,19 @@ class Episode < Integrations::Base::Episode
...
     # All other attributes we might expect back from the Megaphone API
     # (some documented, others not so much)
-    OTHER_ATTRIBUTES = %i[id podcast_id created_at updated_at status
-      download_url audio_file_processing audio_file_status audio_file_updated_at pre_offset post_offset]
+    OTHER_ATTRIBUTES = %i[id podcast_id created_at updated_at status download_url
+      audio_file_processing audio_file_status audio_file_updated_at pre_offset post_offset
+      image_file audio_file size duration uid original_url bitrate samplerate channel_mode vbr
+      id3_file id3_file_processing id3_file_size spotify_identifier custom_fields content_rating
+      promotional_content podcast_title network_id podcast_author podcast_itunes_categories
+      episode_video_thumbnail_url main_feed rss_status spotify_status cuepoints advertising_tags]
```
kookster, line 34: There's a bunch of semi-documented attributes that come back - go ahead and capture them all

app/models/megaphone/episode.rb
```diff
@@ -66,22 +73,27 @@ def self.new_from_episode(megaphone_podcast, feeder_episode)
     end
 
     def self.attributes_from_episode(e)
+      pubdate = if e.published_or_released_date
+        e.published_or_released_date - PUBLISH_EARLY_DURATION
```
kookster, line 77: 30 minutes early?

app/models/megaphone/episode.rb
```diff
@@ -66,22 +73,27 @@ def self.new_from_episode(megaphone_podcast, feeder_episode)
...
         explicit: e.explicit,
         draft: e.draft?,
+        clean_title: e.clean_title,
         subtitle: e.subtitle,
         summary: e.description,
         background_image_file_url: e.ready_image&.href,
+        episode_type: e.itunes_type,
         episode_number: e.episode_number,
         season_number: e.season_number,
-        advertising_tags: e.categories,
```
kookster, line 84: don't set these for now!

app/models/megaphone/episode.rb
```diff
@@ -109,7 +121,7 @@ def create!
       self
     rescue Faraday::ClientError => ce
       self.api_response = ce.response
-      raise e
+      raise ce
```
kookster, line 124: bug fix!

app/models/megaphone/episode.rb
```diff
@@ -163,11 +187,8 @@ def check_audio!
       # Re-request the megaphone api
       find_by_megaphone_id
 
-      # get the audio attributes from feeder
-      set_audio_attributes
-
-      # check to see if the audio on mp matches
-      if audio_is_latest?
+      # check to see if the audio on mp matches latest delivery
+      if original_filename == delivery_status(true).source_filename
```
kookster, line 191: Turns out you _can_ use the `original_filename` this way, but it has to be explicitly set - no problem! we'll do that in the code for setting audio attributes below...

app/models/megaphone/episode.rb
```diff
@@ -163,11 +187,8 @@ def check_audio!
       # Re-request the megaphone api
       find_by_megaphone_id
 
-      # get the audio attributes from feeder
-      set_audio_attributes
```
kookster, line 167: Don't set these, it overrides the values from megaphone with the values in feeder, and we want to check what's in megaphone!

app/models/megaphone/episode.rb
```diff
@@ -191,9 +212,31 @@ def audio_is_latest?
     end
 
     def set_enclosure
-      return unless audio_file_status == "success"
-      return if download_url.blank? || feeder_episode.enclosure_override_url.present?
-      feeder_episode.update(enclosure_override_url: download_url, enclosure_override_prefix: true)
+      return if download_url.blank? || !feeder_episode.complete_media?
+      set_external_media
+      feeder_episode.update!(enclosure_override_url: download_url, enclosure_override_prefix: true)
+    end
```
kookster, line 218: Definite changes here!
first off, don't try to set the enclosure override if no audio has been uploaded to DT or sent to megaphone

next, don't retrieve the attributes by requesting from megaphone - 
it won't work while it's a draft, and since we uploaded it, we know all the attributes

finally, if updating does fail, raise an error

app/models/megaphone/episode.rb
```diff
@@ -191,9 +212,31 @@ def audio_is_latest?
...
+
+    # if it's not published yet, can't get the file from mp, create as "complete"
+    def set_external_media
+      return if feeder_episode.published?
```
kookster, line 222: if the episode is published, we can retrieve the info about the media, and verify it is there.

if it isn't, we rely on what we know the values will be, and by creating the external media, prevent it failing and staying "processing" until published.

There is some missing action, which is to validate the megaphone url on publish/release - which means waiting for the PR which tracks when an episode first shows up in the RSS.

app/models/megaphone/episode.rb
```diff
@@ -191,9 +212,31 @@ def audio_is_latest?
...
+      emr_attributes = {
+        status: :complete,
+        mime_type: "audio/mpeg",
+        medium: "audio",
+        original_url: download_url,
+        file_size: feeder_episode.media_file_size_sum,
+        duration: feeder_episode.media_duration_sum,
+        bit_rate: private_feed.audio_format[:b].to_i,
+        channels: private_feed.audio_format[:c].to_i,
+        sample_rate: private_feed.audio_format[:s].to_i
```
kookster, line 233: I could also get these audio format attributes from megaphone, but I know they will be in DT, there may be timing issues getting them from mp.

app/models/megaphone/episode.rb
```diff
@@ -237,19 +280,27 @@ def handle_response(api_response)
...
+      # if there is not audio yet, we're all done
+      if !feeder_episode.complete_media?
+        delivery_status(true).mark_as_delivered!
+      # if the audio doesn't match, either it was just uploaded, or needs it
+      elsif !has_media_version?
         # if there's audio and we just uploaded it successfully, set attr, then check status
         if background_audio_file_url
-          attrs = source_attributes.merge(uploaded: true, delivered: false)
+          attrs = source_attributes.merge(
+            enclosure_url: background_audio_file_url,
```
kookster, line 291: it wasn't saving this!

app/models/megaphone/episode.rb
```diff
@@ -237,19 +280,27 @@ def handle_response(api_response)
...
+        # if versions don't match, and we didn't upload, it isn't uploaded or delivered
         else
           delivery_status(true).mark_as_not_delivered!
         end
-      # or if there's not audio yet or it didn't change
       else
-        delivery_status(true).mark_as_delivered!
+        # media is complete and has the right version, that's uploaded!
+        # next pass through check_audio! should mark it delivered if mp matches and is complete
+        delivery_status(true).mark_as_uploaded!
```
kookster, line 303: this is a fix - once there is audio, only `check_audio!` should mark the episode as `delivered`, not here in updating the status after a create or update api call.

app/models/megaphone/episode.rb
```diff
@@ -306,21 +357,24 @@ def get_placement_zones(original_count = nil)
...
         if media_info[:media_version] == feeder_episode.media_version_id
           audio_url = arrangement_version_url(media_info[:location], media_info[:media_version])
+          audio_filename = url_filename(audio_url)
 
           self.source_media_version_id = media_info[:media_version]
           self.source_size = media_info[:size]
           self.source_fetch_count = (delivery_status&.source_fetch_count || 0) + 1
           self.source_url = audio_url
-          self.source_filename = url_filename(source_url)
+          self.source_filename = audio_filename
```
kookster, line 374: turns out you can just set this - woo!

app/models/megaphone/episode.rb
```diff
@@ -306,21 +357,24 @@ def get_placement_zones(original_count = nil)
...
+
           self.background_audio_file_url = audio_url
+          self.original_filename = audio_filename
```
kookster, line 377: turns out you can just set this - woo!

app/models/megaphone/publisher.rb
```diff
@@ -66,17 +66,18 @@ def check_episodes(episodes)
...
-        if !status.delivered? && status.uploaded?
+        elsif !status.delivered?
           megaphone_episode.check_audio!
         end
 
-        if !ep.episode_delivery_status(:megaphone).delivered?
+        # make sure we get the latest status on the episode
+        ep.reload
+        episode_status = ep.episode_delivery_status(:megaphone)
+        if !episode_status.delivered? || !episode_status.uploaded?
```
kookster, line 80: Calling `reset` on delivery status, but saw some weird stuff where it seemed like the status wasn't updated - may not have been a thing, but this is just a bit more defensive about reloading and checking status.

app/views/feeds/_form_audio_format.html.erb
```diff
@@ -9,7 +9,9 @@
       <div class="row">
         <div class="col-3">
           <div class="form-floating input-group" aria-label="audio type">
-            <%= form.select :audio_type, audio_format_options, {include_blank: true}, disabled: local_assigns[:lock_format], data: {action: "toggle-field#displayOnlyThisField"} %>
+            <% html_options = {data: {action: "toggle-field#displayOnlyThisField"}} %>
+            <% html_options[:disabled] = true if local_assigns[:lock_format] %>
+            <%= form.select :audio_type, audio_format_options, {include_blank: true}, html_options %>
```
kookster, line 14: if disabled is set to false, then even if it should be disabled (e.g. locked), then it shows is enabled.
Instead only set a disabled option when setting it to true.

app/controllers/episode_media_controller.rb
cavis: Currently, we copy the media and then later (after Porter callbacks, if any) we trigger `ep.publish!`.

Does that not catch this case?  Or you just don't want to wait for Porter to kick off a publishing pipeline?

app/models/concerns/episode_media.rb
```diff
@@ -195,18 +195,26 @@ def media_duration
     if override?
       external_media_resource&.duration
     else
-      media.inject(0.0) { |s, c| s + c.duration.to_f } + podcast.try(:duration_padding).to_f
+      media_duration_sum + podcast.try(:duration_padding).to_f
```
cavis, line 198: Totally forgot we even had a padding!  Is that normally just null?

app/models/megaphone/episode.rb
cavis: Is there a reason not to use the db field?  Just to prevent other unintended consequences, or ensure it's not overridable?

app/views/feeds/_form_audio_format.html.erb
cavis: Ah, yes the formbuilder respecting any val you pass as an override.  Good catch!

cavis (approved):
Couple questions, but :+3: nothing that needs to hold this up!

app/controllers/episode_media_controller.rb
kookster: No, it doesn't catch when there is a draft episode, and the number of segments is changes, for example.

app/models/megaphone/episode.rb
kookster: If I set and use the db field, it has the unintended consequence of applying to when episodes are published in our feed as well, which isn't the point here - it might be safe? Since our megaphone feed is not relying on the offset for DTR access, it might be ok? When I looked through the places where `episode_offset_seconds` was used, it seemed safer to avoid it, but it looks ok when I look again now...

app/models/concerns/episode_media.rb
kookster: 15 or so? old ones I bet
```
feeder(prod)> Podcast.where("duration_padding is not null").count
=> 15
```

cavis (approved):
👍

Issue #1224, opened 2025-03-26
Allow per feed guids


kookster:
as a setting, allow episodes in a feed to have a different guid than the default/public feed
perhaps just add the slug to the guid?

PR #1233, opened 2025-04-03
Add option for unique guids for episodes in a feed

kookster:
implements #1224


config/locales/en.yml
```diff
@@ -917,6 +918,7 @@ en:
         include_podcast_value: In some cases, you may wish to exclude micropayments from a private feed.
         itunes_category: Enter an alternate <a href="https://help.apple.com/itc/podcasts_connect/#/itc9267a2f12" target="_blank">Apple Podcasts category and subcategory</a>, differentiating this Feed from the Podcast.
         subtitle: Enter an alternate teaser
+        unique_guids: If checked, each episode will have a unique GUID in the RSS for this feed, based on the public GUID and the feed slug.
```
cavis, line 921: I mean, all episodes have unique guids in their feeds, right?

Maybe something about how the guid will be _different_ than in the public RSS if you check this box?  (And that you usually wouldn't want that).

app/helpers/podcasts_helper.rb
```diff
@@ -42,6 +42,14 @@ def episode_description(episode, feed)
     description
   end
 
+  def episode_guid(episode, feed)
+    if feed.unique_guids?
+      "#{episode.item_guid}_#{feed.slug || feed.id}"
```
cavis, line 47: This does mean... if the feed slug changes, all the <guid>s in that feed change.

But hopefully that wouldn't happen IRL?  Changing the slug would also change your feed's url.  Which would already be breaking things for listeners.

cavis (approved):
One thought on help text, but otherwise 👍

config/locales/en.yml
kookster: text updated!

app/helpers/podcasts_helper.rb
kookster: switched to the immutable feed.id

cavis (approved):
👍

PR #1150, opened 2024-11-05
Retry asset wait file processing timeouts

svevang:
Fixes https://github.com/PRX/feeder.prx.org/issues/1138

This works towards some state handling changes based on https://github.com/PRX/feeder.prx.org/pull/901 

1) prevent state oscillations around uploads. Depending on the exact set of circumstances, this could include duplicate media upload cycles and skipping polling for remote asset states
2) This breaks up the main publishing routine and adds test coverage for that routine. So now there are two phases. First a media upload step and then the processing and delivery/publishing.

PR #1234, opened 2025-04-07
Guard against delivering already delivered episodes

svevang:
Fixes an issue where an episode is marked drafting async via the podcasts connect API and then is stuck in redelivery because of missing remote podcasts delivery file resources (which are typically pruned after a month).

This PR builds on #1150 by guarding against re-delivery (probing for file processing, waiting on asset state) for episodes that have simply had their drafting status toggled. This leverages the new top level _uploaded_ and _delivered_ state that landed with #1150 (merged via the megaphone branch), and makes sure that for episodes marked uploaded and delivered that no attempt to "re-deliver" is made.

### DIFF
--- a/app/models/apple/publisher.rb
+++ b/app/models/apple/publisher.rb
@@ -91,8 +91,11 @@ module Apple
             upload_media!(eps)
           end
 
-          process_and_deliver!(eps)
+          eps.filter(&:apple_needs_delivery?).tap do |eps|
+            process_and_deliver!(eps)
+          end
 
+          publish_drafting!(eps)
           raise_delivery_processing_errors(eps)
         end
       end
@@ -128,7 +131,6 @@ module Apple
 
       mark_as_delivered!(eps)
 
-      publish_drafting!(eps)
       reset_asset_wait!(eps)
     end
 
--- a/test/models/apple/publisher_test.rb
+++ b/test/models/apple/publisher_test.rb
@@ -611,5 +611,91 @@ describe Apple::Publisher do
 
       mock.verify
     end
+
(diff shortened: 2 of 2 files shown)


app/models/apple/publisher.rb
```diff
@@ -91,8 +91,11 @@ def deliver_and_publish!(eps)
             upload_media!(eps)
           end
 
-          process_and_deliver!(eps)
+          eps.filter(&:apple_needs_delivery?).tap do |eps|
```
kookster, line 94: nice, good use of that status check, and makes sense to filter before calling `process_and_deliver!`

test/models/apple/publisher_test.rb
```diff
@@ -611,5 +611,101 @@
...
+
+      # Set up method stubs
+      apple_publisher.stub(:increment_asset_wait!, increment_mock) do
+        apple_publisher.stub(:wait_for_upload_processing, wait_upload_mock) do
+          apple_publisher.stub(:wait_for_asset_state, wait_asset_mock) do
+            apple_publisher.stub(:mark_as_delivered!, mark_delivered_mock) do
+              apple_publisher.stub(:reset_asset_wait!, reset_mock) do
+                # This should raise an error if publish_drafting! is called
+                apple_publisher.stub(:publish_drafting!, ->(*) { raise "publish_drafting! should not be called!" }) do
+                  apple_publisher.process_and_deliver!([episode])
```
kookster, line 695: so many stubs! definitely not easy to create this test condition, great to see it.

kookster (approved):
looks good - 2 lines of code, 100 more to test it!

PR #1239, opened 2025-04-15
Text filter flickering

cavis:
Fixes a couple things that always irk me:

1. The podcast/episode text search field always flickers, being overwritten by search results coming back.  Sometimes deleting characters you've typed.
2. The new turbo morphing seems to do a better job with images vs `morphdom`.  So no more flickering (I hope) on podcast#index images as you search.
3. Adds loading spinners to podcasts#index and episodes#index
4. Adds a `/` keyboard shortcut to open the podcast switcher.


app/javascript/controllers/dropdown_select_controller.js
```diff
@@ -0,0 +1,25 @@
+import { Controller } from "@hotwired/stimulus"
+
+export default class extends Controller {
+  connect() {
+    this.element.dataset.action = [this.element.dataset.action, "dropdown-select#select"].filter((v) => v).join(" ")
```
cavis, line 5: Little bit of JS to immediately select things when you change sorting/filtering dropdowns on podcasts#index and episodes#index.

app/views/podcast_switcher/_dropdown.html.erb
```diff
@@ -1,6 +1,6 @@
 <div class="prx-podcast-switcher dropdown" data-controller="focus">
 
-  <button class="btn btn-white dropdown-toggle" type="button" id="podcast-switcher" data-bs-toggle="dropdown" aria-expanded="false" data-action="focus#focus">
+  <button class="btn btn-white dropdown-toggle" type="button" id="podcast-switcher" data-bs-toggle="dropdown" aria-expanded="false" data-action="focus#focus" data-controller="hotkey">
```
cavis, line 3: Brought back the switcher for podcasts#index.  So I can just type `/` and open it!

brandonhundt (approved):
Tested. I think this makes sense. Sunny day. @cavis :shipit:

PR #1240, opened 2025-04-17
Downgrade routing 404 log levels

cavis:
Previously, we used a catchall in `routes.rb` to route/downgrade 404s to infos.

But with the latest version of prx_auth-rails ... routes are automagically mounted rather than explicitly added to routes.rb.  Which resulted in the catchall being _before_ the gem's routes.

Had to get rid of the catchall to make things work.  In this PR, I try just downgrading it in the logger instead.


config/routes.rb
```diff
@@ -62,8 +62,4 @@
 
   match "/api", via: [:get], to: redirect("/api/v1")
   root "podcasts#index"
-
-  # unless Rails.configuration.consider_all_requests_local
-  #   match "*unmatched", via: :all, to: "errors#not_found"
-  # end
```
kookster, line 68: so, I may be off in the weeds, but I think the fix should be to have the routes from prx_auth-rails added explicitly in the routes - either as an engine would, or with a callable method that can be added to routes?

Like the note here https://guides.rubyonrails.org/engines.html#mounting-the-engine

image

kookster (changes requested):
maybe we fix this in prx_auth-rails?

config/routes.rb
cavis: Yeah, that was how it was originally setup.  But I was tired of all the `main_app` references you needed in your layouts.  If you miss one, and a user happens to get auth-refreshed while on that view... 500 error.

(And I figured... who wouldn't want the routes included in the parent app?  The gem doesn't work without them).

May need to re-research some of this then - I'd be fine with the explicit routes `mount` if it didn't cause all those main_app issues.

kookster: maybe like devise, have a method called in the routes.rb to add in the new routes, that way we can position them in the loading?

cavis:
Okay @kookster ... smaller PR now!

kookster (approved):
tiny! looks good

PR #1241, opened 2025-04-19
Add max_duration to API create

kookster:


app/models/megaphone/cuepoint.rb
```diff
@@ -8,33 +8,36 @@ class Cuepoint
...
+    def self.from_placement_and_media(placement, media)
       cuepoints = []
       current_cuepoint = nil
       original_duration = 0
       original_count = 0
-      zones.each do |zone|
+      placement.zones.each do |zone|
         # if this is an ad zone, add it to the cue point
-        if ["ad", "sonic_id", "house"].include?(zone[:type])
+        if ["ad", "house", "sonic_id", "billboard"].include?(zone[:type])
```
cavis, line 27: Might be slightly more future-proof to just check for `"original"`, since you're including everything else.

cavis (approved):
One optional thought on zone-type filtering.  But :+5: looks good to me!

app/models/megaphone/cuepoint.rb
kookster: yah, I had the same thought, then I got lazy - will fix

PR #1245, opened 2025-04-24
Restrict uploads to 1GB

cavis:
Saw a prod issue where someone uploaded a video larger than our postgres 4-byte integer column `file_size` could hold.  (Max is 2147483647... which is like ~2GB).

This PR restricts uploads to 1GB, which you'd think would be plenty.  Largest file in prod is just under 1GB right now.

image

### DIFF
--- a/app/javascript/controllers/upload_controller.js
+++ b/app/javascript/controllers/upload_controller.js
@@ -37,6 +37,10 @@ export default class extends Controller {
     if (this.hasNonMp3WarningTarget && file.type !== "audio/mpeg") {
       this.modal = new bootstrap.Modal(this.nonMp3WarningTarget)
       this.modal.show()
+    } else if (file.size >= 1024 * 1024 * 1024) {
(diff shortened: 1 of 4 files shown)


app/javascript/controllers/upload_controller.js
```diff
@@ -37,6 +37,10 @@ export default class extends Controller {
     if (this.hasNonMp3WarningTarget && file.type !== "audio/mpeg") {
       this.modal = new bootstrap.Modal(this.nonMp3WarningTarget)
       this.modal.show()
+    } else if (file.size >= 1024 * 1024 * 1024) {
+      this.fileNameTargets.forEach((t) => (t.innerHTML = file.name))
+      this.fileSizeTargets.forEach((t) => (t.innerHTML = humanBytes(file.size)))
+      this.onError("File too large - upload less than 1GB")
```
kookster, line 43: should this be in `en.yml` somehow?

kookster (commented):
error message in en.yml?

PR #1246, opened 2025-04-25
Feat/pre transcode new episode drops

### LINKED ISSUE #1135: Pre-transcode new episode drops
The first time anyone requests an episode, it _may_ need to have its originals transcoded.  This can be slow.

Can we somehow trigger that transcode ahead of time?

Maybe just HEAD the episode enclosure url (in the particular feed) as soon as it publishes?

Or as soon as the episode is scheduled?

radical-ube:
resolves #1135 

This PR adds a head request step when an episode is published to the default feed for the first time. This is in order to start the transcoding for the originals ahead of time.


test/models/episode_test.rb
```diff
@@ -381,4 +381,46 @@
...
+    it "does not follow more than 10 redirects" do
+      uri_1 = episode_1.enclosure_url
+
+      stub_request(:head, uri_1)
+        .to_return(status: 302, headers: {location: uri_1}).times(50)
+
+      episode_1.head_request
+
+      assert_requested :head, episode_1.enclosure_url, times: 10
+    end
```
radical-ube, line 424: i was trying to add another test for timeouts to make sure they got rescued properly but i was having trouble figuring it out so i left it off.

app/models/episode.rb
```diff
@@ -281,4 +281,21 @@ def published_or_released_date
   def ready_transcript
     transcript if transcript&.status_complete?
   end
+
+  def head_request(uri_str = enclosure_url(podcast.default_feed), redirects = 0)
```
cavis, line 285: If this method has a lot on common with the `RedirectPrefixValidator#head_request`... you could try to refactor into a shared library?

Probably a class method on a model concern (can a validator include a concern?)  And it would have to take in args: the url, max jumps allowed, and timeout.

In no way required!  Just a thought.

cavis (changes requested):
One edge case around new podcasts, and needs to include a timeout value so we're not just waiting on Dovetail to stitch/transcode the entire episode.

app/jobs/publish_feed_job.rb
```diff
@@ -75,11 +75,23 @@ def save_file(podcast, feed, options = {})
     opts[:bucket] = s3_bucket
     opts[:key] = feed.path
     @put_object = s3_client.put_object(opts)
-    Episode.where(id: rss.episodes).where(first_rss_published_at: nil).update_all(first_rss_published_at: DateTime.now) if feed.default?
+
+    update_first_publish_episodes(podcast.first_publish_episodes) if feed.default?
```
cavis, line 79: Hrm... I don't think this `first_publish_episodes` method on the podcast is going to work.  It's not filtering for episodes we just wrote to the RSS (`rss.episodes`).  So right now, this would immediately set `first_rss_published_at` on any draft/scheduled episodes  as well.

I think you'll need to keep that same `.where(id: rss.episodes)` around.  Either inline in this method, or move `def first_publish_episodes` from the Podcast to here.

app/jobs/publish_feed_job.rb
```diff
@@ -75,11 +75,23 @@ def save_file(podcast, feed, options = {})
     opts[:bucket] = s3_bucket
     opts[:key] = feed.path
     @put_object = s3_client.put_object(opts)
-    Episode.where(id: rss.episodes).where(first_rss_published_at: nil).update_all(first_rss_published_at: DateTime.now) if feed.default?
+
+    update_first_publish_episodes(rss.episodes) if feed.default?
```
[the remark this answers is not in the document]
radical-ube: oops i forgot the `first_publish` scope. hold on, will push up one more change. (although maybe i should be worried that my tests didn't catch that?)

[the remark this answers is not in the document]
cavis: If you made a `def first_publish_episodes` in this class, it would be easy to test.  (And separate tests from `def update_first_publish_episodes`, which can sometimes make it easier to isolate your test cases).

cavis (approved):
👍 alright, this all looks/works good to me!

PR #1250, opened 2025-05-03
Uncut API

kookster:
Allows the uncut audio to be set, replaced, deleted, and updated with a new segmentation value.


app/models/concerns/episode_media.rb
```diff
@@ -133,6 +133,19 @@ def media_ids
...
+    if (uncut && self.uncut) && (uncut.href.blank? || (self.uncut.href == uncut.href))
+      self.uncut.segmentation = uncut.segmentation if uncut.segmentation.present?
+    else
+      super
+    end
+
+    self.medium = if uncut.present?
+      "uncut"
+    end
+  end
```
kookster, line 147: re: https://github.com/PRX/feeder.prx.org/pull/1201#discussion_r2010276485

app/models/concerns/episode_media.rb
```diff
@@ -133,6 +133,19 @@ def media_ids
     media.map(&:id)
   end
 
+  def uncut=(uncut)
+    # if uncut is unchanged or missing, update other attributes
+    if (uncut && self.uncut) && (uncut.href.blank? || (self.uncut.href == uncut.href))
```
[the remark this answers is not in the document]
kookster: yes, that's the idea - I could probably get fancier, and slice only the allowed attributes for update, but here I am mostly interested in what you said - don't replace a file if it's the same url, and do allow updating segmentation.

app/models/concerns/episode_media.rb
kookster: that is what I recall (that would break lots of stuff), I can add a test for it if you're concerned?

kookster: hmm, there was a thing a long time ago I am remembering about patch vs put for this, where fetch could be sparse, but if you had missing attributes on put, they would get nilled out. I'll try it!

app/representers/api/uncut_representer.rb
```diff
@@ -0,0 +1,3 @@
+class Api::UncutRepresenter < Api::MediaResourceRepresenter
+  property :segmentation
```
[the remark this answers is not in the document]
kookster: fair enough, and Rick is using the auth version, so I can move it only be there.

app/models/concerns/episode_media.rb
cavis: Must be an internal bit of represantable, right?  Generally I'm ok with assuming your dependencies do their jobs - mostly trying to remember how it works for my own education!

kookster: it works! I added a check for this in the api/auth/episodes_controller.rb below ...

app/representers/api/uncut_representer.rb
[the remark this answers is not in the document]
kookster: removed this property from the public api representer, uncut only on auth now

test/controllers/api/auth/episodes_controller_test.rb
```diff
@@ -146,6 +146,95 @@ class << @controller; attr_accessor :prx_auth_token; end
...
+      episode_update.reload
+
+      contents = episode_update.contents
+      assert_equal contents.size, 0
+
+      uncut = episode_update.uncut
+      assert_not_nil uncut
+      assert_equal "uncut", episode_update.medium
+
+      # put to update without setting should not delete uncut
```
kookster, line 176: This test below shows that when only title is sent in the body on a put, it does not then delete uncut.

cavis (approved):
:+1: looks good to me

PR #1251, opened 2025-05-03
Set from megaphone podcast ad tags from organization tag options

kookster:
- [x] Add the organization id to the config
- [x] Retrieve tags from the API
- [x] Add advertising tags to the megaphone config
- [x] Set tags from the UI
- [x] Use the tags when creating/updating the podcast

### DIFF
--- a/app/models/feeds/megaphone_feed.rb
+++ b/app/models/feeds/megaphone_feed.rb
@@ -75,4 +75,11 @@ class Feeds::MegaphoneFeed < Feed
   def mark_as_not_delivered!(episode)
     episode.episode_delivery_status(:megaphone, true).mark_as_not_delivered!
   end
+
+  def advertising_tag_options
+    return [] unless persisted? && config.valid?
+    Megaphone::OrganizationTag.list_by_feed(self).map do |tag|
+      [tag.value, tag.label]
+    end
+  end
 end
--- a/app/models/megaphone/config.rb
+++ b/app/models/megaphone/config.rb
@@ -1,5 +1,7 @@
 module Megaphone
   class Config < ApplicationRecord
+    serialize :advertising_tags, coder: JSON
+
     belongs_to :feed
 
     validates_presence_of :token, :network_id, :organization_id
--- a/app/models/megaphone/config.rb
+++ b/app/models/megaphone/config.rb
@@ -2,10 +2,11 @@ module Megaphone
   class Config < ApplicationRecord
     belongs_to :feed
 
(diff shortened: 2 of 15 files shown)


app/models/feeds/megaphone_feed.rb
```diff
@@ -75,4 +75,11 @@ def publish_to_megaphone?
   def mark_as_not_delivered!(episode)
     episode.episode_delivery_status(:megaphone, true).mark_as_not_delivered!
   end
+
+  def advertising_tag_options
+    return [] unless persisted? && config.valid?
+    Megaphone::OrganizationTag.list_by_feed(self).map do |tag|
```
cavis, line 81: Ah, okay - so these exist under the `organization_id` in Megaphone, so we can list them as options rather than manually entering them.  +1.

app/views/feeds/_form_megaphone_config.html.erb
```diff
@@ -30,6 +37,13 @@
             <%= field_help_text t(".help.network_name") %>
           </div>
         </div>
+        <div class="col-12 mb-4">
+          <div class="form-floating input-group">
+            <%= config_form.select :advertising_tags, feed.advertising_tag_options, {include_blank: false}, multiple: true %>
```
cavis, line 42: Are these quick "enough" that we don't need to worry about caching them?

(Seems like this will hit megaphone on every page load... but it's so rare that someone would (A) have a megaphone feed, and (B) be on this page editing things).

cavis (approved):
Couple questions, but +8 looks good to me!

app/models/megaphone/config.rb
```diff
@@ -1,11 +1,14 @@
...
+    serialize :advertising_tags, coder: JSON
+
     belongs_to :feed
 
-    validates_presence_of :token, :network_id
+    validates_presence_of :token, :network_id, :organization_id
 
     encrypts :token
     encrypts :network_id
+    encrypts :organization_id
```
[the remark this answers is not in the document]
kookster: I think "ish" is right - they are in the url, but they aren't really "exposed", i.e. the API assumes you have it, but there is no way to get it except looking at the url while logged in. It seems sensitive enough to go ahead and encrypt, esp as it is useful in combination with the private token.

app/views/feeds/_form_megaphone_config.html.erb
kookster: so far it's been very quick when I test it, so I didn't worry about it.

PR #1252, opened 2025-05-06
Import fixes

### LINKED ISSUE #1007: Handle invalid urls on import, if all they lack is a scheme
Had an import fail b/c the `<link/>` was a url, but missing the "https://" protocol

### LINKED ISSUE #1189: Support more podcast 2.0 tags on import
e.g. https://feeds.transistor.fm/never-post

They have podcast:transcript and podcast:guid tags, which we have now too, but I don't think we set on import?
https://github.com/PRX/feeder.prx.org/blob/main/app/models/imports/podcast_rss_import.rb
https://github.com/PRX/feeder.prx.org/blob/main/app/models/imports/episode_rss_import.rb

kookster:
- [x] Map podcast:guid
- [x] Map podcast:funding 
- [x] Respect podcast:locked
- [x] Try to fix urls w/o a scheme

fixes #1189
fixes #1007

### DIFF
--- a/app/models/concerns/import_utils.rb
+++ b/app/models/concerns/import_utils.rb
@@ -92,6 +92,10 @@ module ImportUtils
     str.strip
   end
 
+  def clean_yes_no(str)
+    (clean_string(str) || "").downcase == "yes"
+  end
+
   def clean_url(url)
     url = clean_string(url)
     return nil if url.blank?
--- a/app/models/concerns/import_utils.rb
+++ b/app/models/concerns/import_utils.rb
@@ -93,7 +93,7 @@ module ImportUtils
   end
 
   def clean_yes_no(str)
-    (clean_string(str) || "").downcase == "yes"
+    ["yes", "true", "1"].include?((clean_string(str) || "").downcase)
   end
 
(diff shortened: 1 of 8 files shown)


app/models/concerns/import_utils.rb
```diff
@@ -92,6 +92,32 @@ def clean_string(str)
     str.strip
   end
 
+  def clean_yes_no(str)
+    ["yes", "true", "1"].include?((clean_string(str) || "").downcase)
+  end
+
+  def clean_url(url)
```
cavis, line 99: +10... these have given us issues from time to time!

app/models/imports/podcast_rss_import.rb
```diff
@@ -73,6 +73,10 @@ def validate_rss
       self.feed_episode_count = feed.entries.count
       channel
       first_entry
+
+      if clean_yes_no(channel[:podcast_locked])
```
cavis, line 77: Nice to respect this tag!

cavis (approved):
:+2: some nice additions/improvements!

PR #1253, opened 2025-05-06
Add podcast guid to podcast and feed api responses

kookster:


app/representers/api/podcast_representer.rb
```diff
@@ -28,6 +28,7 @@ class Api::PodcastRepresenter < Api::BaseRepresenter
   property :locked
   property :payment_pointer
   property :donation_url
+  property :guid
```
kookster, line 31: Adding it here as well, for completeness (and if we want to sync via API we could)

app/representers/api/auth/feed_representer.rb
```diff
@@ -22,6 +22,7 @@ class Api::Auth::FeedRepresenter < Api::BaseRepresenter
   property :include_donation_url
   property :serve_drafts
   property :unique_guids
+  property :guid
```
kookster, line 25: Use this in DTR

cavis (approved):
👍

PR #1259, opened 2025-05-14
Start and end times for cuepoints must be floats in json

kookster:
this is a limited fix, another option would be to change the behavior of BigDecimal as_json to always convert to_f rather than just affecting these specific attributes in this one model

and another option is to use Oj for this, which has an explicit option for bigdecimal_as_decimal https://www.ohler.com/oj/doc/Oj.html


cavis (approved):
👍 good catch!

PR #1264, opened 2025-05-30
Add in explicit require to load rails specific strategy classes

kookster:
trying to fix any more
```
'<main>': uninitialized constant SayWhen::Processor (NameError)
```


cavis (approved):
👍 

I still don't understand how intermittent this problem is.  Or how it got stuck on this error this time.  But can't hurt!

PR #1268, opened 2025-06-03
HOTFIX: Sync global styles (dark mode checkboxes)

### LINKED ISSUE #1267: More contrast on checkboxes in dark mode?
I missed that one of these was not checked (they are both checked now)

Image

Maybe I still would have missed it, but the low contrast between the white check and the grey box didn't help my poor brain.

brandonhundt:
closes #1267 

-syncs global styles with https://github.com/PRX/augury.prx.org/pull/1834
-fixes checkboxes/toggles in darkmode


cavis (approved):
👍

PR #1271, opened 2025-06-17
Adds a metrics tab and skeleton layout

radical-ube:
This PR adds the metrics tab (development only) and a skeleton layout for the tab.

### DIFF
--- a/app/views/layouts/_subnav.html.erb
+++ b/app/views/layouts/_subnav.html.erb
@@ -11,6 +11,9 @@
         <%= active_link_to "Settings", edit_podcast_path(@podcast), active: podcast_settings_active?, class: "nav-link" %>
         <%= active_link_to "Episodes", podcast_episodes_path(@podcast), active: /\/episodes/, class: "nav-link" %>
         <%= active_link_to "Feeds", podcast_feeds_path(@podcast), active: /\/feeds/, class: "nav-link" %>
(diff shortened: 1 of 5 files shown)


app/views/layouts/_subnav.html.erb
```diff
@@ -11,6 +11,9 @@
         <%= active_link_to "Settings", edit_podcast_path(@podcast), active: podcast_settings_active?, class: "nav-link" %>
         <%= active_link_to "Episodes", podcast_episodes_path(@podcast), active: /\/episodes/, class: "nav-link" %>
         <%= active_link_to "Feeds", podcast_feeds_path(@podcast), active: /\/feeds/, class: "nav-link" %>
+        <% if ENV["RAILS_ENV"] == "development" %>
```
cavis, line 14: Or just `Rails.env.development?` ?

config/routes.rb
```diff
@@ -21,6 +21,7 @@
     end
     resources :placements_preview, only: [:show]
     get "rollups_demo", to: "podcasts#rollups_demo"
+    resource :metrics, controller: :podcast_metrics
```
cavis, line 24: I suspect eventually this will be several routes?  But this works just to get things stubbed in.

cavis (approved):
One suggestion on checking rails env, but 👍 looks good to me!

PR #1272, opened 2025-06-17
Handle when the placements api response is missing sections

kookster:
which it shouldn't, in general, but have seen errors for this, so making it more resilient


cavis (approved):
👍

PR #1285, opened 2025-07-04
Add podping to publishing pipeline

kookster:


app/models/concerns/publishing_notify.rb
```diff
@@ -0,0 +1,38 @@
...
+  end
+
+  def podping_enabled?(feed)
+    ENV["PODPING_AUTH_TOKEN"].present? && feed.public?
+  end
+
+  def podping_headers
+    {
+      "User-Agent" => "PRX",
+      "Authorization" => ENV["PODPING_AUTH_TOKEN"]
```
kookster, line 35: we have a PRX specific auth token, and the User-Agent has to be "PRX" to go with it.

app/jobs/publish_feed_job.rb
```diff
@@ -4,6 +4,7 @@ class PublishFeedJob < ApplicationJob
   queue_as :feeder_publishing
 
   include PodcastsHelper
+  include PublishingNotify
```
cavis, line 7: Redundant include?  (`PodcastHelper` also includes `PublishNotify`).

cavis (commented):
One thought on grouping after-rss-is-published hooks.  (Feel free to argue that one).  But otherwise this is looking good.

app/jobs/publish_feed_job.rb
kookster: yeah, that's true - I added it here first, then the helper later, but yeah, I can clean this up.

kookster: fixed!

app/jobs/publish_feed_job.rb
```diff
@@ -63,6 +64,7 @@ def publish_integration(podcast, feed)
   def publish_rss(podcast, feed)
     res = save_file(podcast, feed)
     PublishingPipelineState.publish_rss!(podcast)
+    notify_rss_published(podcast, feed)
```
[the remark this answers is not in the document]
kookster: I added a new hook method for both

PR #1287, opened 2025-07-06
Add import locked for `podcast:import` tag

kookster:
implements #1288

### DIFF
--- a/app/views/podcasts/show.rss.builder
+++ b/app/views/podcasts/show.rss.builder
@@ -110,6 +110,8 @@ xml.rss "xmlns:atom" => "http://www.w3.org/2005/Atom",
       xml.podcast :follow, url: @podcast.subscribe_links_path
     end
 
+    xml.podcast :locked, @feed.import_locked ? "yes" : "no"
+
     @episodes.each_with_index do |ep, index|
       xml.item do
         xml.guid(episode_guid(ep, @feed), isPermaLink: !!ep.is_perma_link)
--- a/db/migrate/20250706201302_add_import_locked_to_feeds.rb
+++ b/db/migrate/20250706201302_add_import_locked_to_feeds.rb
@@ -1,0 +1,6 @@
+class AddImportLockedToFeeds < ActiveRecord::Migration[7.2]
+  def change
+    add_column :feeds, :import_locked, :boolean, default: true, null: false
+    execute "UPDATE feeds SET import_locked = true"
+  end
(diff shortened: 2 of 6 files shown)


app/views/podcasts/show.rss.builder
```diff
@@ -110,6 +110,8 @@ xml.rss "xmlns:atom" => "http://www.w3.org/2005/Atom",
       xml.podcast :follow, url: @podcast.subscribe_links_path
     end
 
+    xml.podcast :locked, @feed.import_locked ? "yes" : "no"
```
cavis, line 113: Thinking of cases where we'd lock some Feeds, but not others.

Maybe we'd make a new private RSS that had both public-RSS and bonus episodes, let the new host import that, but keep the real feeds locked that entire time?

Or can you think of others?

db/migrate/20250706201302_add_import_locked_to_feeds.rb
```diff
@@ -0,0 +1,6 @@
+class AddImportLockedToFeeds < ActiveRecord::Migration[7.2]
+  def change
+    add_column :feeds, :import_locked, :boolean, default: true, null: false
+    execute "UPDATE feeds SET import_locked = true"
```
cavis, line 4: It's kind of confusing to me that `podcast.locked` is named so similarly.  Maybe we'll change that to `write_locked` or something someday.  Ah, naming-tech-debt!

cavis (approved):
👍 looks good to me!

app/views/podcasts/show.rss.builder
kookster: @brandonhundt I put this `podcast:locked` checkbox at the feed level, not the podcast level?
As @cavis is pointing out, that's perhaps unnecessary, as when I go to move a feed I might move all the feeds?

I was also thinking in the process of a feed transition, we might create an ad-free private feed that is unlocked, while keeping the main feed locked, so it made sense to have at the feed level for that common transition case?

cavis: (I actually think it's a good idea to put it at the Feed level... just thinking through cases for it).

db/migrate/20250706201302_add_import_locked_to_feeds.rb
cavis: Agreed - out of scope for this one.  Opened #1290.

PR #1309, opened 2025-07-17
Do not send podping for blank urls

### LINKED ISSUE #1308: Do not call podping if the feed url is blank

kookster:
fixes #1308


cavis (approved):
👍 likely fixes a couple ops-errors logged over the weekend.

PR #1311, opened 2025-07-22
Handle non-parseable urls

cavis:
Was seeing an API error, where the original_url of media got set to `"<br/>"` (it was from a bad RSS import).  And URI.parse was erroring out.


kookster (approved):
Looks good - I wonder about even creating media resources if the url is bad, but this helps.

cavis:
Yeah, I was also surprised to see we don't have a format validator on `original_url`.

PR #1313, opened 2025-07-27
Default podcast values for megaphone import

### LINKED ISSUE #1314: Megaphone has different requirements on create and update for some attributes
A number of attributes are required but only on create, not on update?

For import from megaphone, this causes issues as the megaphone models validate for required fields that are not actually or always required, and may have been empty on import - `subtitle` for podcasts being one.

kookster:
fixes #1314


cavis (approved):
Ah, missed this one!  👍 looks good to me

Issue #1298, opened 2025-07-10
metrics > feeder: date selector, timezone parsing


radical-ube:
- [x] [date presets](https://github.com/PRX/feeder.prx.org/pull/1318#discussion_r2291849168) label
- [ ] [date selector container](https://github.com/PRX/feeder.prx.org/pull/1318#discussion_r2291875184) approach
- [ ] [revisit](https://github.com/PRX/feeder.prx.org/pull/1318/files#r2308331421) `generate_date_range` method
- [x] date range [preset options](https://github.com/PRX/feeder.prx.org/pull/1318/files#r2308356133)
- [x] date selection [needs to account for UTC](https://github.com/PRX/feeder.prx.org/pull/1318/files#r2308370773)

PR #1318, opened 2025-08-14
Feat/add uniques and dropday tabs

### LINKED ISSUE #1292: metrics > feeder parity: uniques, dropday, and main card tabs

### LINKED ISSUE #1296: metrics > feeder: lazy loading turbo frames

radical-ube:
resolves #1292 
resolves #1296 

This PR's main goal is to create the Downloads `main_card`, which tabs between 4 different downloads views, and the respective sub-selections within each view. The different views on the `main_card` are now separated into different controller routes, and lazy loaded into their respective turbo frames. Each view is hidden from view until their respective tab is selected, which is controlled by the `apex_nav_controller.js`. Many apex utilities are abstracted away into the `utils/apex.js` file, in efforts to simplify each view's own `apex_(view)_controller.js` and reuse code where possible.

As an aside, the agents views and controller route have been commented out for simplicity and will be addressed in a different PR.


app/controllers/podcast_metrics_controller.rb
```diff
@@ -1,112 +1,273 @@
 class PodcastMetricsController < ApplicationController
   before_action :set_podcast
+  before_action :set_date_range, only: %i[show downloads episodes uniques]
```
radical-ube, line 3: everything except `dropdays` uses the user-specified date range.

app/controllers/podcast_metrics_controller.rb
```diff
@@ -1,112 +1,273 @@
...
 
   def show
-    authorize @podcast, :show?
+  end
+
+  def downloads
```
radical-ube, line 8: in the previous PR, all queries were made in the `show` route, and now all the main downloads tabs have been moved to their own respective route/action. this is the main `downloads` route, which is for the podcast's total downloads within the selected date range.

app/controllers/podcast_metrics_controller.rb
```diff
@@ -1,112 +1,273 @@
...
+    }
+
+    render partial: "downloads_card", locals: {
+      interval: @interval,
+      date_range: @date_range,
+      downloads: @downloads
+    }
+  end
 
+  def episodes
```
radical-ube, line 31: `episodes` is the downloads aggregated by episode, paginated in groups of 10.

app/controllers/podcast_metrics_controller.rb
```diff
@@ -1,112 +1,273 @@
...
+  end
+
+  def dropdays
+    @episodes =
+      @podcast.episodes
+        .published
+        .order(first_rss_published_at: :desc)
+        .paginate(params[:episode_dropdays], params[:per])
+
+    @dropdays = @episodes.map do |ep|
```
radical-ube, line 97: since dropdays relies on a date range relative to the episode, i'm currently looping through the selected episodes and running the query for each one. i do think building subqueries and running them through a single query will improve efficiency and performance, but for a first pass iteration i believe this is sufficient and the current result is not too slow.

app/controllers/podcast_metrics_controller.rb
```diff
@@ -96,17 +197,77 @@ def metrics_params
...
+
+    while start_range + i.send(:"#{interval.downcase.pluralize}") <= end_range
+      range << start_range + i.send(:"#{interval.downcase.pluralize}")
+      i += 1
+    end
+
+    range
+  end
+
+  def colors
```
radical-ube, line 259: colors were moved from the helper to here. i wanted to make sure they were just in one place but each of the different charts were pulling from the same set.

app/javascript/controllers/apex_agents_controller.js
```diff
@@ -0,0 +1,53 @@
+import { Controller } from "@hotwired/stimulus"
```
radical-ube, line 1: agents will be fleshed out in a different PR. just setting up the controller here.

app/views/podcast_metrics/_main_card.html.erb
```diff
@@ -0,0 +1,30 @@
+<%= turbo_frame_tag "main_card" do %>
```
radical-ube, line 1: this `main_card` was the goal for this PR. the `card-header` houses the tabs and on the right-hand side will display/hide a dropdown specific to whichever card is displayed at the moment.

app/views/podcast_metrics/_dropdays_card.html.erb
farski, line 1: It's not super clear which episodes are being included in the drop day chart. I think it's the most recent 10? Changing the date range doesn't seem to change which episodes are included, which is maybe okay, but it's a bit confusing. Bare minimum I think would be to include somewhere e.g., "latest 10 episodes" (if that's what it is)

farski, line 1: For the sake of consistency with Insights and the Data Explorer, I think we should have a clear standard for how these episode lifecycle calculations are done.

The way the Data Explorer works (and/because that's how I think it _should_ work), is that the clock starts ticking from the episode's drop date/time. If an episode drops at noon, it's "Day 1" value will be from the moment it dropped until noon the next day (the first `24 * 60 * 60` seconds of the episode's life). It's "Day 7" value would be `24 * 60 * 60 * 7`. It's "Week 1" value would _also_ be `24 * 60 * 60 * 7`. 

I think Metrics used calendar days, meaning for an episode that dropped at noon, it's "Day 0" value would only include 12 hours worth of downloads.

I'm pushing for two changes to be made there:
1. Standardizing on relative calculations and not using any calendar days/weeks/etc for drop day. Everything starts from the drop, and is N hours/days/weeks _worth of time_ from then, but they are just shorthands for a bunch of seconds.
2. Stop using a "Day/Week/Month 0". Currently we have the graph saying "0 months since drop, this episode got 50,000 downloads", when it really means, literally, "1 month since drop, this episode got 50,000 downloads". I think for clarity, we should retire the use of the 0th period on this charts. If we want to include a 0 on the X-axis, I think, by definition, the series value should always be 0 at that point.

I think in Metrics there was maybe an argument for it, because the first data point didn't necessarily represent a full day. But sometimes it did, and sometimes it nearly did. That's why I think resolving both of these potential confusion points at once would be worth it.

farski, line 1: For me, sometimes changing dropday range values takes several seconds to resolve, but there's no visual indicator that anything is happening. At one point i just kept clicking and then I saw the graph keep redrawing itself over and over for a good 30 seconds.

farski, line 1: In Metrics we used "drop day", and now we're using "dropday". Just making sure that deliberate.

farski, line 1: In Metrics, I don't really understand why we offer daily/weekly/monthly options for drop day. The values have always been very hard to sort out. For example, looking at the **daily** chart, an episode may get to `9,121` on day 43, but on the **monthly** chart it gets to `9,121` at month `1`. I still don't entirely know how to predict how that will happen for any given episode.

I also think those options introduce an unnecessary and unhelpful level of smoothing to the data. If an episode has a spike between day 30 and day 60, we draw it as a straight line. I don't really get why we throw that data away, even though we have it and could display it without making things any more difficult for the user.

My preference would be to always draw these lines with at most 24 hours of data between data points, and if we want to give the user an option for "weekly" or other values, we keep drawing the lines with that resolution, and do something to highlight the periods they've asked for (e.g., circles at 7/14/21/28 days to make it easy to find the "weeks").

I'd like to be careful about using "month" in the context of drop day. What is it? 28 days?

farski, line 1: Image

I think we should be really cautious about how the line smoothing we're using affects the way it's representing the underlying data. This line is showing rate changes that may or may not actually exist. It's saying that at the start of Day 0, downloads were flat, then they ramped up halfway between Day 0 and Day 1, and then right around day 1 they flattened out again to a rate of 0. 

Image

My suspicion is that in this time range, the rate was actually pretty constant. I don't think there was a chunk of time ~1 day after the episode dropped that it got zero downloads, but that's what the line is showing.

But either way, we aren't generating the data to know, even if that's the case. We are only generating these daily/24-hour values, and the smoothing that's being applied is introducing an artificial tweening of the data at a finer resolution that I would consider to be inaccurate, and I think we should reconsider how we're handling that.

app/views/podcast_metrics/_episodes_card.html.erb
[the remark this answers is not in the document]
farski: I have found some cases where the tooltip isn't quite a big, but it's still big, and it's not uncommon for it to be this big

farski (changes requested):

app/views/podcast_metrics/show.html.erb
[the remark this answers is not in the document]
radical-ube: can this be a discussion with @brandonhundt ? it sounds like there are some fruitful possibilities to explore and i'm happen to listen in as the developer, though it sounds a bit higher level design-wise and i generally don't think i would have much opinion at that level.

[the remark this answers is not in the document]
radical-ube: potentially an `xaxis.tooltip` thing apex does between line and bar types.

app/views/podcast_metrics/_episodes_card.html.erb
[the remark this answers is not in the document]
radical-ube: i don't really like the tooltip as is. is there a fixed placement for it that would be better? in Metrics it's always to the right of the cursor until it hits the end of the chart. i'm not sure where i would put it in Feeder since it's so big. left of the chart?

app/views/podcast_metrics/_dropdays_card.html.erb
radical-ube: and it sounds like by implementing 1, then 2 becomes possible because that first data point becomes a "full" data point.

farski: > ryan and i have talked about 1. and [i was trying to implement it](https://prx.slack.com/archives/G0FKLK2Q5/p1752612071310829?thread_ts=1752606621.836199&cid=G0FKLK2Q5) for `dropdays` in this PR but i kept running into problems with the sql working because i couldn't get the interpolation to work properly.

Can you say more about what you're trying to interpolate?

app/views/podcast_metrics/show.html.erb
farski, line 1: I think it would be good to improve another aspect of our tooltips, over what Metrics currently does.

In Metrics, when there are multiple series in a viz, and you get a tooltip, it will include the data points for all series that exist for that time:

Image

However, the tooltip looks the same regardless of which series you are actually pointing at with your mouse.

I think we should make the connection between the data currently being pointed at with the mouse, and the data in the tooltip more direct.

There are two ways I generally see this handled.

One is to keep all the values in the tooltip, but make some visual distinction for the one that the mouse is currently interacting with:

https://github.com/user-attachments/assets/6f0641a1-3bee-4732-87c2-bbda4a328cbd

The other is to only show the value for the particular series the mouse is interacting with, and if the mouse isn't close enough to be considered interacting with any particular series, then the tooltip includes all values for the X-axis value the mouse is within.

https://github.com/user-attachments/assets/a3d26d2d-d04f-4a5d-96eb-25fd96344d2a

app/javascript/controllers/apex_dropdays_controller.js
```diff
@@ -0,0 +1,45 @@
...
+        return {
+          name: d.episode.title,
+          data: this.normalizeDropdayDownloads(d.rollups),
+          color: d.color,
+        }
+      })
+    }
+  }
+
+  normalizeDropdayDownloads(downloads) {
```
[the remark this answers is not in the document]
radical-ube: adjusted to nullify the value instead of 0.

app/javascript/controllers/apex_nav_controller.js
```diff
@@ -0,0 +1,21 @@
+import { Controller } from "@hotwired/stimulus"
+
+export default class extends Controller {
```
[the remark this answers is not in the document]
radical-ube: added more methods to make sure the cards update independently, and so that changing dates affects all cards without resetting the entire page (as in a previous commit).

app/views/podcast_metrics/_filter_dates.html.erb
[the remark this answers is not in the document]
radical-ube: resolved. user selections should now remain as how they are selected, even when tabbing and changing dates.

app/views/podcast_metrics/show.html.erb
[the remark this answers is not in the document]
radical-ube: captured in #1333

app/views/podcast_metrics/_filter_dates.html.erb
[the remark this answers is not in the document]
radical-ube: captured in #1298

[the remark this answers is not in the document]
radical-ube: captured in #1298

app/views/podcast_metrics/_episodes_card.html.erb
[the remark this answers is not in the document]
radical-ube: captured in #1333

app/views/podcast_metrics/_dropdays_card.html.erb
radical-ube: captured in #1338

radical-ube: captured in #1337

app/views/podcast_metrics/show.html.erb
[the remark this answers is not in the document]
radical-ube: captured in #1334

radical-ube: captured in #1333

app/controllers/podcast_metrics_controller.rb
```diff
@@ -96,17 +224,84 @@ def metrics_params
...
+  def generate_date_range(date_start, date_end, interval)
+    start_range = date_start.to_datetime.utc.send(:"beginning_of_#{interval.downcase}")
+    end_range = date_end.to_datetime.utc.send(:"beginning_of_#{interval.downcase}")
+    range = []
+    i = 0
+
+    while start_range + i.send(:"#{interval.downcase.pluralize}") <= end_range
+      range << start_range + i.send(:"#{interval.downcase.pluralize}")
+      i += 1
+    end
```
cavis, line 284: While loops do sometimes make me nervous.  But, this seems to work pretty well.

If this does became slow someday, I'd probably look into something more like the [DateTime.step](https://ruby-doc.org/stdlib-2.6.3/libdoc/date/rdoc/Date.html#method-i-step) method.  But, probably need more than a few hundred to need any optimizing.

app/helpers/podcast_metrics_helper.rb
```diff
@@ -16,81 +12,49 @@ def sum_rollups(rollups)
...
+      ["Last 7 Days", [7.days.ago.utc.to_date, Time.zone.now.utc.to_date].to_json],
+      ["Week to Date", [Time.zone.now.utc.beginning_of_week.to_date, Time.zone.now.utc.to_date].to_json],
+      ["Last 14 Days", [14.days.ago.utc.to_date, Time.zone.now.utc.to_date].to_json],
+      ["Last 28 Days", [28.days.ago.utc.to_date, Time.zone.now.utc.to_date].to_json],
+      ["Last Month", [1.month.ago.utc.to_date, Time.zone.now.utc.to_date].to_json],
+      ["Month to Date", [Time.zone.now.utc.beginning_of_month.to_date, Time.zone.now.utc.to_date].to_json],
+      ["Last 3 Months", [3.months.ago.utc.to_date, Time.zone.now.utc.to_date].to_json],
+      ["Last 6 Months", [6.months.ago.utc.to_date, Time.zone.now.utc.to_date].to_json],
+      ["Last Year", [1.year.ago.utc.to_date, Time.zone.now.utc.to_date].to_json],
+      ["Year to Date", [Time.zone.now.utc.beginning_of_year.to_date, Time.zone.now.utc.to_date].to_json]
```
cavis, line 35: So I think all these `.zone`s are unnecessary? I might suggest DRY-ing this up a bit by:

- Throwing the utc_today used all over the place as a var up top
- Or do something like I did [in spots](https://github.com/PRX/dovetail-spots/blob/main/config/initializers/extensions.rb#L27), and add a `Date.utc_today` method
  - Then you can change things like `Time.zone.now.utc.beginning_of_week.to_date` into `utc_today.beginning_of_year` or `Date.utc_today.beginning_of_year`
- Do a `.map` over this thing to do the `.to_json` calls from every line.  And maybe to add the 2nd param to them all, which seems to always be `utc_today`.

Would also be nice to translate these at some point.  But I'm okay if all the i18n stuff comes later.

app/javascript/controllers/apex_dropdays_controller.js
```diff
@@ -0,0 +1,58 @@
...
+    seriesData: Array,
+    range: Number,
+    interval: String,
+  }
+
+  static targets = ["chart"]
+
+  connect() {
+    const series = this.buildSeries()
+    const title = `${this.intervalValue.toLowerCase()}s since episode drop`
```
cavis, line 16: Yeah, lots of stuff that could be i18n'd at some point, and passed in as stimulus values.  Seems like a good separate-PR task though.  This all works to get started.

app/views/podcast_metrics/_downloads_card.html.erb
```diff
@@ -1,66 +1,17 @@
...
-                </tbody>
-              </table>
-              <div class="col my-4">
-                <%= paginate @episodes, param_name: :episode_rollups %>
-              </div>
-            </div>
-          </div>
-        </div>
-      </div>
+  <%= form_with(url: downloads_podcast_metrics_path(podcast: podcast, date_start: date_start, date_end: date_end, interval: interval), method: :get, id: "downloads_metrics", data: {controller: "click", click_debounce_value: 200}) do |form| %>
```
cavis, line 2: As I think one of your earlier comments mentioned... I wonder how we want to treat browser history on these pages?  May or may not want query params showing up there, and advancing history state so the back/forward buttons work.  An issue for another day!

app/views/podcast_metrics/_dropdays_card.html.erb
```diff
@@ -0,0 +1,49 @@
+<% chart_id = SecureRandom.uuid %>
```
cavis, line 1: Out of curiousity... why so random?  (And not just `chart_id = "dropdays_card"` or something).

app/views/podcast_metrics/_filter_dates.html.erb
```diff
@@ -1,22 +1,26 @@
...
-          <%= option[0] %>
-        </button>
-      </li>
-    <% end %>
-  </ul>
-</div>
+<%= form_with(url: podcast_metrics_path(podcast: podcast, date_start: date_start, date_end: date_end, interval: interval), method: :get, id: "metrics") do |form| %>
+  <div class="row d-flex gx-1">
+    <div class="col form-floating input-group">
+      <%= form.date_field :date_start, min: DateTime.now - 18.months, max: DateTime.now - 1.day, data: {apex_nav_target: "mainStart", action: "apex-nav#updateStartDate"}, class: "form-control form-control-sm", value: date_start %>
```
cavis, line 4: These don't seem to account for UTC?  (Maybe another use for a `Date.utc_today` extension).

config/routes.rb
```diff
@@ -21,7 +21,13 @@
     end
     resources :placements_preview, only: [:show]
     get "rollups_demo", to: "podcasts#rollups_demo"
-    resource :metrics, controller: :podcast_metrics
+    resource :metrics, controller: :podcast_metrics do
```
cavis, line 24: Should probably throw an `only: :show` on here.  So you don't get the full CRUD routes.

cavis (changes requested):
Looking good, the code layout/organization is really coming together.

Had a few suggestions, and a couple time-zone-y bugs.  But besides that, it would be good to open tickets to later figure out:

1. Browser history and query param changes
2. Translations/i18n

app/helpers/podcast_metrics_helper.rb
```diff
@@ -16,81 +12,47 @@ def sum_rollups(rollups)
...
-      ]
-    else
-      [
-        ["Hourly", "HOUR"],
-        ["Daily", "DAY"],
-        ["Weekly", "WEEK"],
-        ["Monthly", "MONTH"]
-      ]
-    end
+  def interval_options
```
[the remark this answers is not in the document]
radical-ube: captured in #1337

app/views/podcast_metrics/_downloads_card.html.erb
radical-ube: captured in #1337

app/views/podcast_metrics/_filter_dates.html.erb
radical-ube: captured in #1298

PR #1323, opened 2025-08-15
Megaphone Fixes III: Fixing the fixed fixes

kookster:
- [ ] Remove expiration from DTR audio urls
- [ ] Handle adfree episodes by sending no cuepoints, pre/post counts 0, adhash ""

### DIFF
--- a/app/models/integrations/base/episode.rb
+++ b/app/models/integrations/base/episode.rb
@@ -15,6 +15,10 @@ module Integrations
         raise NotImplementedError, "Subclasses must implement archived?"
       end
 
+      def ad_free?
+        feeder_episode.categories.include?("adfree")
+      end
+
       def video_content_type?
         feeder_episode.video_content_type?
       end
--- a/app/models/megaphone/episode.rb
+++ b/app/models/megaphone/episode.rb
@@ -98,8 +98,7 @@ module Megaphone
         background_image_file_url: e.ready_image&.href,
         episode_type: e.itunes_type,
         episode_number: e.episode_number,
-        season_number: e.season_number,
-        ad_free: e.categories.include?("adfree")
+        season_number: e.season_number
       }
     end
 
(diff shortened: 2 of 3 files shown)


app/models/megaphone/episode.rb
```diff
@@ -98,8 +98,7 @@ def self.attributes_from_episode(e, feed)
         background_image_file_url: e.ready_image&.href,
         episode_type: e.itunes_type,
         episode_number: e.episode_number,
-        season_number: e.season_number,
-        ad_free: e.categories.include?("adfree")
```
kookster, line 102: no point in setting this - I think it is actually informational? like, if it is adFree (no ads setup), then this is true, but setting it is a no-op

app/models/megaphone/episode.rb
```diff
@@ -260,17 +259,25 @@ def set_external_media
...
-
-      # create cuepoint instances from that
-      cuepoints = Megaphone::Cuepoint.from_placement_and_media(placement, media)
+      cuepoints = get_cuepoints || []
 
       # put those as a list to the mp api
       cuepoints_batch!(cuepoints)
     end
 
+    def get_cuepoints
```
kookster, line 268: Made it a standalone method for testing, and to set to [] when adfree

app/models/megaphone/episode.rb
```diff
@@ -350,15 +357,20 @@ def delivery_status(with_default = false)
     end
 
     def set_placement_attributes
-      if (zones = get_placement_zones(feeder_episode.segment_count))
+      self.expected_adhash = ""
+      self.pre_count = 0
+      self.post_count = 0
+
+      if !ad_free?
+        zones = get_placement_zones(feeder_episode.segment_count)
```
kookster, line 365: only bother getting placements and setting these when it has ads

app/models/megaphone/episode.rb
```diff
@@ -431,7 +443,8 @@ def get_media_info(enclosure)
     end
 
     def arrangement_version_url(location, media_version)
-      uri = URI.parse(location)
+      uri = Addressable::URI.parse(location)
+      uri.query_values = uri.query_values&.except("exp")
```
kookster, line 447: gotta love Addressable, it's just better

app/models/integrations/base/episode.rb
```diff
@@ -15,6 +15,10 @@ def archived?
         raise NotImplementedError, "Subclasses must implement archived?"
       end
 
+      def ad_free?
+        feeder_episode.categories.include?("adfree")
```
cavis, line 19: Also have `paidadfree` and `houseadfree` ... but those are seldom used.

cavis (approved):
👍 looks good to me!

PR #1326, opened 2025-08-19
Change label, and do not default to removing prefix

### LINKED ISSUE #1325: Label/help text  of `enclosure_override_prefix` is opposite what it does
enclosure_override_prefix is also set to true when it should be false for megaphone integration default.

kookster:
Fixes #1325

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -349,7 +349,7 @@ en:
         categories: Categories
         clean_title: Clean Title
         description: Description
-        enclosure_override_prefix: Use Enclosure Override URL as is? (i.e. no prefix?)
+        enclosure_override_prefix: Include Prefix?
         enclosure_override_url: Enclosure Override URL
         embed_player_iframe: Embeddable Player IFrame
         embed_player_type: Player Style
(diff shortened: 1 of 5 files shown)


config/locales/en.yml
```diff
@@ -349,7 +349,7 @@ en:
         categories: Categories
         clean_title: Clean Title
         description: Description
-        enclosure_override_prefix: Include Prefix?
+        enclosure_override_prefix: Use Enclosure Override URL as is? (i.e. no prefix?)
```
cavis, line 352: Hrm... this makes things more confusing to me.

So there's a field called `enclosure_override_url`.  My interpretation of this is if `enclosure_override_prefix?` is true, you should enclosure the prefix on the enclosure override.

Could we just flip the logic in the [podcast_episode_url](https://github.com/PRX/feeder.prx.org/blob/main/app/models/enclosure_url_builder.rb#L30) instead?  Because we also don't want to exclude the prefix from a _non overridden_ enclosure, because of this boolean, right?

More like:

```ruby
if episode.enclosure_override_url.blank?
  url = base_enclosure_url(podcast, episode, feed)
  enclosure_prefix_url(url, prefix)
elsif episode.enclosure_override_prefix?
  enclosure_prefix_url(episode.enclosure_override_url, prefix)
else
  episode.enclosure_override_url
end
```

cavis (changes requested):
I guess I'd rather we flip the logic instead of labeling it as the inverse?  Mainly because I don't think this should impact non-overridden enclosures.

PR #1330, opened 2025-08-20
Vary DTR user agent

cavis:
Avoid hitting the recent-decision cache when requesting DTR mp3s.  To make sure the media-version is accurate.

Related to https://github.com/PRX/dovetail-router.prx.org/issues/375


kookster (approved):
Lgtm!

PR #1342, opened 2025-09-03
Faster latest-media-version for episodes#show

cavis:
The show endpoints are lagging!

### DIFF
--- a/app/models/concerns/episode_media.rb
+++ b/app/models/concerns/episode_media.rb
@@ -94,7 +94,13 @@ module EpisodeMedia
   end
 
   def cut_media_version!
-    latest_version = latest_media_version[-1]
+    latest_version =
+      if latest_media_version.loaded?
+        latest_media_version[-1]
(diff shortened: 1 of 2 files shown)


app/models/concerns/episode_media.rb
```diff
@@ -94,7 +94,13 @@ def feed_ready?
   end
 
   def cut_media_version!
-    latest_version = latest_media_version[-1]
+    latest_version =
+      if latest_media_version.loaded?
+        latest_media_version[-1]
+      else
+        media_versions.latest(id).first
```
cavis, line 101: Well this isn't ideal.  I tried defining the association as:

```ruby
has_many :latest_media_version, ->(ep) { latest(ep) }, class_name: "MediaVersion"
```

But, that just ends up being an N+1 on the episodes#index again.

Probably some better solution involving a join... but that's wednesday morning ryan's problem.

PR #1343, opened 2025-09-03
Even better latest_media_version query

cavis:
Rely on aliases instead of using different SQL for index/show.

In other words... this works for both `Episode.find(...).latest_media_version` and `Episode.limit(100).includes(:latest_media_version)`.


cavis:
Thanks for the tip @svevang !

PR #1351, opened 2025-09-22
Feat/metrics agents cards

### LINKED ISSUE #1294: metrics > feeder parity: agents queries with date/time pagination

radical-ube:
resolves #1294

This PR adds the `agents_card` to both the Podcast Metrics and Episode Metrics views. There is a slight deviation from the original design of separate agents cards, to following the pattern of the `main_card`, in which separate tabs house the different charts/tables of each agent grouping.


app/javascript/controllers/apex_nav_controller.js
```diff
@@ -51,30 +63,40 @@ export default class extends Controller {
...
-        target.addEventListener("change", function () {
-          target.value = value
-        })
+        target.value = value
+        target.dispatchEvent(new Event("change"))
+      }
+    })
+  }
+
+  updateInputs(event) {
```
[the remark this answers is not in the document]
radical-ube: (this is actually still buggy and i'm in the middle of solving for the extra event looping that's occurring)

app/controllers/concerns/metrics_queries.rb
```diff
@@ -0,0 +1,48 @@
+require "active_support/concern"
+
+module MetricsQueries
```
radical-ube, line 3: new concern where i've moved some reusable queries. wanted to keep these separate from the other `MetricsUtils` just to keep things somewhat organized.

app/controllers/podcast_metrics_controller.rb
```diff
@@ -71,9 +66,9 @@ def uniques
     @uniques_rollups =
       Rollups::DailyUnique
         .where(podcast_id: @podcast.id, day: (@date_start..@date_end))
-        .select("DATE_TRUNC('#{@interval}', day) AS day, MAX(#{@uniques_selection}) AS #{@uniques_selection}")
-        .group("DATE_TRUNC('#{@interval}', day) AS day")
-        .order(Arel.sql("DATE_TRUNC('#{@interval}', day) ASC"))
+        .select("DATE_TRUNC('#{@uniques_interval}', day) AS day, MAX(#{@uniques_selection}) AS #{@uniques_selection}")
```
radical-ube, line 69: Uniques on it's own `uniques_interval` param so that it doesn't get updated by the global `interval`.

app/controllers/podcast_metrics_controller.rb
```diff
@@ -97,22 +92,12 @@ def dropdays
...
-        Rollups::HourlyDownload
-          .where(episode_id: ep[:guid], hour: (ep.first_rss_published_at..(ep.first_rss_published_at + @fudged_range.send(:"#{@interval.downcase}s"))))
-          .select(:episode_id, "DATE_TRUNC('#{@interval}', hour) AS hour", "SUM(count) AS count")
-          .group(:episode_id, "DATE_TRUNC('#{@interval}', hour) AS hour")
-          .order(Arel.sql("DATE_TRUNC('#{@interval}', hour) ASC"))
-          .load_async
+        date_start = ep.first_rss_published_at
+        date_end = ep.first_rss_published_at + (@dropday_range.to_i - 1).days
+
+        downloads_date_range_query(ep, date_start, date_end, "DAY")
```
radical-ube, line 100: keeping dropdays as "DAY" interval for now and removed any options for otherwise.

brandonhundt:
@radical-ube I really appreciate you found a way to make the Interval selector global to the page. Thank you! That's working really well.

Also, I just noticed that in the tables for the agent cards, there aren't checkboxes to deselect items in the chart. Is that something we can add (like we do for episodes) to remain consistent? 

StaticShot_30-09-2025_10-29-44

app/controllers/podcast_metrics_controller.rb
```diff
@@ -204,12 +209,28 @@ def set_date_range
...
     @dropday_range = dropdays_params[:dropday_range]
   end
 
+  def set_total_agents
+    @total_agents =
+      Rollups::DailyAgent
+        .where(podcast_id: @podcast.id)
+        .select("SUM(count) AS count")
+        .load_async
+    @totals_in_range =
```
cavis, line 225: Couple things on these...

I think this needs to include the `Other` datapoint (which is distinct from `Unknown` / NULL).  Without that, the chart can be pretty misleading.  I can find several shows in prod that get more than 10% of their downloads from Apps outside of their top 10.

So, we could do either top 9 plus "other", or top 10 plus the 11th "other.

On the plus side, that saves you having to re-query for the totals here.  Instead you move that array of top-10 IF into the where clause.  In raw sql (_probably_ possible to Activerecord this?):

```sql
SELECT
  podcast_id,
  IF(agent_name_id IN (25, 77, 14, 16, 142, 0, 18, 8, 29, 94), agent_name_id, -1) AS agent_name_id,
  SUM(count) AS num
FROM daily_agents
WHERE podcast_id = 93
AND day >= '2025-09-01'
GROUP BY podcast_id, agent_name_id
ORDER BY num DESC
```

That yields 11 results, with `-1` being the "other", and `0` being "unknown":

```
    ┌─podcast_id─┬─agent_name_id─┬────num─┐
 1. │         93 │            25 │ 323427 │
 2. │         93 │            77 │ 137153 │
 3. │         93 │            14 │  83435 │
 4. │         93 │            -1 │  41043 │
 5. │         93 │            16 │  32769 │
 6. │         93 │           142 │  13153 │
 7. │         93 │             0 │  10851 │
 8. │         93 │            18 │  10110 │
 9. │         93 │             8 │   9846 │
10. │         93 │            29 │   8560 │
11. │         93 │            94 │   4425 │
    └────────────┴───────────────┴────────┘
```

cavis (changes requested):
This is looking really good!  Code organization and functionality definitely works for me.

I think my only ask is around how you generate the "Top 10" list... and I think it should really be 11 (including "other"... which could be a substantial amount of the total).

When you get to geos, we'll definitely need the 11th "other" category.  So seems good to include it here as well.

PR #1356, opened 2025-09-30
Fixup multiple calls to archive and unarchive

svevang:
Fixes a problem where we're iterating over chunks but referencing the captured reference to the whole set.

### DIFF
--- a/test/models/apple/publisher_test.rb
+++ b/test/models/apple/publisher_test.rb
@@ -315,6 +315,62 @@ describe Apple::Publisher do
         end
       end
     end
+
+    describe "#archive! chunking behavior" do
+      it "should pass chunked episodes to Apple::Episode.archive, not all episodes" do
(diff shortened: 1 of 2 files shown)


test/models/apple/publisher_test.rb
```diff
@@ -315,6 +315,62 @@
...
+        end
+
+        Apple::Episode.stub(:archive, archive_mock) do
+          apple_publisher.archive!(episodes)
+        end
+
+        # Should have been called twice: first chunk of 25, second chunk of 5
+        assert_equal 2, archive_calls.length
+        assert_equal 25, archive_calls[0].length
+        assert_equal 5, archive_calls[1].length
```
kookster, line 338: nice, shows that chunking at work!

kookster (approved):
lgtm!

PR #1360, opened 2025-10-03
Add apple asset processing duration logging

svevang:
Adjusts some logging to allow for a uniform logging of the asset wait times.

Notable changes:

 - Adds tags to some important methods in the apple publisher flow
 - Adds a tag to the publish feed job. Hopefully makes tailing the logs across publishing attempts possible
 - Sets up uniform logging around the asset processing time, all episodes are logged the same way
 - Removed the separate `reset_asset_wait!` method - this functionality is now handled within `mark_as_delivered!` -- no dupe calls.

Tests are fixed up to cover the `mark_as_delivered!` as the single point of reset. Replace multiple redundant resets with a single.


app/models/apple/publisher.rb
```diff
@@ -107,41 +107,49 @@ def upload_and_process!(eps)
...
-      # Mark episodes as delivered as they are processed
-      wait_for_asset_state(eps) do |ready_eps|
-        mark_as_delivered!(ready_eps)
-        reset_asset_wait!(ready_eps)
+        wait_for_upload_processing(eps)
+
+        # Wait for the audio asset to be processed by Apple
+        # Mark episodes as delivered as they are processed
+        wait_for_asset_state(eps) do |ready_eps|
+          log_asset_wait_duration!(ready_eps)
```
kookster: So besides the tagged logger block, this is the difference here - `log_asset_wait_duration!` instead of `reset_asset_wait!`, and do it before marking delivered

kookster (approved):
Looks good! Good balance of using status inserts and asset_processing_attempts to track this pesky async processing

Issue #1346, opened 2025-09-04
Delegated Delivery: Single episode in batch slows down progress


svevang:
The processing of episodes in happens in batches. So we peel off the first N episodes and send them through a pipeline of actions. We go through actions the following actions: `upload -> delivery`. Because the asset wait occurs in delivery action, work can pile up in the delivery action and will often pile up on a single episode.

It seems that we should completely publish any finished work before waiting on the unfinished episodes. This would also allow us to decrease the wait times so we can get the episodes that finish more quickly out the door.

PR #1368, opened 2025-10-27
Quiet noisy ops alarms

svevang:
Quieting down the alarms that were amped up in https://github.com/PRX/feeder.prx.org/issues/1346

This restores the prior behavior where the asset timeout errors log level determine if the message lands in #ops-error.

### DIFF
--- a/app/models/apple/api_waiting.rb
+++ b/app/models/apple/api_waiting.rb
@@ -20,11 +20,7 @@ module Apple
       end
 
       def self.work_done?(remaining_records, waited, wait_timeout)
-        if remaining_records.empty?
-          true
-        else
-          false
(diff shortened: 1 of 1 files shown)


app/models/apple/asset_state_timeout_error.rb
```diff
@@ -17,6 +17,22 @@ def raise_publishing_error?
...
+        log_level,
+        message,
+        {
+          podcast_id: podcast_id,
+          attempts: attempts,
+          asset_wait_duration: asset_wait_duration
+        }
+      )
+    end
+
```
kookster: That is nice to encapsulate into its own method +1

Issue #1337, opened 2025-08-27
metrics > feeder: known bugs


radical-ube:
- [ ] browser/history bug and double chart rendering. to reproduce: go to episodes tab, click on a link to an episode, then press the back button. a second chart will render under the first.
- [ ] [session timeout](https://github.com/PRX/feeder.prx.org/pull/1318#discussion_r2294018444) causes metrics failure

brandonhundt:
@radical-ube shall we close this?

PR #1370, opened 2025-10-29
Revision/metrics nav history refactor

radical-ube:
This PR does not add a specific feature; rather, the intent of this PR is to revise some code structure and organization in order to simplify readability, address some known bugs/issues, and to set a cleaner foundation for incoming design work. Much of this PR will be code removal/DRYing, moving, or renaming for clarity of purpose.

Because of re-design work currently happening, this PR is now intended to merge before #1351 as that PR may need to be reworked depending on how the designs shift.

Known issues/bugs addressed:

- [x] double/multiple chart rendering #1337 
- [x] date preset dropdown reflects current selection, or custom #1298 
- [x] form consolidation, for bookmark-able browser history

### DIFF
--- a/app/helpers/metrics_helper.rb
+++ b/app/helpers/metrics_helper.rb
@@ -26,13 +26,13 @@ module MetricsHelper
 
   def podcast_date_preset_options
     Rollups::HourlyDownload::PODCAST_DATE_PRESETS.map do |preset|
-      [label_for_date_preset(preset), preset.to_s]
+      [label_for_date_preset_option(preset), preset.to_s]
     end
   end
 
   def episode_date_preset_options
     Rollups::HourlyDownload::EPISODE_DATE_PRESETS.map do |preset|
-      [label_for_date_preset(preset), preset.to_s]
+      [label_for_date_preset_option(preset), preset.to_s]
     end
   end
 
@@ -42,7 +42,15 @@ module MetricsHelper
     end
   end
 
-  def label_for_date_preset(preset)
+  def label_for_date_preset_field(preset, date_start, date_end)
+    preset_label = label_for_date_preset_option(preset)
+    date_start_label = date_start.strftime("%b %d, %Y")
+    date_end_label = date_end.strftime("%b %d, %Y")
+
+    "#{preset_label}: #{date_start_label} - #{date_end_label}"
+  end
+
+  def label_for_date_preset_option(preset)
     I18n.t(".helpers.label.metrics.date_presets.#{preset}")
   end
 
--- a/app/views/metrics/_filter_dates.html.erb
+++ b/app/views/metrics/_filter_dates.html.erb
@@ -3,8 +3,8 @@
     <%= form.select :interval, interval_options, {selected: interval}, value: interval, data: {action: "change->click#submit"}, class: "form-select form-select-sm" %>
     <%= form.label :interval %>
   </div>
-  <div class="col form-floating btn-group">
-    <%= form.text_field nil, value: label_for_date_preset(date_preset), readonly: true, class: "form-control form-control-sm" %>
+  <div class="col-9 form-floating btn-group">
+    <%= form.text_field nil, value: label_for_date_preset_field(date_preset, date_start, date_end), readonly: true, class: "form-control form-control-sm" %>
     <%= form.label :date_preset, "Date Preset" %>
     <%= form.hidden_field :date_preset, value: date_preset, data: {apex_nav_target: "datePreset", action: "change->click#submit"} %>
     <%= form.hidden_field :date_start, value: nil, data: {apex_nav_target: "startDate"} %>
--- a/app/views/podcast_metrics/_form.html.erb
+++ b/app/views/podcast_metrics/_form.html.erb
@@ -1,10 +1,12 @@
 <% data = {controller: "apex-nav click", apex_nav_main_card_value: main_card, click_debounce_value: 200} %>
 
 <%= form_with(url: podcast_metrics_path(podcast), method: :get, id: "podcast_metrics_tab", data: data) do |form| %>
-  <div class="container col-12 d-flex align-items-center justify-content-between my-4">
-    <h1 class="fw-bold my-0 mb-0 me-4"><%= t(".header") %></h1>
-    <div class="">
-      <div class="d-flex justify-items-end">
+  <div class="container col-12 align-items-center my-4">
+    <div class="row justify-content-between">
+      <div class="col-4">
+        <h1 class="fw-bold my-0 mb-0 me-4"><%= t(".header") %></h1>
+      </div>
+      <div class="col-5">
         <%= render "metrics/filter_dates", form: form, date_preset: date_preset, date_start: date_start, date_end: date_end, interval: interval, episode: nil %>
       </div>
     </div>


app/helpers/metrics_helper.rb
```diff
@@ -12,51 +12,73 @@ def sum_rollups(rollups)
...
+  def episode_main_card_options
+    %i[downloads]
+  end
+
   def interval_options
     Rollups::HourlyDownload::INTERVALS.map { |i| [I18n.t(".helpers.label.metrics.interval.#{i.downcase}"), i] }
   end
 
-  def date_range_options(episode = nil)
+  def date_preset_options(episode = nil)
```
radical-ube: this is used in the `filter_dates` partial, which is reused between the Podcast Metrics and Episode Metrics views, which is why it's formatted with this conditional.

app/helpers/metrics_helper.rb
```diff
@@ -12,51 +12,73 @@ def sum_rollups(rollups)
...
+  def episode_date_preset_options
+    Rollups::HourlyDownload::EPISODE_DATE_PRESETS.map do |preset|
+      [label_for_date_preset(preset), preset.to_s]
+    end
   end
 
-  def metrics_date_presets(options, episode = nil)
-    options.map do |opt|
-      count, interval, type = opt.to_s.split("_")
+  def dates_from_preset(preset, episode = nil)
```
radical-ube: given a preset, this method now just returns a `date_start` and `date_end`. there is a little bit of difference between Podcast/Episode Metrics of the default `start_date` return value, but otherwise these are calculated with the helper methods below.

app/helpers/metrics_helper.rb
```diff
@@ -12,51 +12,73 @@ def sum_rollups(rollups)
...
+    if type == "last"
+      count.to_i.send(interval).ago.utc_date
+    elsif type == "previous"
+      (Date.utc_today - count.to_i.send(interval)).send(:"beginning_of_#{interval.singularize}")
+    elsif type == "todate"
+      Date.utc_today.send(:"beginning_of_#{interval.singularize}")
+    end
+  end
+
+  def date_end_from_preset(preset, date_start)
```
radical-ube: `date_end` sometimes depends on what the `date_start` is.

example of type `drop` is 30 Days Since Drop. only applies to Episode level Metrics, since Podcast level Metrics has its own `Dropdays` tab. `date_start` in this case uses the default for an episode which is `episode.first_publish_utc_date`, and then counts from there.

example of type `previous` is Previous 3 Months. if we are in October, `date_start` would start at the beginning of July. this condition then adds (3 - 1) months to bring us to September, and sets the value to the end of month (Sep 30).

app/javascript/controllers/apex_downloads_controller.js
```diff
@@ -24,4 +24,8 @@ export default class extends Controller {
 
     chart.render()
   }
+
+  disconnect() {
```
radical-ube: i realized that the double/multiple chart rendering bug was because i wasn't cleaning up the DOM of the chart when navigating away. so when the back button is pushed, the OG chart would still be there (but unusable), but a new chart would be rendered. this method properly destroys the chart when the element is disconnected.

app/javascript/controllers/apex_nav_controller.js
```diff
@@ -1,81 +1,55 @@
...
+    agentsCard: String,
   }
 
-  updateEndDate(event) {
-    this.updateAllTargets(this.endDateTargets, event.target.value)
+  connect() {
+    this.tabTargets.forEach((tab) => {
+      if (tab.dataset.card === "main") {
+        if (tab.dataset.tab === this.mainCardValue) {
+          tab.click()
```
radical-ube: this is my current solution for "removing" a default tab client-side. it's clearer in the view partial, but essentially i do not set any tab as active as a default, and this method basically just makes a check to see what the current `main_card` param is and auto-clicks the corresponding tab. this way, a user can bookmark when the view is on a different tab than Downloads and can load directly from there.

it's not quite the right solution because bootstrap is in the way, but it feels close.

app/javascript/controllers/apex_nav_controller.js
```diff
@@ -1,81 +1,55 @@
...
+        }
+      }
+    })
   }
 
-  updateInterval(event) {
```
radical-ube: all of these extra methods were only used because when i had multiple forms and i wanted to target the correct inputs. and now they are no longer needed.

app/views/podcast_metrics/_main_card.html.erb
```diff
@@ -2,35 +2,35 @@
...
-        <li class="nav-item"><button class="nav-link" id="uniques-tab" data-bs-toggle="tab" data-bs-target="#uniques" type="button" role="tab" aria-controls="uniques" aria-selected="false" data-action="apex-nav#changeTab" data-apex-nav-tab-param="uniques"><%= t(".tabs.uniques") %></button></li>
-        <li class="nav-item"><button class="nav-link" id="dropdays-tab" data-bs-toggle="tab" data-bs-target="#dropdays" type="button" role="tab" aria-controls="dropdays" aria-selected="false" data-action="apex-nav#changeTab" data-apex-nav-tab-param="dropdays"><%= t(".tabs.dropdays") %></button></li>
+        <li class="nav-item"><button class="nav-link" id="downloads-tab" data-bs-toggle="tab" data-bs-target="#downloads" type="button" role="tab" aria-controls="downloads" aria-selected="false" data-action="apex-nav#displayCard" data-apex-nav-card-param="main" data-apex-nav-tab-param="downloads" data-apex-nav-target="tab" data-card="main" data-tab="downloads"><%= t(".tabs.downloads") %></button></li>
```
radical-ube: so here i remove `active` and change `aria-selected` to false so that there is technically no default tab to begin with, and the tab to be displayed relies on the `main_card` param.

this works well for the most part, but where i'm stuck (i think) is because of bootstrap. changing tabs with `data-bs-toggle` doesn't seem to advance browser history, so the params don't immediately change when just changing from tab to tab. one of the other form inputs needs to also change because those inputs are the ones submitting a new request each time.

app/controllers/concerns/metrics_queries.rb
```diff
@@ -0,0 +1,48 @@
...
+  include MetricsUtils
+
+  def downloads_date_range_query(model, date_start, date_end, interval)
+    model_column, model_id = set_model_vars(model)
+
+    Rollups::HourlyDownload
+      .where("#{model_column}": model_id, hour: (date_start..date_end))
+      .select(model_column.to_s, "DATE_TRUNC('#{interval}', hour) AS hour", "SUM(count) AS count")
+      .group(model_column.to_s, "DATE_TRUNC('#{interval}', hour) AS hour")
+      .order(Arel.sql("DATE_TRUNC('#{interval}', hour) ASC"))
```
cavis: Just noting, you'll have to add `.final` (the [FINAL](https://clickhouse.com/docs/sql-reference/statements/select/from#final-modifier) modifier) to all these queries, now that updates are coming in throughout the day.

app/javascript/controllers/apex_nav_controller.js
cavis: I think clicking a tab needs to alter the current query params.  Whether just manually changing via javascript, or actually firing a turbo request.

Then you can set the correct active tab in the ruby view.

app/views/metrics/_filter_dates.html.erb
```diff
@@ -1,26 +1,26 @@
...
+    <%= form.select :interval, interval_options, {selected: interval}, value: interval, data: {action: "change->click#submit"}, class: "form-select form-select-sm" %>
+    <%= form.label :interval %>
   </div>
-<% end %>
+  <div class="col-9 form-floating btn-group">
+    <%= form.text_field nil, value: label_for_date_preset_field(date_preset, date_start, date_end), readonly: true, class: "form-control form-control-sm" %>
+    <%= form.label :date_preset, "Date Preset" %>
+    <%= form.hidden_field :date_preset, value: date_preset, data: {apex_nav_target: "datePreset", action: "change->click#submit"} %>
+    <%= form.hidden_field :date_start, value: nil, data: {apex_nav_target: "startDate"} %>
+    <%= form.hidden_field :date_end, value: nil, data: {apex_nav_target: "endDate"} %>
```
radical-ube: not yet using these `date_start` and `date_end` fields, just using them as placeholders for when i will work on Custom date selection.

app/javascript/controllers/apex_nav_controller.js
```diff
@@ -1,81 +1,52 @@
...
-    this.tabTargets.forEach((el) => {
-      if (el.dataset.tab.includes(event.params.tab)) {
-        el.classList.remove("d-none")
-      } else {
-        el.classList.add("d-none")
-      }
-    })
+    const card = this[`${event.params.card}CardTarget`]
+    card.value = event.params.tab
+    card.dispatchEvent(new Event("change"))
```
radical-ube: removed the bootstrap actions so that i could form submit when changing tabs, which should now update the url params.

PR #1377, opened 2025-11-14
Verify episode publishing state

### LINKED ISSUE #1367: Delegated delivery: Probe episode state prior to publishing actions
There are enough cases where the remote publishing state can become decoupled with the local state that it seems worthwhile to probe the remote episode state with each run of the publishing routine. In this latest case, there were several 500s on the remote publishing endpoint, and when it came back up the episode had moved to `PUBLISHED` state but was not locally recorded.

- Still TBD, is if we probe all episodes or just the current working set? Or just the episodes we intend to publish.
- Check the episode api endpoint for api throttling

svevang:
Closes #1367 

This handles the case where the the publishing state transition throws errors. When this happens the remote and local publishing state become decoupled. If we've experienced one of these 500s and are trying to publish again, this will first verify the remote publishing state and then publish. If the state has drifted, then the whole publishing routine is retried and the episodes now recognized as published will drop out of the publishing state machine.

If publishing state desynchronization continues to crop up, we should be able to extend this to probe the entire remote episode set. Probing all the episodes is a somewhat expensive operation, taking upwards of 2 minutes and ~1k api requests for some podcasts. For now, it's clear that these 500s are producing a very limited set of drifted episodes, so this PR addresses this case first.

### DIFF
--- a/test/models/apple/publisher_test.rb
+++ b/test/models/apple/publisher_test.rb
@@ -872,9 +872,11 @@ describe Apple::Publisher do
       apple_publisher.stub(:increment_asset_wait!, increment_mock) do
         apple_publisher.stub(:wait_for_upload_processing, wait_upload_mock) do
           apple_publisher.stub(:wait_for_asset_state, wait_for_asset_state_stub) do
-            Apple::Episode.stub(:probe_asset_state, probe_mock) do
-              apple_publisher.stub(:mark_as_delivered!, mark_delivered_mock) do
-                apple_publisher.process_delivery!([episode])
+            apple_publisher.stub(:verify_publishing_state!, ->(*) {}) do
+              Apple::Episode.stub(:probe_asset_state, probe_mock) do
+                apple_publisher.stub(:mark_as_delivered!, mark_delivered_mock) do
+                  apple_publisher.process_delivery!([episode])
+                end
               end
             end
           end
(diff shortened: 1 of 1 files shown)


app/models/apple/publisher.rb
```diff
@@ -398,11 +398,47 @@ def mark_as_uploaded!(eps)
...
+      # Capture initial state before polling to detect drift
+      initial_states = eps.to_h { |ep| [ep.feeder_id, ep.publishing_state] }
+
+      poll_episodes!(eps)
+
+      drifted_count = eps.count do |ep|
+        initial_state = initial_states[ep.feeder_id]
+        (ep.publishing_state != initial_state).tap do |drifted|
+          if drifted
+            Rails.logger.warn("Episode publishing state found to be out of sync", {
```
kookster: great to have visibility into this drift

svevang:
@kookster  Re: https://github.com/PRX/feeder.prx.org/pull/1377#pullrequestreview-3485149459
Increasingly it looks like there are just two main phases to the Apple::Publisher publishing process. So this adds a synchronization touchpoint in the `Apple::Publisher#process_delivery` in addition to the existing one in https://github.com/PRX/feeder.prx.org/blob/main/app/models/apple/publisher.rb#L117 Both process_delivery and upload_media phases are reentrant, and now both will probe the remote state (at some point). 

But the focus is on the publishing state (DRAFT, ARCHIVED etc). We don't currently handle sync-drift mutations elsewhere e.g. in someone unlinking the file in the podcast container right before we publish. So this is all to say, that I think you're right about the API error status on the episode delivery status. Aside: I'm wondering if we can track this in the upload and delivery phases themselves?

Also, there is an "unknown unknowns" here. E.g. we have "complete" episodes that have a settled delivery status. And, unbeknownst to us, their remote state has mutated in a way that necessitates a redelivery (if we knew about that mutation).

PR #1379, opened 2025-11-19
Remove podcast and episode metrics specific tabs in UI

radical-ube:
This PR removes the podcast and episode `Metrics` tabs from the UI. "Filter" partials are also being removed as they won't be necessary moving forward. I'm mainly focusing on removing views/partials in this PR as I am fairly confident that I won't need these in particular. I am not yet removing any controller or JS code quite yet, as I may still be able to repurpose them for the metrics that will still be used in Feeder.


cavis (approved):
May want to remove translations as you remove stuff?  But otherwise 👍 looks good to me!

Issue #1215, opened 2025-03-20
Use `serveDrafts` for Apple feeds too?


kookster:
This new flag on a feed will allow DTR to serve audio if there is audio available on feeder, regardless of publish status.
This means we could sync audio when it is changed/uploaded instead of waiting for publishing, or using the offset for an earlier publish window.

PR #1381, opened 2025-11-19
Add episode publish offset for apple feeds

### LINKED ISSUE #1380: Add publish offset to the apple feed
For this apple delegated delivery scenario:

"Early access via RSS linking: subscribers will get episodes before they are made available in the RSS feed to everyone else"

Add this to the apple feed so you can publish episodes to subscribers early

kookster:
fixes #1380 and #1215 

I added options only for publishing early to subscribers, and made it multiples of days or weeks:

image

Also enabled `serve_drafts` for Apple feeds so DTR will not look at publish date for serving audio when it is for the Apple feed.

### DIFF
--- a/app/helpers/feeds_helper.rb
+++ b/app/helpers/feeds_helper.rb
@@ -3,6 +3,10 @@ module FeedsHelper
     I18n.t("feeds.helper.episode_offset_options").invert.to_a
   end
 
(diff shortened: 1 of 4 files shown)


cavis (approved):
👍 works for me!

PR #1384, opened 2025-11-20
Handle feed inclusion delegated delivery UI

### LINKED ISSUE #1354: Delegated delivery UI missing episode state
There are two pieces of missing state in the delegated delivery UI:

1) Feed inclusion. Currently the apple delivery status is displayed based on the apple publishing config being enabled. But it should also depend on feed inclusion. 

2) Archived State. Episodes in an archived state should have no apple status.

svevang:
closes #1354 

This PR improves how episode integration statuses are displayed in the UI, particularly for feeds using delegated delivery where not all episodes are published to external integrations. It also

- Updates the Apple publishing flow to ensure that the publishing routine is nestled under the topmost delivery status. So if an episode is marked as `delivered?` then it is marked as published in Apple's system.
- Ensures that feed inclusion is take into account when displaying status via `integration_feed_episode?`. This prevents showing misleading "processing" or "incomplete" statuses for episodes that will never be published to an integration.
- Delegated delivery uses the same badges and view code as the other integrations and media (via media_statuses translations)

### DIFF
--- a/app/models/integrations/episode_integrations.rb
+++ b/app/models/integrations/episode_integrations.rb
@@ -27,8 +27,6 @@ module Integrations::EpisodeIntegrations
     integration_episode_method = "#{integration}_episode"
     if respond_to?(integration_episode_method)
       send(integration_episode_method)
-    else
-      nil
     end
   end
 


app/views/episodes/_form_status.html.erb
```diff
@@ -35,15 +35,19 @@
       <%= render "form_feeds", form: form, episode: episode %>
     </div>
 
-    <% if episode.persisted? && episode.publish_to_apple? %>
+    <% if episode.persisted? && episode.publish_to_integration?(:apple) %>
```
kookster: seems like the status display code for megaphone and apple are identical now, so maybe we could just loop through each of those integrations, instead of duplicating the block of code?

test/helpers/episodes_helper_test.rb
```diff
@@ -0,0 +1,85 @@
+require "test_helper"
+
+class TestHelper
+  include EpisodesHelper
+end
+
+describe EpisodesHelper do
```
kookster: great to have tests for all this!

app/models/integrations/episode_integrations.rb
```diff
@@ -28,6 +28,24 @@ def publish_to_integration?(integration)
...
+  end
+
+  def integration_feed(integration)
+    podcast.feeds.find { |f| f.integration_type == integration }
+  end
+
+  def integration_error_state?(integration)
+    case integration
+    when :apple
+      apple_episode&.audio_asset_state_error? || apple_episode&.delivery_file_errors?
```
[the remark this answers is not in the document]
svevang: > could we have this call a generic method that could be implemented by any of the integrations

Yep. I'm in a functional mindset where I wanted to see all possible branches, but def this is not good OOP style. Will extract this into helpers!

> we should also introduce a more generic error status on the episode delivery status record

Yep, also thinking about state, and that discussion we had earlier about stale / drifted state on the sync. Part of me wishes for just a pure functional reduction on the episode delivery state, but of course the publish-job retry loop is stateless. So it seems inevitable that we end up with more error semantics (status) on that episode delivery status record.

app/views/episodes/_form_status.html.erb
svevang: 👍

app/models/integrations/episode_integrations.rb
```diff
@@ -38,9 +45,9 @@ def integration_feed(integration)
   end
 
   def integration_error_state?(integration)
-    case integration
-    when :apple
-      apple_episode&.audio_asset_state_error? || apple_episode&.delivery_file_errors?
+    integration_episode_method = "#{integration}_episode"
+    if respond_to?(integration_episode_method)
+      send(integration_episode_method)&.error_state? || false
```
kookster: very clean, thank you!

app/models/episode.rb
```diff
@@ -12,7 +12,7 @@ class Episode < ApplicationRecord
   include PublishingStatus
   include TextSanitizer
   include EmbedPlayerHelper
-  include AppleDelivery
+  include AppleIntegration
```
kookster: that is a better name I think

app/views/episodes/_form_status.html.erb
```diff
@@ -35,37 +35,22 @@
...
+            <span class="badge rounded-pill prx-badge-<%= integration_status %>">
+              <%= t("helpers.label.episode.media_statuses.#{integration_status}") %>
+            </span>
+            <br>
+            <strong><%= integration.to_s.titleize %> <%= t(".last_updated") %></strong>
+            <br>
+            <%= local_time_ago(episode_integration_updated_at(integration, episode)) %>
+          </p>
+        </div>
+      <% end %>
```
kookster: looks great

PR #1386, opened 2025-11-24
Feat/episode sparklines

radical-ube:
This PR adds sparklines and trend numbers to the episode cards on the Podcast dashboard. With the updated designs, I am slowly refactoring the `podcast_metrics_controller` and various apex stimulus controllers because some of the code can still be repurposed. The updated podcast dashboard has several charts in it, so I plan on using the `podcast_metrics_controller` to handle as much of the metrics querying as possible so as not to bloat the `podcast_controller`.

With the case of the sparklines, each episode card has its own turbo frame, which sources from `podcast_metrics#episode_sparkline` for each of their individual querying needs. `alltime_downloads` pulls data for the sparkline itself, while the `episode_trend` include the dropdate queries for the episode and the previously published episode.

There is a little bit of a question around what is our opinion of `trending_flat`, and the threshhold we would consider whether a change is significant enough to signal trend in a particular direction.

@cavis for code review
@brandonhundt for design review

### DIFF
--- a/app/controllers/podcast_metrics_controller.rb
+++ b/app/controllers/podcast_metrics_controller.rb
@@ -2,7 +2,7 @@ class PodcastMetricsController < ApplicationController
   include MetricsUtils
 
   before_action :set_podcast
-  before_action :check_clickhouse, except: %i[show]
+  # before_action :check_clickhouse, except: %i[show]
 
   def show
   end
@@ -12,9 +12,10 @@ class PodcastMetricsController < ApplicationController
     @prev_episode = Episode.find_by(guid: metrics_params[:prev_episode_id])
 
     @episode_trend = calculate_episode_trend(@episode, @prev_episode)
+
     @sparkline_downloads =
       Rollups::HourlyDownload
-        .where(episode_id: @episode[:guid], hour: (@episode.first_rss_published_at..))
+        .where(episode_id: @episode[:guid], hour: (publish_hour(@episode)..publish_hour(@episode) + 6.months))
         .final
         .select(:episode_id, "DATE_TRUNC('DAY', hour) AS hour", "SUM(count) AS count")
         .group(:episode_id, "DATE_TRUNC('DAY', hour) AS hour")
@@ -219,7 +220,7 @@ class PodcastMetricsController < ApplicationController
   end
 
   def calculate_episode_trend(episode, prev_episode)
-    return nil unless episode.in_default_feed? && episode.first_rss_published_at.present? && prev_episode.present?
+    return nil unless episode.first_rss_published_at.present? && prev_episode.present?
     return nil if (episode.first_rss_published_at + 1.day) > Time.now
 
     ep_dropday_sum = episode_dropday_query(episode)
@@ -227,11 +228,11 @@ class PodcastMetricsController < ApplicationController
 
     return nil if ep_dropday_sum <= 0 || previous_ep_dropday_sum <= 0
 
-    ((ep_dropday_sum.to_f / previous_ep_dropday_sum.to_f) - 1).round(2)
+    ((ep_dropday_sum.to_f / previous_ep_dropday_sum.to_f) - 1).round(3)
   end
 
   def episode_dropday_query(ep)
-    lowerbound = ep.first_rss_published_at.beginning_of_hour
+    lowerbound = publish_hour(ep)
     upperbound = lowerbound + 24.hours
 
     Rollups::HourlyDownload
@@ -239,5 +240,13 @@ class PodcastMetricsController < ApplicationController
       .final
       .load_async
       .sum(:count)
+  end
+
+  def publish_hour(episode)
+    if episode.first_rss_published_at.present?
+      episode.first_rss_published_at.beginning_of_hour
+    else
+      episode.published_at.beginning_of_hour
+    end
   end
 end
--- a/app/helpers/metrics_helper.rb
+++ b/app/helpers/metrics_helper.rb
@@ -17,13 +17,13 @@ module MetricsHelper
 
     if trend > 0
       {
-        percent: "+#{(trend * 100).round(2)}%",
+        percent: "+#{trend * 100}%",
         color: modified_trend_color(trend, "text-success"),
         direction: modified_trend_direction(trend, "up")
       }
     elsif trend < 0
       {
-        percent: "-#{(trend * -100).round(2)}%",
+        percent: "-#{trend * -100}%",
         color: modified_trend_color(trend, "text-danger"),
         direction: modified_trend_direction(trend, "down")
       }


app/helpers/metrics_helper.rb
```diff
@@ -12,6 +12,40 @@ def sum_rollups(rollups)
     rollups.sum(&:count)
   end
 
+  def parse_trend(trend)
+    return if trend.nil?
```
radical-ube: for the view, i use this helper to determine the visuals. for `trending_flat` i've arbitrarily chosen 5%, but could easily change it to 2 or 3% or whichever number we decide on makes sense.

app/views/metrics/_episode_sparkline.html.erb
```diff
@@ -0,0 +1,13 @@
+<%= turbo_frame_tag "episode_sparkline" do %>
+  <% trend = parse_trend(episode_trend) %>
+  <div class="d-flex align-items-center <%= trend[:color] %>">
+    <span><%= trend[:percent] if trend.present? %></span>
+    <span class="material-icons ms-1" aria-hidden="true"><%= trend[:direction] if trend.present? %></span>
```
radical-ube: trend shouldn't render unless it has passed all the conditionals

app/views/podcasts/show.html.erb
```diff
@@ -41,17 +41,27 @@
...
         <div class="card-body episode-card mb-0 h-auto border-start-0 p-0 d-flex flex-column">
-          <% if @recently_published.present? %>
-            <% @recently_published.each do |ep| %>
-              <div class="episode-card-info inside-card flex-fill px-3 py-2">
-                <p class="episode-card-date"><%= local_date(ep.published_at, format: :short) %></p>
-                <div class="episode-card-inner inside-card">
-                  <h2 class="h5 episode-card-title m-0"><%= link_to ep.title, episode_path(ep) %></h2>
-                </div>
-              </div>
+          <% if Rails.env.development? %>
```
radical-ube: keeping the sparklines in development until we release metrics > feeder.

app/helpers/metrics_helper.rb
brandonhundt: @radical-ube 5% seems a good place to start.

radical-ube: i wonder about this at various tiers though. like a podcast going from 2000 to 2100 (5% increase) maybe feels kinda small and not too significant. but 20000 to 21000 (still 5%) might be a bit more notable. so are there certain threshholds, maybe based on the size of the podcast or some other characteristics, where we do want to highlight, "hey, your latest episode is trending (up or down) much larger than usual" that might not be solely based on the percentage?

or should it be equal across all podcasts?

app/views/metrics/_episode_sparkline.html.erb
```diff
@@ -0,0 +1,13 @@
...
+  </div>
+  <div class="position-absolute top-50 start-50 translate-middle w-100 h-100 z-0"
+    data-controller="apex-downloads"
+    data-apex-downloads-id-value="<%= episode.guid %>"
+    data-apex-downloads-downloads-value="<%= downloads.to_json %>">
+    <div data-apex-downloads-target="chart"></div>
```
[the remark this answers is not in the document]
brandonhundt: +1 to making the the border line transparent.

Can we change the colors in the gradient?

app/controllers/podcast_metrics_controller.rb
```diff
@@ -195,46 +214,31 @@ def set_podcast
...
+    prev_ep_sum = previous_ep_dropday.sum(&:count)
 
-  def uniques_params
-    params
-      .permit(:uniques_selection)
-      .with_defaults(
-        uniques_selection: "last_7_rolling"
-      )
-      .merge(metrics_params)
+    ((ep_sum.to_f / prev_ep_sum.to_f) - 1).round(2)
```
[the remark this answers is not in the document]
radical-ube: Screenshot 2025-11-25 at 3 43 42 PM

i think i'm still missing some guards to prevent weird edge cases like this. maybe something around the case where one of the dropday queries doesn't return anything?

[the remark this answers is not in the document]
brandonhundt: Is there a chance we can load this against a prod db? This way we might be able to see the edge cases a little easier.

app/controllers/podcast_metrics_controller.rb
```diff
@@ -3,13 +3,32 @@ class PodcastMetricsController < ApplicationController
...
-  before_action :set_dropday_range, only: %i[show dropdays]
 
   def show
   end
 
+  def episode_sparkline
+    @episode = Episode.find_by(guid: metrics_params[:episode_id])
+    @prev_episode = Episode.find_by(guid: metrics_params[:prev_episode_id])
+
+    @alltime_downloads =
```
cavis: Why are you querying for all-time?  I can't quite figure out how this impacts the sparklines.

Or are those little arrow icons not static?  I sort of assumed they were just the static "up or down".

app/controllers/podcasts_controller.rb
```diff
@@ -17,7 +17,12 @@ def index
   def show
     authorize @podcast
 
-    @recently_published = @podcast.episodes.published.dropdate_desc.limit(3)
+    @recently_published_episodes = @podcast.episodes.published.dropdate_desc.limit(4)
+    @trend_episodes = @podcast.default_feed.episodes.published.dropdate_desc.where.not(first_rss_published_at: nil).offset(1).limit(4)
```
radical-ube: here i am pulling the episodes to be compared against in the trend calculation. offset by 1 because i don't want the most recent one.

app/controllers/podcast_metrics_controller.rb
```diff
@@ -3,13 +3,31 @@ class PodcastMetricsController < ApplicationController
...
   end
 
+  def episode_sparkline
+    @episode = Episode.find_by(guid: metrics_params[:episode_id])
+    @prev_episode = Episode.find_by(guid: metrics_params[:prev_episode_id])
+
+    @episode_trend = calculate_episode_trend(@episode, @prev_episode)
+    @sparkline_downloads =
+      Rollups::HourlyDownload
+        .where(episode_id: @episode[:guid], hour: (@episode.first_rss_published_at..))
```
cavis: I think you need to set the lowerbound to the start of the hour... otherwise you'll miss out on the first hour.

Imagine we wrote the RSS at 5 minutes after the hour - there could be a substantial spike for those next 55 minutes.

cavis: Also, can we cap this at 6 months of data or something?  I happened to pull up Transistor (last episode published in 2017)... and this was a lot of data!

app/controllers/podcast_metrics_controller.rb
```diff
@@ -195,46 +213,31 @@ def set_podcast
...
+    return nil if ep_dropday_sum <= 0 || previous_ep_dropday_sum <= 0
 
-  def uniques_params
-    params
-      .permit(:uniques_selection)
-      .with_defaults(
-        uniques_selection: "last_7_rolling"
-      )
-      .merge(metrics_params)
+    ((ep_dropday_sum.to_f / previous_ep_dropday_sum.to_f) - 1).round(2)
```
cavis: Shouldn't this be `round(3)` - since you're displaying 3 digits in the UI?  Seeing everything with a `.0` on it, which seems unlikely in every case.

app/controllers/podcast_metrics_controller.rb
```diff
@@ -3,13 +3,31 @@ class PodcastMetricsController < ApplicationController
...
+
+    @episode_trend = calculate_episode_trend(@episode, @prev_episode)
+    @sparkline_downloads =
+      Rollups::HourlyDownload
+        .where(episode_id: @episode[:guid], hour: (@episode.first_rss_published_at..))
+        .final
+        .select(:episode_id, "DATE_TRUNC('DAY', hour) AS hour", "SUM(count) AS count")
+        .group(:episode_id, "DATE_TRUNC('DAY', hour) AS hour")
+        .order(Arel.sql("DATE_TRUNC('DAY', hour) ASC"))
+        .load_async
```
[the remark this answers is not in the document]
radical-ube: captured in #1400

app/controllers/podcast_metrics_controller.rb
```diff
@@ -195,46 +214,39 @@ def set_podcast
...
   end
 
-  def dropdays_params
-    params
-      .permit(:dropday_range, :interval)
-      .with_defaults(
-        dropday_range: 7
-      )
-      .merge(metrics_params)
+  def publish_hour(episode)
```
radical-ube: is this accurate? trying to address the `.beginning_of_hour` comment but also catching episodes that don't have `.first_rss_published_at`.

cavis: I would probably just return a 0 / empty array, instead of querying clickhouse at all, if that's nil.

The implication is, this episode was never written to the default feed.  So you're not going to find any data in Clickhouse.

PR #1395, opened 2025-12-11
Feat/podcast scorecards dashboard

radical-ube:
follows after #1386 

This PR adds the scorecard dashboard to the Podcast overview page. There is some repurposing and moving of some already created views/queries/controllers, so some of this PR may appear to have larger changes than there actually are.

Some things to note, as this was in development:
- [ ] initial querying may produce errors from Clickhouse because currently only 4 queries can run simultaneously. some caching seems to help resolve some collisions after initial loading.
- [ ] there is some weird gradient behavior on the line charts that does not happen in the apex demo chart that remains to be explained. for now, lines will be in the solid blue until the reason is found.

@cavis for code review
@brandonhundt for design review

### DIFF
--- a/app/models/episode.rb
+++ b/app/models/episode.rb
@@ -312,19 +312,14 @@ class Episode < ApplicationRecord
   end
 
   def previous_trend_episode
-    prev_ep_guid = Rails.cache.fetch("#{cache_key_with_version}/previous_trend_episode", expires_in: 30.days) do
-      if feeds.include?(podcast.default_feed) && first_rss_published_at.present?
-        podcast.default_feed.episodes.published.dropdate_desc.where("episodes.first_rss_published_at IS NOT NULL AND episodes.first_rss_published_at < ?", first_rss_published_at).first&.guid
-      end
-    end
-
-    if prev_ep_guid.present?
-      Episode.find_by(guid: prev_ep_guid)
+    if feeds.include?(podcast.default_feed) && first_rss_published_at.present?
+      podcast.default_feed.episodes.published.dropdate_desc.where("episodes.first_rss_published_at IS NOT NULL AND episodes.first_rss_published_at < ?", first_rss_published_at).first
     end
   end
 
   def dropday_sum
     return nil unless first_rss_published_at.present?
+    return nil if (first_rss_published_at + 1.day) > Time.now
 
     Rails.cache.fetch("#{cache_key_with_version}/dropday_sum", expires_in: 30.days) do
       lowerbound = first_rss_published_at.beginning_of_hour


app/javascript/controllers/apex_monthly_controller.js
```diff
@@ -0,0 +1,24 @@
+import { Controller } from "@hotwired/stimulus"
+import { buildDateTimeChart, buildDownloadsSeries, BAR_TYPE, destroyChart } from "util/apex"
```
radical-ube: new controller for the Monthly Downloads chart.

app/controllers/podcast_metrics_controller.rb
```diff
@@ -8,81 +9,47 @@ def show
...
-        .where(podcast_id: @podcast.id, hour: (@date_start..@date_end))
-        .select("DATE_TRUNC('#{@interval}', hour) AS hour", "SUM(count) AS count")
-        .group("DATE_TRUNC('#{@interval}', hour) AS hour")
-        .order(Arel.sql("DATE_TRUNC('#{@interval}', hour) ASC"))
-        .load_async
+  def episode_trend
+    @episode = Episode.find_by(guid: params[:episode_id])
+    render partial: "metrics/episode_trend", locals: {
+      episode: @episode,
+      episode_trend: @episode.episode_trend
```
radical-ube: same here. basically moving the `calculate_episode_trend` method below into the `Episode` model for caching.

app/models/episode.rb
```diff
@@ -310,4 +310,72 @@ def head_request(uri_str = enclosure_url(podcast.default_feed), redirects = 0)
...
+        podcast.default_feed.episodes.published.dropdate_desc.where("episodes.first_rss_published_at IS NOT NULL AND episodes.first_rss_published_at < ?", first_rss_published_at).first&.guid
+      end
+    end
+
+    if prev_ep_guid.present?
+      Episode.find_by(guid: prev_ep_guid)
+    end
+  end
+
+  def dropday_sum
```
radical-ube: used for calculating trend. maybe cache shouldn't need to expire? there is a guard in the `episode_trend` method below so that i make the assumption that this is query is only run if it has been at least 1 day since drop.

app/models/episode.rb
```diff
@@ -310,4 +310,66 @@ def head_request(uri_str = enclosure_url(podcast.default_feed), redirects = 0)
...
+
+    Rails.cache.fetch("#{cache_key_with_version}/sparkline_downloads", expires_in: 28.days) do
+      Rollups::HourlyDownload
+        .where(episode_id: guid, hour: publish_hour..(publish_hour + 1.month))
+        .final
+        .select(:episode_id, "DATE_TRUNC('DAY', hour) AS hour", "SUM(count) AS count")
+        .group(:episode_id, "DATE_TRUNC('DAY', hour) AS hour")
+        .order(Arel.sql("DATE_TRUNC('DAY', hour) ASC"))
+        .load_async
+        .pluck(Arel.sql("DATE_TRUNC('DAY', hour) AS hour"), Arel.sql("SUM(count) AS count"))
```
radical-ube: documentation recommends not to cache the instance records themselves, so here i'm just plucking the `hour` and `count` as a tuple(?) since that's all that i need from this query.

brandonhundt:
Noting this passes my design review. We can come back later, if we need to, when all the dashboard pieces are in place. Looking good! 

(I'm going to take myself off the PR as a reviewer, but let me know if you need anything else on this.)

app/models/episode.rb
```diff
@@ -310,4 +310,66 @@ def head_request(uri_str = enclosure_url(podcast.default_feed), redirects = 0)
...
+    prev_ep_guid = Rails.cache.fetch("#{cache_key_with_version}/previous_trend_episode", expires_in: 30.days) do
+      if feeds.include?(podcast.default_feed) && first_rss_published_at.present?
+        podcast.default_feed.episodes.published.dropdate_desc.where("episodes.first_rss_published_at IS NOT NULL AND episodes.first_rss_published_at < ?", first_rss_published_at).first&.guid
+      end
+    end
+
+    if prev_ep_guid.present?
+      Episode.find_by(guid: prev_ep_guid)
+    end
+  end
```
cavis: Can these metrics-y methods move into a concern instead?  `EpisodeMetrics` or something.  Just to keep this model as slim as we can.

app/controllers/concerns/metrics_queries.rb
```diff
@@ -0,0 +1,56 @@
+require "active_support/concern"
+
+module MetricsQueries
```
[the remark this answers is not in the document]
cavis: These would also be a good candidate for a model concern?  I could imagine wanting an easy way to get these important numbers from the rails console.  `podcast.downloads_total` or something.

Also a convenient place to handle caching, and maybe someday testing (though that's problematic at the moment, since it would need a clickhouse db to work).

(Not asking for a change just yet - that would mess up all the PRs after this!  But it would be a nice refactor at some point).

PR #1398, opened 2025-12-15
Feat/bar chart cards on bottom half of podcast metrics

radical-ube:
follows after #1395 

This PR adds the bar chart cards which live on the bottom half of the podcast dashboard: feeds (conditional), seasons (conditional), countries, and apps.

@cavis code review
@brandonhundt design review

### DIFF
--- a/app/models/podcast.rb
+++ b/app/models/podcast.rb
@@ -292,8 +292,7 @@ class Podcast < ApplicationRecord
         .group(:feed_slug)
         .order(Arel.sql("SUM(count) AS count DESC"))
         .final
-        .load_async
-        .pluck(:feed_slug, Arel.sql("SUM(count) AS count"))
+        .sum(:count)
     end
   end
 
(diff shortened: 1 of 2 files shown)


app/models/podcast.rb
```diff
@@ -280,4 +280,131 @@ def respond_to_missing?(method, *args)
       super
     end
   end
+
+  def feeds_downloads_query
```
radical-ube: these queries get refactored into a concern in the 3rd PR, fyi

app/models/rollups/daily_agent.rb
```diff
@@ -1,154 +1,39 @@
...
-    136 => "Snipd",
-    137 => "Telmate",
-    138 => "castget",
-    139 => "Newsboat",
-    140 => "Anghami",
-    141 => "VLC",
-    142 => "PRX Play",
-    143 => "mowPod"
-  }.freeze
+  LABELS =
```
radical-ube: pulled the labeling code that exists in Augury.

cavis: So I think, like Augury, you'll want to change the `Dockerfile` to download/update this every time the app is built.  And then `.gitignore` the actual yml file instead of committing it to the repo.

app/models/podcast.rb
```diff
@@ -280,4 +280,131 @@ def respond_to_missing?(method, *args)
...
+
+    Rails.cache.fetch("#{cache_key_with_version}/top_countries_downloads", expires_in: 1.day) do
+      Rollups::DailyGeo
+        .where(podcast_id: id, day: date_start..date_end)
+        .select(:country_code, "SUM(count) AS count")
+        .group(:country_code)
+        .order(Arel.sql("SUM(count) AS count DESC"))
+        .final
+        .limit(10)
+        .load_async
```
cavis: Hmm... I'm not sure you're actually loading this async, since it's inside the cache block.  I _think_ the `.pluck` is just running the query immediately.

Which means, you can simplify these queries somewhat.  Doing a `.sum(:count)` instead of the select/pluck.

Issue #1375, opened 2025-11-04
Stream recording architecture


cavis:
Trying to capture the general architecture we talked about for stream recording:

1. Feeder database is the source of truth for what streams we're recording when
2. Feeder writes "stream recording config" to some bulletproof _global_ location when changed
    - Could be as simple as a JSON file in `s3://prx-feed` or some other location
    - Or dynamodb global table
    - Indicates what streams to record, what redundancy level, what start/stop timestamps, what regions etc etc
4. Some "stream recording orchestrator" lambda lives in `feeder.yml` CFN stack
    - Triggered at some interval (10 minutes before every hour or something?) by CW cron event
    - Wakes up, reads the config from (2), launches Oxbow
    - Lives in both regions (us-east-1, us-west-2)
    - Has its own "temp" S3 bucket to store stream recordings
5. After oxbow completes, sends a callback _to Feeder SQS_ rails worker
    - Indicate what stream, what hour this is, where the file is, etc
    - Uses the "temp" bucket in (4)
6. After getting the callback, Feeder (rails) decides how to process the file
    - Probably messaging Porter to trim the file to just the hour
    - Probably generating a waveform
    - NOTE: this does assume we'll lose the temp full-recording file, and won't be able to re-trim.  Is that ok?

PR #1402, opened 2025-12-18
Stream recording callbacks

cavis:
More of #1375.

Schema/Tasks to process Oxbow callbacks for stream recordings, setting up db records and copying the audio into Feeder's S3 bucket.

## Schema

General schema I'm aiming towards:

image

## Process Flow

1. User creates a `StreamRecording` for their podcast.  This communicates which url/hours to record over to the [stream-recording lambda](https://github.com/PRX/stream-recorder?tab=readme-ov-file#stream-configuration).
2. A few minutes before the hour, the lambda fires up Oxbow(s) to record to an S3 file.
3. Feeder gets an SQS `JobReceived` and creates _a brand new Tasks::RecordStreamTask_ to capture these callbacks.  All the data needed is in the Oxbow Job ID:
    ```
    284/3/2026-01-09T18:00Z/2026-01-09T19:00Z/afe3b72f-77fe-4be9-be37-34c528cc9783.mp3
    ```
4. Feeder finds or creates a `StreamResource` for the timeframe we're attempting to capture - `2026-01-09T18:00Z` to `2026-01-09T19:00Z`, for `stream_recording_id = 3`.
5. Oxbow finishes recording, sends a final `JobResult` that includes the S3 output (in the stream-recorder's bucket) and additional metadata:
    ```
    "Duration" => 60000, "Size" => 960788, "StartEpoch" => 1767982842
    ```
6. Feeder decides if this file captures _the most_ of the timeframe we're after.  If so, we update the `StreamResource` `actual_start_at` and `actual_end_at`, and kick off a `Tasks::CopyMediaTask` to inspect this file and copy it into the Feeder S3 bucket.

### DIFF
--- a/app/models/stream_resource.rb
+++ b/app/models/stream_resource.rb
@@ -60,10 +60,12 @@ class StreamResource < ApplicationRecord
     (status_complete? || status_invalid?) ? url : original_url
   end
 
+  # NOTE: for compatibility with copy_media_task.rb
   def medium
     "audio"
   end
 
+  # NOTE: for compatibility with copy_media_task.rb
   def medium=(val)
     medium
   end


app/models/stream_resource.rb
```diff
@@ -0,0 +1,137 @@
...
+  end
+
+  # NOTE: called twice, because podcast won't be there on initialize
+  def set_defaults
+    set_default(:status, "created")
+    set_default(:guid, SecureRandom.uuid)
+    set_default(:url, published_url)
+  end
+
+  def copy_media(force = false)
```
cavis: A whole bunch of this mirrors `MediaResource`.  Because (A) it's similar, and (B) I wanted to reuse the `CopyMediaTask`.

(I had a [slack thread](https://prx.slack.com/archives/G0FKLK2Q5/p1766073162493249) about this, and landed on making brand new tables for these resources, and someday the "clips").

app/models/stream_resource.rb
```diff
@@ -0,0 +1,137 @@
...
+      end
+    return unless podcast_id > 0 && recording_id > 0 && start_at && end_at
+
+    rec = StreamRecording.find_by_id(recording_id)
+    return unless rec
+
+    res = rec.stream_resources.find_by(start_at: start_at, end_at: end_at)
+    return res if res
+
+    rec.stream_resources.build(start_at: start_at, end_at: end_at)
```
cavis: It's possible we're recording a single hour multiple times.  So in here we find or create the `StreamResource` for that hour.

app/models/tasks/record_stream_task.rb
```diff
@@ -0,0 +1,119 @@
...
+  end
+
+  def start_at
+    job_id_parts[2]&.to_time
+  rescue
+    nil
+  end
+
+  def end_at
+    job_id_parts[3]&.to_time
```
svevang: Some shared job id parsing here and in steam resource. Is it worth extracting a little helper interface for these?

app/models/tasks/record_stream_task.rb
```diff
@@ -0,0 +1,119 @@
...
+  end
+
+  def stream_recording_id
+    job_id_parts[1].to_i if job_id_parts[1].to_i > 0
+  end
+
+  def start_at
+    job_id_parts[2]&.to_time
+  rescue
+    nil
```
svevang: The timstamp is malformed here. Can we log out the malformed job id?

svevang (commented):
Questions!

app/models/stream_resource.rb
```diff
@@ -0,0 +1,137 @@
...
+  validates :original_url, presence: true, if: :done_recording?
+
+  after_initialize :set_defaults
+  before_validation :set_defaults
+
+  acts_as_paranoid
+
+  # find/build for a specific oxbow job_id
+  # <podcast_id>/<stream_recording_id>/<start_time>/<end_time>/<guid>.mp3
+  def self.decode(str)
```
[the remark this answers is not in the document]
cavis: Oh, whoops!  I moved all that logic into `record_stream_task.rb` but never removed this version.

app/models/tasks/record_stream_task.rb
cavis: Removed the unused `.decode` method

svevang (approved):
👍 +1

PR #1405, opened 2026-01-07
Delivery file processing timeouts

### LINKED ISSUE #1390: Delegated delivery, overloaded timeout error handling
A delegated delivery failed to complete when the call to `Apple::PodcastDeliveryFile.wait_for_processing` became wedged.  There were no error level log lines due to 1) the way that API timeouts are handled in job error handling routine (they're assumed to come from episode asset timeouts) and 2) that this was a `sync_blocks_rss == false)` config.

So it looks like we need to:
- ensure that errors are thrown for non-asset timeouts (esp. when those errors are multiplexed with the sync_blocks_rss)

Additionally, we have logic to handle 'extreme' timeouts in remote services, essentially triggering a new upload attempt when the current attempt takes 'a long time'.  In order to automatically reattempt the upload and delivery stages for this wedged file processing means we need to extend this timeout handling in Apple::Publisher -- not just the asset processing/polling.

svevang:
closes #1390 
This PR does the following:
  - Fix timeout errors during delivery file processing not being logged when `sync_blocks_rss` is disabled
  - Add stuck episode detection to `wait_for_delivery` and `wait_for_processing` stages
  - Raise `AssetStateTimeoutError` (instead of generic errors) for all delivery timeouts so they're properly handled and logged

### DIFF
--- a/test/models/apple/publisher_test.rb
+++ b/test/models/apple/publisher_test.rb
@@ -607,7 +607,7 @@ describe Apple::Publisher do
       # Simulate: episode1 delivered successfully, episode2 still waiting when timeout
       # wait_for_delivery returns [timed_out, still_waiting_pdfs]
       # Only pdf2 is still waiting
-      wait_for_delivery_stub = ->(api, pdfs, &block) {
+      ->(api, pdfs, &block) {
         [true, [pdf2]]  # Timeout with only episode2's PDF still waiting
       }
     end


app/jobs/publish_feed_job.rb
```diff
@@ -54,11 +54,11 @@ def publish_integration(podcast, feed)
     # We always mark the integration as errored in the pipeline state
     PublishingPipelineState.error_integration!(podcast)
 
+    # Log at the error's specified level (INFO, WARN, or ERROR)
+    e.log_error!
```
svevang: Make sure we log irrespective of the `sync_blocks_rss` flow branches below.

app/models/apple/episode.rb
```diff
@@ -12,9 +12,6 @@ class Episode < Integrations::Base::Episode
     AUDIO_ASSET_FAILURE = "FAILURE"
     AUDIO_ASSET_SUCCESS = "SUCCESS"
 
-    EPISODE_ASSET_WAIT_TIMEOUT = 15.minutes.freeze
-    EPISODE_ASSET_WAIT_INTERVAL = 10.seconds.freeze
```
svevang: Removing unused code.

app/models/apple/podcast_delivery_file.rb
```diff
@@ -44,32 +44,33 @@ class DeliveryFileError < StandardError; end
       end
     end
 
-    def self.wait_for_delivery_files(api, pdfs)
-      wait_for_delivery(api, pdfs)
-      wait_for_processing(api, pdfs)
-    end
```
svevang: Unused

app/models/integrations/episode_delivery_status.rb
```diff
@@ -56,7 +56,7 @@ def mark_as_delivered!
     end
 
     def mark_as_not_delivered!
-      self.class.update_status(integration, episode, delivered: false, uploaded: false)
+      self.class.update_status(integration, episode, delivered: false, uploaded: false, asset_processing_attempts: 0)
```
svevang: Fixes the case where an episode is stuck, then this  method is called to force re-upload. But since the `asset_processing_attempts` was not-reset, the calculated 'duration' doesn't move -- the episode is still stuck.

This change ensures that the duration calculation resets with this method call.

All this would go away when we switch to timestamps (e.g. uploaded_at, delivered_at).

app/models/apple/asset_state_timeout_error.rb
```diff
@@ -1,46 +1,41 @@
...
+      duration = asset_wait_duration || 0
+      if duration >= Apple::STUCK_EPISODE_THRESHOLD
         :error
+      elsif duration >= Apple::SLOW_EPISODE_THRESHOLD
+        :warn
       else
-        :fatal
+        :info
       end
     end
```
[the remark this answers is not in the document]
kookster: > I'd like to add uploaded_at and delivered_at timestamps into the apple_episode_delivery_status model

yeah, I think those are on the sync log? that makes some sense, and I agree, is a bit of a bigger change, as it would also affect the megaphone use of those models...at least if we want those sync and delivery status models to continue to be used consistently.
Maybe a new ticket for that refactor?

PR #1412, opened 2026-01-20
Multiple recording callbacks

### LINKED ISSUE #1411: StreamRecordings - handle multiple region callbacks
Fallout from #1402.

In many cases, we're going to be recording streams from 2 regions simultaneously.  So we can expect the SQS callbacks from those to come in very close to one another.

Do a better job of this in `RecordStreamTask.update_stream_resource`.

- [ ] Ensure only 1 `StreamResource` will ever get created for that hour
- [ ] Handle 2+ recordings completing simultaneously, and make sure we only copy 1
- [ ] Don't flip the status back from `processing` (copying) to `recording`
- [ ] Why are we running the "fix audo" tasks on these?

cavis:
Closes #1411.

Puts some gatekeeping on stream recording callbacks:

- [x] Custom `Tasks::RecordStreamTask.with_lock` logic, that also locks the parent `StreamResource` or the grandparent `StreamRecording` anytime we're processing a callback.
- [x] Refactor `Task`s to have a shared `before_save :update_owner` method
- [x] Allow cancelling in progress `Task`s - they will update their own result, but won't `.update_owner`

### DIFF
--- a/app/models/task.rb
+++ b/app/models/task.rb
@@ -97,7 +97,7 @@ class Task < ApplicationRecord
     owner && status_changed? && !cancelled?
   end
 
-  # after save hook, implemented by child tasks
+  # before save hook, implemented by child tasks
   def update_owner
   end
 end


app/models/stream_resource.rb
```diff
@@ -29,13 +28,23 @@ def set_defaults
...
       Tasks::CopyMediaTask.create! do |task|
         task.owner = self
       end.start!
     end
   end
 
+  def needs_copy?
+    if status_complete?
+      false
+    elsif done_recording? && copy_task
```
cavis: It's important the `has_one :copy_task` ignores canceled tasks, so we kick off a new one whenever there's a new recording to copy.

app/models/task.rb
```diff
@@ -20,9 +20,14 @@ def result
...
 
   before_validation { self.status ||= :started }
+  before_save :update_owner, if: :update_owner?
 
+  scope :analyze_media, -> { where(type: "Tasks::AnalyzeMediaTask") }
   scope :copy_media, -> { where(type: "Tasks::CopyMediaTask") }
   scope :copy_image, -> { where(type: "Tasks::CopyImageTask") }
+  scope :copy_transcript, -> { where(type: "Tasks::CopyTranscript") }
+  scope :fix_media, -> { where(type: "Tasks::FixMediaTask") }
+  scope :record_stream, -> { where(type: "Tasks::RecordStreamTask") }
```
kookster: These are handy - almost seems like a loop could create the scopes based on a list or something, but this is about the same amount of code

Issue #1414, opened 2026-01-29
Ignore DUPLICATE asset state in delegated delivery file processing


svevang:
The file processing step in https://github.com/PRX/feeder.prx.org/pull/1403 adds retry logic for the duplicate asset state error. However this approach was somehow masking the checks on Apple's side of things. Let's modify the approach in #1403 and simply ignore the duplicate asset state. Because, if the media in the podcast_container is actually duplicated, we can proceed to the final delivery phase of the the Apple::Publisher.

PR #1415, opened 2026-01-30
Remove duplicate check

svevang:
solves #1414


kookster (approved):
Looks right to me, and the tests show the dupe is logged now, not raised 👍

PR #1423, opened 2026-02-19
Feat/metrics release to prod

radical-ube:
This PR releases the metrics to feeder rollout into production.

### DIFF
--- a/app/models/concerns/metrics_queries.rb
+++ b/app/models/concerns/metrics_queries.rb
@@ -13,7 +13,7 @@ module MetricsQueries
       .final
   end
 
-  def daterange_downloads_query(model_id: metrics_default_id, column: metrics_default_column, date_start: metrics_default_time_start, date_end: metrics_default_time_end, interval: "DAY")
+  def daterange_downloads_query(model_id: metrics_default_id, column: metrics_default_column, date_start: metrics_default_date_start, date_end: nil, interval: "DAY")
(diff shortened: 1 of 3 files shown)


app/helpers/metrics_helper.rb
```diff
@@ -1,4 +1,12 @@
 module MetricsHelper
+  def show_metrics?
+    if Rails.env.production?
+      false
```
cavis: Oh, so realizing, you removed the old `podcast#show` chart (coming from Castle), but we don't replace it with anything yet.  If this is returning false.

So I'd say, get rid of the if condition, and we just wait to merge this PR until clickhouse performance is sorted out - (that's on me, and I'm getting to it soon i hope!).

app/models/concerns/metrics_queries.rb
cavis: Okay, so I did some QA-ing to compare numbers, and found these bounds to be slightly different than what Metrics uses.

I'm going to suggest that we just pick a consistent window for everything, when we say "last 28 days".  And we do the same thing Metrics does ... where "utc today" is _the 28th_ day, and incomplete.

So:

- This would need to be `Date.utc_today - 27.days`
- I _think_ we should use the same def even on hourly downloads?  So change `metrics_default_time_start` to be the same beginning-of-UTC-day?
- And maybe just have the default `date_end` / `time_end` return nil?  There is no upperbound on these ranges.

PR #1424, opened 2026-02-19
WCAG 2.2 audit & fixes

### LINKED ISSUE #1418: WCAG 2.2 AA audit
Prompted by a discussion with a potential client that needs to be WCAG 2.2 AA compliant. Regardless of if they choose Dovetail or not, we should ensure our pages in Feeder are compliant as that aligns with latest government standards, as we have a lot of stations associated with universities.

*Lighthouse is not necessarily built to this standard, test should be completed with the [Deque Axe web extension](https://chromewebstore.google.com/detail/axe-devtools-web-accessib/lhdoppojpmngadmnindnejefpokejbdd)

brandonhundt:
Closes #1418

To review you are going to want to do download the [Axe Dev tools](https://chromewebstore.google.com/detail/axe-devtools-web-accessib/lhdoppojpmngadmnindnejefpokejbdd) Chrome Extension and set it to the WCAG 2.2 AA standard. Go through the app to confirm I didn't miss any issues. (Can compare to the live prod site to see existing issues).

*Note there's one issue in Play that is causing one issue where we have the embed player. I opened a ticket for that.

Some of this was easy to fix by hand. ~Other stuff, like Slim Select was less so, so I went to the LLMs to fix, therefore, its possible some of the fixes are bloated or not quite there. It didn't necessarily feel smooth, took me a few convos to get to this. (I also want to test how the LLM's respond to Senior Dev feedback, so I'll keep trying this.)~ EDIT: Removed from the code as a Slim Select upgrade seems to solve these issues.

Also, some of these files may need to be shared across our other frontend repos once this is merged.

https://github.com/user-attachments/assets/21358454-be64-4f86-b7be-4895593785c0

### DIFF
--- a/app/views/imports/_import_progress.html.erb
+++ b/app/views/imports/_import_progress.html.erb
@@ -2,14 +2,14 @@
   &mdash;
 <% elsif import.undone? %>
   <div class="progress flex-grow-1 me-2">
-        <div class="progress-bar progress-bar-striped progress-bar-animated"
-          role="progressbar"
-          aria-label="Import progress"
-          style="width: <%= import_progress(import) %>%; min-width: 100px;"
-          aria-valuenow="<%= import_progress(import) %>"
-          aria-valuemin="0"
-          aria-valuemax="100">
-        </div>
+    <div class="progress-bar progress-bar-striped progress-bar-animated"
+      role="progressbar"
+      aria-label="<%= t("imports.progress") %>"
+      style="width: <%= import_progress(import) %>%; min-width: 100px;"
+      aria-valuenow="<%= import_progress(import) %>"
+      aria-valuemin="0"
+      aria-valuemax="100">
+    </div>
   </div>
 <% else %>
   <span class="badge rounded-pill prx-badge-<%= import.status %>">
--- a/config/importmap.rb
+++ b/config/importmap.rb
@@ -10,7 +10,6 @@ pin "@rails/ujs", to: "@rails--ujs.js", preload: true # @7.0.4
 pin "popper", to: "popper.js", preload: true
 pin "bootstrap", to: "bootstrap.min.js", preload: true
 pin "flatpickr", preload: true # @4.6.13
-pin "accessibility/sanitize_aria", to: "accessibility/sanitize_aria.js"
 pin "bootstrap5-tags", to: "https://ga.jspm.io/npm:bootstrap5-tags@1.6.4/tags.js"
 pin "trix", to: "https://ga.jspm.io/npm:trix@2.0.5/dist/trix.esm.min.js", preload: true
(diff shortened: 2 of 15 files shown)


app/views/imports/_import_progress.html.erb
```diff
@@ -2,13 +2,14 @@
...
 <% elsif import.undone? %>
   <div class="progress flex-grow-1 me-2">
-    <div class="progress-bar progress-bar-striped progress-bar-animated"
-         role="progressbar"
-         style="width: <%= import_progress(import) %>%; min-width: 100px;"
-         aria-valuenow="<%= import_progress(import) %>"
-         aria-valuemin="0"
-         aria-valuemax="100">
-    </div>
+        <div class="progress-bar progress-bar-striped progress-bar-animated"
```
cavis: This seems over-indented now.

app/javascript/controllers/slim_select_controller.js
```diff
@@ -20,6 +20,41 @@ export default class extends Controller {
         beforeChange: this.beforeChange.bind(this),
       },
     })
+    // Ensure ARIA roles are applied to the actual list element (fix accessibility scanners)
```
cavis: Hrm... I have questions about all this sanitize-aria stuff...

app/views/metrics/_agent_apps_card.html.erb
```diff
@@ -1,11 +1,13 @@
...
-    <% end %>
+  <div class="position-relative">
+    <ul class="list-group list-group-flush">
+      <% agents.map do |agent| %>
+        <li class="list-group-item d-flex justify-content-between">
+          <span class="z-1"><%= agent[0] %></span>
+          <span class="z-1"><%= number_with_delimiter(agent[1]) || 0 %></span>
+        </li>
+      <% end %>
+    </ul>
```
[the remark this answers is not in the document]
cavis: Oh wait, I see.  It was a bunch of `<li>` and then a `<div>`.  I get it.  +1.

config/importmap.rb
```diff
@@ -10,6 +10,7 @@
 pin "popper", to: "popper.js", preload: true
 pin "bootstrap", to: "bootstrap.min.js", preload: true
 pin "flatpickr", preload: true # @4.6.13
+pin "accessibility/sanitize_aria", to: "accessibility/sanitize_aria.js"
```
cavis: Generally we don't pin `to:` a place - we leave these in the vendor directory (or many don't even get stored in this git repo at all).  Is that part necessary?

cavis (changes requested):
Ok ok, I think my main ask is... we should translate all these aria labels.  (Unless you disagree on that).

Otherwise a couple minor things, and we should figure out what to do with slimselect.

cavis (approved):
👍 looks good to me!

PR #1425, opened 2026-02-19
Attach clips to episodes

cavis:
Implements creating/attaching 1 hour recorded clips to episodes.

Note that this literally makes a synchronous S3 copy_object in the controller action.  Which generally seems to take 1-2 seconds to complete.  But - you instantly get redirected to the episode media page with the waveform displayed.

image

image

### DIFF
--- a/app/models/concerns/copy_resource.rb
+++ b/app/models/concerns/copy_resource.rb
@@ -41,6 +41,7 @@ module CopyResource
     copy_resource_s3_client.copy_object(bucket: bucket, key: to_path, copy_source: from_src)
     "s3://#{bucket}/#{from_path}"
   rescue Aws::S3::Errors::NoSuchKey
+    Rails.logger.warn("copy_resource_s3 no such key", bucket: bucket, path: to_path, copy_source: from_src)
     nil
   end
 


app/controllers/podcast_clips_controller.rb
```diff
@@ -9,14 +9,30 @@ def index
...
 
   def show
   end
 
+  def attach
+    if (ep = attach_to_episode(clip_params[:episode]))
+      redirect_to episode_media_path(ep), notice: t(".attached")
+    elsif (ep = create_new_episode(clip_params[:title], clip_params[:released_at]))
+      redirect_to episode_media_path(ep), notice: t(".created")
+    elsif request.referrer&.ends_with?("/clips")
```
kookster: not sure what this case is, and why this is needed in addition to the "else" redirect, but I believe you.

app/controllers/podcast_clips_controller.rb
```diff
@@ -27,4 +43,22 @@ def set_clip
     @clip = @podcast.stream_resources.find(params[:id])
     authorize @clip
   end
+
+  def attach_to_episode(guid = nil)
+    ep = @podcast.episodes.find_by_guid(guid) if guid
+    return unless ep
+
+    uncut = @clip.copy_resource_to(ep.build_uncut)
```
kookster: I see, this is the in controller thread "copy" call - I think that's fine, s3 copy is pretty speed.

app/models/concerns/copy_resource.rb
```diff
@@ -0,0 +1,58 @@
...
+      if respond_to?(k) && to_resource.respond_to?(:"#{k}=")
+        to_resource.send(:"#{k}=", send(k))
+      end
+    end
+
+    to_resource
+  rescue => err
+    Rails.logger.error("copy_resource_to error", error: err)
+    NewRelic::Agent.notice_error(err)
+    to_resource.status = "error"
```
kookster: Does every possible resource have a status that includes "error" as a possible value?

app/models/concerns/copy_resource.rb
```diff
@@ -0,0 +1,58 @@
...
+  private
+
+  # S3 copying will be faster
+  def copy_resource_s3(from_path, to_path)
+    bucket = ENV["FEEDER_STORAGE_BUCKET"]
+    from_src = "/#{bucket}/#{from_path}"
+    copy_resource_s3_client.copy_object(bucket: bucket, key: to_path, copy_source: from_src)
+    "s3://#{bucket}/#{from_path}"
+  rescue Aws::S3::Errors::NoSuchKey
+    nil
```
kookster: just "eating" this error? no logging or raising?

app/helpers/stream_recordings_helper.rb
```diff
@@ -80,4 +80,11 @@ def stream_date_hour(resource)
       "#{date} #{hour}"
     end
   end
+
+  def stream_episode_options(podcast)
+    @stream_episode_options ||= {}
+    @stream_episode_options[podcast.id] ||= podcast.episodes.draft_or_scheduled.dropdate_asc.map do |e|
```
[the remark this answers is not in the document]
cavis: Oh... yeah, good call.  Will update.

I also thought, maybe I shouldn't allow attaching to episodes already having audio?  But, can always add more restrictions later.

app/models/concerns/copy_resource.rb
cavis: In local dev I have my bucket set to `prx-feed-development`, but my db is a prod backup.  So those S3 keys only exist out in `prx-feed`.  Catching the `copy_source` not existing here is what lets it retry HTTP-downloading.

I'll add a log.warning, but since we retry the HTTP download method, it's not really a fatal error yet.

PR #1427, opened 2026-02-20
Save uploaded image urls to database

cavis:
Apparently we've been leaving `url: nil` in the db for uploaded images and transcripts.

This is fine in prod (they're generated just in time).  But since they're not in the db, I was seeing all sorts of broken images locally, running of a prod copy.

### DIFF
--- a/test/models/apple/publisher_test.rb
+++ b/test/models/apple/publisher_test.rb
@@ -594,6 +594,7 @@ describe Apple::Publisher do
     end
 
     it "only includes still-waiting episodes in timeout error, not all episodes" do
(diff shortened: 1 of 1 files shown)


app/models/concerns/image_file.rb
```diff
@@ -54,6 +54,7 @@ def published_url
   def initialize_attributes
     self.status ||= :created
     guid
+    url
```
cavis: Note: `media_resource.rb` was already doing this.  But not images/transcripts.  So we both needed to call this, and have the `published_url` method return nil if there is no parent of the image/transcript.

Issue #1417, opened 2026-02-04
Delegated delivery: Prevent reuploading unchanged media


svevang:
- probe audio version
- prevent reupload based on that

PR #1431, opened 2026-02-26
Apple drafts upload media

### LINKED ISSUE #1382: Publish audio to Apple when it is a new media version, even for draft episodes
The goal is to be able to upload the up to date audio to Apple as soon as it is processed on feeder.

That usually means when the audio is ready to publish, and there is a new media version for it that DTR can then serve.

This would mean several changes, such as the publishing pipeline and apple publisher needing to look at drafts as well.

The existing delivery status "uploaded" could be used to track when audio needs to be uploaded or not, and the version number to keep track of what version was most recently sent to Apple.

svevang:
closes  https://github.com/PRX/feeder.prx.org/issues/1382

- Draft and scheduled episodes with complete media (enclosure_ready?(true)) are now treated as part of the Apple subscription feed and uploaded to Apple as drafts
- Media upload happens immediately for eligible drafts, but final delivery (publishing on Apple's side) is gated until the episode is actually published
- UI badges now show "uploaded" status for draft/scheduled episodes that have been synced to Apple

`AppleSubscription#feed_episode?` extends the base feed check so unpublished episodes with complete enclosures are considered feed members (via `draft_or_scheduled` scope + `enclosure_ready?(true)`).

Does not implement the dupe media check to avoid reuploading https://github.com/PRX/feeder.prx.org/issues/1417. That will come in a second PR.


app/models/apple/show.rb
```diff
@@ -70,6 +70,31 @@ def initialize(api:, public_feed:, private_feed:)
...
+      podcast_episodes
+        .filter { |e| draft_ids.include?(e.feeder_episode.id) }
+    end
+
+    def episodes
+      @episodes ||= begin
+        published = super
+        published_ids = Set.new(published.map { |e| e.feeder_episode.id })
+        drafts = draft_upload_candidates.reject { |e| published_ids.include?(e.feeder_episode.id) }
+        published + drafts
```
svevang: Add in the draft episodes at the top of the publisher call chain.

app/helpers/episodes_helper.rb
```diff
@@ -16,8 +16,9 @@ def episode_explicit_options
   end
 
   def episode_integration_status(integration, episode)
-    return "draft" if episode.draft?
-    return "not_publishable" unless episode.integration_feed_episode?(integration)
+    publishable = episode.integration_feed_episode?(integration)
+    return "draft" if episode.draft? && !publishable
+    return "not_publishable" unless publishable
```
kookster: publishable checks if the podcast publishes to this integration, and if the episode is included in the feed for that integration.

So wouldn't we want to check first that this episode is included in the feed at all, and calling that not_publishable, and then if it is included in the feed but is a draft, that would be "draft"?

Maybe I'm confused about what `publishable` here means....

app/helpers/episodes_helper.rb
```diff
@@ -30,7 +31,7 @@ def episode_integration_status(integration, episode)
     elsif episode.integration_error_state?(integration)
       "error"
     elsif !status.delivered?
-      "processing"
+      episode.integration_processing_status(integration)
```
kookster, line 32: +1 that's an improvement

app/models/apple/episode.rb
```diff
@@ -272,22 +272,24 @@ def create_episode_bridge_params
...
+        title: feeder_episode.title_safe,
+        description: feeder_episode.description_safe,
+        websiteUrl: feeder_episode.url,
+        explicit: explicit,
+        episodeNumber: feeder_episode.episode_number,
+        seasonNumber: feeder_episode.season_number,
+        episodeType: feeder_episode.itunes_type.upcase,
+        appleHostedAudioIsSubscriberOnly: true
+      }
+      attributes[:originalReleaseDate] = (feeder_episode.published_at || Time.now.utc).iso8601
```
kookster: This is a default for a draft, that won't necesarilly have a pub date? should we be using the scheduled date if it isn't published yet? would a date in the future be an issue? I don't think so, I think the assets would still be able to process?

app/models/apple/publisher.rb
```diff
@@ -90,11 +90,24 @@ def publish!
 
     def upload_and_process!(eps)
       Rails.logger.tagged("Apple::Publisher#upload_and_process!") do
-        eps.filter(&:apple_needs_upload?).each_slice(PUBLISH_CHUNK_LEN) do |batch|
+        eps, skipped = eps.partition { |ep| ep.feeder_episode.enclosure_ready?(true) }
+        skipped.each do |ep|
+          Rails.logger.warn("Episode needs ready enclosure. Skipping", {episode_id: ep.id})
```
kookster, line 95: +1 helpful warning

app/models/apple/publisher.rb
```diff
@@ -90,11 +90,24 @@ def publish!
...
+        eps
+          .select(&:apple_needs_upload?)
+          .each_slice(PUBLISH_CHUNK_LEN) do |batch|
           upload_media!(batch)
         end
 
-        eps.filter(&:apple_needs_delivery?).each_slice(PUBLISH_CHUNK_LEN) do |batch|
+        eps
+          .filter(&:apple_needs_delivery?)
+          .filter { |ep| ep.feeder_episode.published? }
```
kookster, line 109: +1 so upload (above) we do for eligible episodes that can now contain drafts, but here for process_delivery we only do for the published episodes

app/models/apple/publisher.rb
```diff
@@ -109,8 +122,7 @@ def upload_media!(eps)
         # Soft delete any existing delivery and delivery files.
         prepare_for_delivery!(eps)
 
-        # Only create if needed.
-        sync_episodes!(eps)
```
kookster, line 113: So this is basically replaced by the above? (line 99?)

app/models/apple/show.rb
```diff
@@ -70,6 +70,31 @@ def initialize(api:, public_feed:, private_feed:)
       @api = api
     end
 
+    # Draft inclusion is Apple-only (not in Base::Show) because other
+    # integrations don't need drafts. Gate on enclosure_ready? to prevent
```
kookster: fwiw - I think the megaphone integration does send the episode audio as soon as it is available and ready, even for drafts

app/models/feeds/apple_subscription.rb
```diff
@@ -132,6 +132,20 @@ def publish_integration?
...
+  def draft_episodes
+    episodes.draft_or_scheduled
+  end
+
+  def feed_episode?(episode)
+    if episode.published?
+      feed_episodes.where(id: episode.id).exists?
+    else
+      # Use enclosure_ready? (not feed_ready?) to exclude medialess drafts from UI status
+      episode.enclosure_ready?(true) &&
```
kookster: I might put this in an elsif instead of inside the else, since it's another status on the episode you're checking, and then just false in the else? idk, not important

app/models/integrations/base/episode.rb
```diff
@@ -28,9 +32,9 @@ def video_content_type?
       end
 
       def has_media_version?
-        return false unless delivery_status.present? && delivery_status.source_media_version_id.present?
+        return false unless delivery_status.present?
 
-        delivery_status.source_media_version_id == feeder_episode.media_version_id
+        !delivery_status.needs_media_version?
```
kookster, line 37: +1 Nice to use a helper for this logic

kookster (commented):
A few questions? generally looks great

app/helpers/episodes_helper.rb
svevang: `publishable` here is an overloaded and too broad. We should probably use `in_integration_feed` to tighten this down.

+1, for drafts to show as "draft" and then transition to not_publishable upon publish. And I agree: it would be a bit more consistent to show "not_publishable" immediately.

app/models/apple/episode.rb
svevang: Good point will update this to include the `released_at` column as well. After all, drafts are usually queued up for release.

app/models/apple/publisher.rb
svevang: yup, this comes from bc1dd1b1 — "Ensure that the draft episodes are patched" .. basically, the intent here is to update attributes (as we can) for remote apple state. But now we need to do it above, in the regime that _includes_ drafts.

kookster: awesome

kookster (approved):
changes look great, and ty for clarifying!

PR #1432, opened 2026-03-02
Fix duplicate media changed version

### LINKED ISSUE #1417: Delegated delivery: Prevent reuploading unchanged media

svevang:
closes #1417 

- Prevent uploading unchanged media.
- This also sets up new state handling (MediaInfo) so that we can coordinate `uploaded?`.
- `media_version_id` changes, this triggers pulling episodes into the upload routine.


app/models/apple/publisher.rb
```diff
@@ -113,22 +112,23 @@ def upload_media!(eps)
         sync_episodes!(eps)
         sync_podcast_containers!(eps)
 
-        wait_for_versioned_source_metadata(eps)
+        media_infos = wait_for_versioned_source_metadata(eps)
```
svevang: This maps across the episodes and returns a set of MediaInfo objects. There are the heart of this refactor. So instead of storing state in the delivery status model, or computing in various places, we compute the state once, and the write in atomically at the end of the routine.

The problem was that the `source_media_version_id` and the `uploaded?` state are conjoined but were being recorded separately.  Effectively, this made it really hard to reason about any one aspect of the state. With this refactor, we pull the state management up from the call-chain leafs and put it in a single single place.

app/models/apple/publisher.rb
```diff
@@ -250,9 +250,10 @@ def raise_for_processing_state(eps, filter_method)
     def wait_for_versioned_source_metadata(eps)
       Rails.logger.tagged("##{__method__}") do
         # wait for the audio version to be created
-        (waiting_timed_out, _) =
-          Apple::PodcastContainer.wait_for_versioned_source_metadata(api, eps)
+        (waiting_timed_out, media_infos) =
+          Apple::MediaInfo.wait_for_versioned_source_metadata(api, eps)
```
svevang: Extracts all the "probe" related code out into a dedicated Apple::MediaInfo

db/migrate/20260302000000_drop_source_fetch_count_from_podcast_containers.rb
```diff
@@ -0,0 +1,5 @@
+class DropSourceFetchCountFromPodcastContainers < ActiveRecord::Migration[7.2]
+  def change
+    remove_column :apple_podcast_containers, :source_fetch_count, :integer, default: 0, null: false
```
svevang: Cleanup. This was no longer being used.

app/models/apple/api_join.rb
```diff
@@ -0,0 +1,73 @@
+# frozen_string_literal: true
+
+module Apple
+  module ApiJoin
```
svevang: MediaInfo is not an "ApiResponse". But we needed some of the `join_*` functions here.  So this extracts the api response join logic to a simple utility module.

test/models/apple/publisher_entry_state_test.rb
```diff
@@ -0,0 +1,525 @@
+# frozen_string_literal: true
+
+require "test_helper"
+
+# Entry State Matrix for Apple::Publisher#upload_and_process!
```
svevang: Trying something here, formulating the tests as coverage over the pipeline. Some of this is duped with existing coverage

app/models/integrations/episode_delivery_status.rb
```diff
@@ -56,7 +56,22 @@ def mark_as_delivered!
...
+      # current one.
       self.class.update_status(integration, episode, delivered: false, uploaded: false, asset_processing_attempts: 0)
     end
+
+    def needs_upload?
+      !uploaded || !has_media_version?
+    end
+
+    def has_media_version?
+      MediaVersion.current?(source_media_version_id, episode.media_version_id)
```
kookster: `has_media_version?` is a 1 line function calling another 1 line function `current?`, which seems like it isn't that helpful?

app/models/media_version.rb
```diff
@@ -7,4 +7,8 @@ class MediaVersion < ApplicationRecord
   scope :latest, -> { where("id IN (SELECT max(id) FROM media_versions mv WHERE mv.episode_id = media_versions.episode_id)") }
 
   validates :media_version_resources, length: {minimum: 1}
+
+  def self.current?(candidate_id, expected_id)
```
kookster: current as a class method I might expect to get the latest, and compare that to a candidate, but just doing a comparison seems like not the most useful method?

app/models/apple/publisher.rb
kookster: and MediaInfo isn't an AR model, so this value object has the logic, but then we still end up with the state in those 2 places, yah? makes sense to keep them together for the processing +1

app/models/apple/publisher.rb
```diff
@@ -109,26 +111,25 @@ def upload_media!(eps)
         # Soft delete any existing delivery and delivery files.
         prepare_for_delivery!(eps)
 
-        # Only create if needed.
-        sync_episodes!(eps)
         sync_podcast_containers!(eps)
 
-        wait_for_versioned_source_metadata(eps)
+        media_infos = wait_for_versioned_source_metadata(eps)
+        versioned_eps = media_infos.map(&:episode)
```
kookster: so you get the media info from the episodes, then extract episodes back from the media info?

seems like `wait_for_versioned_source_metadata` has a side effect of filtering episodes, instead of being explicit?

Also not clear to me why these are "versioned" episodes? I think they need to be delivered, or are in the process of being delivered?

app/models/apple/publisher.rb
```diff
@@ -385,22 +387,23 @@ def poll_podcast_delivery_files!(eps)
       end
     end
 
-    def sync_podcast_delivery_files!(eps)
+    def sync_podcast_delivery_files!(media_infos)
+      eps = media_infos.map(&:episode)
```
kookster: In the method where this is called, the versioned_episodes were just extracted from the media infos.
maybe just pass them both in? Or maybe the polling method should work off the media infos?

app/models/integrations/episode_delivery_status.rb
```diff
@@ -56,7 +56,22 @@ def mark_as_delivered!
...
 
     def mark_as_not_delivered!
+      # source_media_version_id is intentionally omitted — it is preserved so
+      # we can still compare the previously uploaded media version against the
+      # current one.
       self.class.update_status(integration, episode, delivered: false, uploaded: false, asset_processing_attempts: 0)
     end
+
+    def needs_upload?
+      !uploaded || !has_media_version?
```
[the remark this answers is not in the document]
kookster: ok, and I see that gets used in `apple_needs_upload?`, so that logic gets used all the places that gets called +100

kookster (approved):
The core change which is to check media version as well uploaded status before uploading again looks good.

A few minor things I might do differently, but not important.

app/models/apple/publisher.rb
svevang: > seems like wait_for_versioned_source_metadata has a side effect of filtering episodes, instead of being explicit?

It doesn't filter any episodes, only [waits until all the episodes' respective versions settle or raises a timeout](https://github.com/PRX/feeder.prx.org/pull/1432/changes#diff-c7e01eae960681272b9e7feca67ec9c3ccd4ef420b18fd165990ed80aa424ad4R76).

Yeah, "versioned" was meant to mean the source metadata has caught up to the current Feeder media version, but that's not super clear. I'm trying to establish the contract by showing the connection between episodes and media_infos.

It's not clear that:
```
Set(eps) == Set(media_infos.map(&:episode))
```

I'm going to rework to clarify that equality.

app/models/media_version.rb
svevang: This is used in both EpisodeDeliveryStatus and Apple::MediaInfo, so it was meant to centralize the comparison contract .. it's a bit tedious to write out / handle the nil case and compare. Also I wanted to guard on future changes to the contract.

PR #1436, opened 2026-03-04
Replace uncut media

### LINKED ISSUE #1164: Can't replace Uncut media in single step
Bug with replacing audio for an Uncut episode.  Does impact published eps.

1. Upload/segment an Uncut episode
2. Publish the episode
3. Delete and upload a new audio file, click save
4. Uncut is deleted (!!!) on the published episode, but new file isn't saved.  Error message shown.
5. After uploading again, the Uncut still needs ad breaks marked, but the little media files tab badge shows "complete" anyways.

What the huh?

cavis:
Fixes #1164.

### DIFF
--- a/app/controllers/episode_media_controller.rb
+++ b/app/controllers/episode_media_controller.rb
@@ -26,7 +26,9 @@ class EpisodeMediaController < ApplicationController
     @episode.assign_attributes(parsed_episode_params)
 
     # when an uncut is destroyed, also destroy sliced contents
-    @episode.contents.each(&:mark_for_destruction) if @episode.uncut&.marked_for_destruction?
+    if @episode.uncut&.marked_for_destruction? || @episode.uncut&.new_record?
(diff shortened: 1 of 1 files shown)


app/views/episode_media/_form_uncut.html.erb
```diff
@@ -12,7 +12,7 @@
       </div>
     </div>
 
-    <% if fields.object.status_complete? %>
+    <% if fields.object.status_complete? && !fields.object.marked_for_destruction? %>
```
cavis: Side effects!

Previously, the waveform editor for the old Uncut was still displayed when you were replacing it.  And it was sending a `segmentation = []` empty array, which is invalid.  So the old Uncut gets destroyed (the `has_one` autosave strikes again).  But the new one was invalid, so not saved.

Hiding the waveform entirely makes more sense for the UX, and doesn't send that invalid `segmentation` field.

PR #1447, opened 2026-03-09
Hide feeds/seasons charts when not more than 1

### LINKED ISSUE #1442: Hide seasons metrics card when only one season exists
This should help preserve space on the page

radical-ube:
resolves #1442 

This PR updates the conditional to hide the feeds and seasons charts if there aren't more than 1 in the list. I move some methods from `metrics_helper.rb` over to `metrics_queries.rb` and `podcast_metrics.rb` because although they are used as view conditionals, the methods relate to the models and don't stand on their own.

There is also a small fix on the feeds chart. Because the feeds query only pulls data for known downloads, the chart doesn't technically represent when a particular feed has 0 downloads (within the last 28 days). I don't adjust the query itself, rather i adjust the feed labeling to account for any feeds that don't return results.

### DIFF
--- a/app/models/concerns/metrics_queries.rb
+++ b/app/models/concerns/metrics_queries.rb
@@ -33,12 +33,14 @@ module MetricsQueries
     end
   end
 
-  def feed_downloads_query(feeds:, model_id: metrics_default_id, column: metrics_default_column, date_start: metrics_default_date_start, date_end: nil)
-    slugs = feeds.pluck(:slug).map { |slug| slug.nil? ? "" : slug }
+  def metrics_feed_slugs
+    feeds.pluck(:slug).map { |slug| slug.nil? ? "" : slug }
+  end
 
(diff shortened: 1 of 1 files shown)


app/models/concerns/metrics_queries.rb
```diff
@@ -33,18 +33,32 @@ def daterange_downloads_query(model_id: metrics_default_id, column: metrics_defa
...
       .where(podcast_id: metrics_podcast_id)
-      .where("#{column}": model_id, feed_slug: slugs, hour: (date_start..date_end))
+      .where("#{column}": model_id, feed_slug: metrics_feed_slugs, hour: (date_start..date_end))
       .group(:feed_slug)
       .order(Arel.sql("SUM(count) AS count DESC"))
       .final
       .sum(:count)
   end
 
+  def label_feed_results(results)
```
radical-ube: i abstracted the feed labeling into this shared method that both `podcast_metrics` and `episode_metrics` use for `feed_downloads`.

PR #1450, opened 2026-03-12
Allow the key or apple config to be added if missing after create

### LINKED ISSUE #1451: Can't replace API keys on Apple Subscription
You can delete the entire apple feed, and start over, but if the key or apple config are missing, can't replace them.
Came up with folks moving shows between Apple accounts and wanting to replace the API key file.

kookster:
fixes #1451


svevang (approved):
Just a couple non-blocking style nits. Looks good!

PR #1469, opened 2026-04-27
Migrate existing keys

svevang:
This PR brings the Apple side of the integrations up to match the Megaphone side. Adds a `encrypts :key_pem_b64` to the Apple::Key model def.


cavis (approved):
👍

PR #1473, opened 2026-04-30
Handle draft published cycle

### LINKED ISSUE #1382: Publish audio to Apple when it is a new media version, even for draft episodes

### LINKED ISSUE #1470: Delegated delivery: Unarchived episodes can have media pruned
Seen: When episodes have been archived for an extended time, the podcast containers can show no files, but local state is still marking them as uploaded. We should mark unarchived episodes for reupload when they come back into DRAFTING status.

svevang:
This PR improves Apple draft episode reconciliation so local draft episodes stay aligned with their remote Apple publishing state before media upload runs.

followup from #1431
closes #1382
closes #1470

- Separates RSS/private-feed-visible episodes from Apple draft upload candidates.
- Polls draft upload candidates before deciding any transition.
- Handles local draft + Apple ARCHIVED by unarchiving back to DRAFTING.
- Handles local draft + Apple PUBLISHED by cycling ARCHIVE -> UNARCHIVE to reach DRAFTING.
- Resets local Apple delivery state after draft unarchive paths so media is re-uploaded when Apple may have pruned archived assets.
- Keeps draft candidates out of the generic archive/unarchive paths.
- Adds regression coverage for PUBLISHED, ARCHIVED, and already-DRAFTING draft candidates.


app/models/integrations/base/publisher.rb
```diff
@@ -19,7 +19,7 @@ def episodes_to_archive
       end
 
       def episodes_to_unarchive
-        filter_episodes_to_unarchive(show.episodes)
+        filter_episodes_to_unarchive(show.feed_episodes)
```
kookster: Looks like in `Integrations::Base::Show` that `episodes` has been redefined to call `feed_episodes`, so this doesn't change anything?

app/models/apple/show.rb
```diff
@@ -87,7 +87,7 @@ def draft_upload_candidates
 
     def episodes
       @episodes ||= begin
-        published = super
+        published = feed_episodes
```
kookster: Looks like in `Integrations::Base::Show` that `episodes` has been redefined to call `feed_episodes`, so this doesn't change anything?

app/models/integrations/base/show.rb
```diff
@@ -25,15 +25,21 @@ def podcast_episodes
...
           feed_episode_ids = Set.new(private_feed.feed_episodes.feed_ready.map(&:id))
           podcast_episodes
             .filter { |e| feed_episode_ids.include?(e.feeder_episode.id) }
         end
       end
 
+      def episodes
+        @episodes ||= feed_episodes
+      end
+
```
kookster: I'm not seeing the point here - `feed_episodes` and `episodes` do the same thing, why add `feed_episodes`?

kookster (changes requested):
The draft handling logic makes sense, the `feed_episodes` change in Show I don't see what it does?

app/models/apple/publisher.rb
```diff
@@ -88,6 +90,31 @@ def publish!
...
 
+    def sync_drafting_episode_states!
+      eps = show.draft_upload_candidates
+      return if eps.empty?
+
+      poll_episodes!(eps)
+
+      # Handles local draft + Apple ARCHIVED by unarchiving back to DRAFTING
+      # Handles local draft + Apple PUBLISHED by cycling ARCHIVE -> UNARCHIVE to reach DRAFTING
+      eps = eps.reject(&:integration_new?)
```
kookster, line 101: Makes sense. This is rejecting if the `integration_new? -> apple_new? -> !apple_persisted? -> apple_json.present?`
so basically, we're left with episodes that have been saved to apple already. Later in `upload_and_process!` drafts that haven't been saved at all get taken care of. So, as it says on the box, this is evaluating episode which have an existing state with Apple.

kookster (approved):
Looks good to me, thanks for the changes

PR #1474, opened 2026-05-01
Guard asset state

### LINKED ISSUE #1462: Guard for appleHostedAudioAssetState in FAILURE after processing wait completes
After the processing phase and before publishing, set up a guard for the appleHostedAudioAssetState to make sure it is in `SUCCESS`. If it's in `FAILURE` then mark the episode for re-upload and retry the publishing routine.

svevang:
closes https://github.com/PRX/feeder.prx.org/issues/1462

Sets up a guard before publishing against asset states settling into ERROR

### DIFF
--- a/app/models/apple/publisher.rb
+++ b/app/models/apple/publisher.rb
@@ -144,8 +144,10 @@ module Apple
 
         # Wait for the audio asset to be processed by Apple
         wait_for_asset_state(eps) do |ready_eps|
-          ready_eps, errored = ready_eps.partition(&:audio_asset_state_success?)
+          errored, ready_eps = ready_eps.partition(&:audio_asset_state_error?)
+          ready_eps = ready_eps.select(&:audio_asset_state_success?)
           error_audio_state_eps.concat(errored)
+          mark_asset_state_failures_for_reupload!(errored)
 
           log_asset_wait_duration!(ready_eps)
           # Publish the ready episodes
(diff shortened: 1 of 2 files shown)


app/models/apple/publisher.rb
```diff
@@ -140,14 +140,22 @@ def process_delivery!(eps)
 
         wait_for_upload_processing(eps)
 
+        error_audio_state_eps = []
+
         # Wait for the audio asset to be processed by Apple
         wait_for_asset_state(eps) do |ready_eps|
+          errored, ready_eps = ready_eps.partition(&:audio_asset_state_error?)
+          error_audio_state_eps.concat(errored)
+          mark_asset_state_failures_for_reupload!(errored)
```
kookster, line 149: very clear. nice.

app/models/apple/publisher.rb
```diff
@@ -247,6 +255,27 @@ def raise_for_processing_state(eps, filter_method)
...
+          {episode_id: ep.feeder_id,
+           audio_asset_state: ep.audio_asset_state})
+
+        ep.apple_mark_for_reupload!
+      end
+    end
+
+    def raise_asset_state_failure_retry!(failure_eps)
+      raise Apple::RetryPublishingError.new(
+        "Found FAILURE appleHostedAudioAssetState on #{failure_eps.length} episodes, marked for reupload"
```
kookster: It would be helpful to log the episode ids that are failing.

kookster (approved):
Looks good - perhaps log episode ids instead of count?

PR #1475, opened 2026-05-04
hide private feed url when the Public Feed URL is set

### LINKED ISSUE #1312: Hide the f.prxu.org private feed url when a public url is set
Saw an example where a user tried to submit the f.prxu.org private feed when the publicfeeds.net was available, and the correct one to use.

I can't think of a good reason for that private feed url to be visible to an end user - maybe a prx admin or distro staff?

brandonhundt:
closes #1312 

- Hides the Private URL Feed field when the feed is public and the Public Feed URL is set

When Public Feed URL is NOT set on feed
Screenshot 2026-05-04 at 4 20 23 PM

When feed Public Feed URL is set on feed
Screenshot 2026-05-04 at 4 20 12 PM

### DIFF
--- a/app/views/feeds/_form_distribution.html.erb
+++ b/app/views/feeds/_form_distribution.html.erb
@@ -32,7 +32,7 @@
       </div>
 
       <% if feed.persisted? %>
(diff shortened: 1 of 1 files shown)


cavis (approved):
👍

PR #1480, opened 2026-06-03
Prevent json indexing on html error response

svevang:
Fallout from the June 1 batch.  This adds a guard on the local non-bridge api handling to guard against indexing into a response that is a 5xx.


cavis (approved):
👍

PR #1484, opened 2026-06-11
Fix missing podcast container polling

### LINKED ISSUE #1483: Apple returns 404 for local podcast_container
Apple podcast publishing can get stuck when Feeder has local podcast container or delivery records that no longer exist in Apple.

When polling Apple, a missing container or 404 from the container/delivery endpoints should be treated as stale local state. Feeder should clear the local container and dependent delivery records, reset the episode to not uploaded/not delivered, and retry publishing so the episode can recreate a fresh Apple container and continue through upload/delivery.

svevang:
Closes #1483


app/models/apple/podcast_container.rb
```diff
@@ -161,21 +202,23 @@ def self.get_podcast_containers_via_episodes(api, episodes)
...
       api.bridge_remote_and_retry!("getPodcastContainers", formatted_bridge_params, batch_size: 2)
     end
 
     def self.get_urls_for_episode_podcast_containers(api, episode_podcast_containers_json)
-      containers_json = episode_podcast_containers_json["api_response"]["val"]["data"]
-      raise "Unsupported number of podcast containers for episode: #{ep.feeder_id}" if containers_json.length > 1
+      containers_json = episode_podcast_containers_json["api_response"]["val"]["data"] || []
+      if containers_json.length > 1
+        apple_episode_id = episode_podcast_containers_json.dig("request_metadata", "apple_episode_id")
+        raise "Unsupported number of podcast containers for episode: #{apple_episode_id}"
```
kookster, line 161: Did we run into this, or just being careful?

kookster (approved):
pretty straight forward - looks for the out of sync condition, cleans up the objects, and throws an error to cause retry ,+1

app/models/apple/podcast_container.rb
svevang: Just being careful. This has been a check since the original integration. And given the NameError for `ep.feeder_id` this branch has not been exercised. Also, a slight oddity here that this is a has_many relation (returns an enumerable) — though that is maybe an artifact of their json api?

PR #1486, opened 2026-06-22
Fix incomplete episodes filter

### LINKED ISSUE #1481: Episode filters broken
When adding "Incomplete", incomplete episode is still filtered out:

it's there:
Image

and it's gone when "Incomplete" filter applied:
Image

svevang:
closes https://github.com/PRX/feeder.prx.org/issues/1481


cavis (approved):
👍

PR #1491, opened 2026-07-08
HLS video media

cavis:
Allows uploading videos, which will be transcoded to HLS.  The actual Porter HLS transcode task is a TODO - but this gives you an idea how it will work.

image


app/helpers/episodes_helper.rb
```diff
@@ -112,8 +112,14 @@ def episode_media_label(episode, media)
     I18n.t("helpers.label.media_resource.original_url.#{medium}", position: media.position)
   end
 
-  def episode_medium_options
-    Episode.mediums.keys.map { |k| [I18n.t("helpers.label.episode.mediums.#{k}"), k] }
+  def episode_medium_options(episode = nil)
+    ordered_keys =
+      if Rails.env.production? && !episode&.video?
+        %w[audio uncut passthru override]
```
cavis, line 118: Disallow uploading in prod, since these alt enclosures won't _really_ work.

And DTR will just redirect you to the `f.prxu.org` mp4 at the moment.

app/helpers/uploads_helper.rb
```diff
@@ -135,4 +145,9 @@ def upload_status_class(rec)
       "danger"
     end
   end
+
+  def upload_file_size_error(media)
+    size = number_to_human_size(media.max_file_size)
+    t("activerecord.errors.models.media_resource.attributes.file_size.less_than", size: size)
```
cavis, line 151: We restrict audio to < 1GB... this allows up to 10GB for video.

app/models/concerns/episode_enclosure.rb
```diff
@@ -33,28 +33,48 @@ def enclosure_dovetail_url(feed: podcast&.default_feed, prefix: true, auth: nil,
...
     else
       orig_fn = media_file_name || "audio.mp3"
       orig_ext = File.extname(orig_fn)
       orig_base = File.basename(orig_fn, orig_ext)
-      ext = feed&.file_ext || orig_ext[1..] || "mp3"
-      "#{orig_base}.#{ext}"
+      if ext
+        "#{orig_base}.#{ext}"
+      elsif video?
+        "#{orig_base}.#{feed&.file_ext || "mp3"}"
```
cavis, line 55: For video uploads, we transcode an MP3 for the public feed.  So by default, calling `ep.enclosure_url` with no `ext:` param, you get an MP3 enclosure.

Whereas `ep.enclosure_alt_url` is going to give you a M3U8 url to dovetail router.

app/models/concerns/episode_media.rb
```diff
@@ -4,7 +4,7 @@ module EpisodeMedia
   extend ActiveSupport::Concern
 
   included do
-    enum :medium, [:audio, :uncut, :video, :override], prefix: true
+    enum :medium, [:audio, :uncut, :passthru, :override, :video, :video_uncut], prefix: true
```
cavis, line 7: Realized this enum is the integer style... so I've renamed the old `video` to `passthru` (DTR just redirects you to the file).  Naming is hard.

But using `video` / `video_uncut` for the new HLS stuff makes the most sense to me.

At some point we'll want to update DTR to just check `ep.medium` in the API, rather than checking `ep.media[0].mime_type`.

app/models/tasks/copy_media_task.rb
```diff
@@ -24,91 +19,49 @@ def update_owner
...
-    media_resource.save!
-
-    # set an explicit format for ffmpeg to use if possible
-    fmt = porter_callback_inspect.dig(:Audio, :Format) || porter_callback_inspect.dig(:Video, :Format)
-
-    # set an explicit bitrate if vbr
-    bit = next_highest_bitrate if bad_audio_vbr?
-
-    Tasks::FixMediaTask.create!(owner: owner, media_format: fmt, media_bitrate: bit).start!
+    media_resource.after_copy(self) if media_resource.status_complete?
```
cavis, line 29: Moved slicing into its own task, rather than reusing the CopyMediaTask to do it.  Some churn, but I think it comes out cleaner to have more single-purpose tasks.  You'll see the logic calling these reorganized in the models `copy_media` method.

And added this handy `after_copy` hook, which solves a lot of problems when we need to fix VBR, etc.

app/models/uncut.rb
```diff
@@ -13,6 +13,36 @@ def set_defaults
...
+  end
+
+  def copy_media(force = false)
+    if force || needs_copy?
+      Tasks::CopyMediaTask.start!(self)
+    end
+  end
+
+  def after_copy(copy_task)
+    # optionally set ad breaks from ID3 tags
```
cavis, line 31: NOTE: I changed the copy task to just _always_ pull those ad-break tags from the ID3.

But only the `Uncut` model callback does anything with them.

app/representers/api/auth/media_resource_representer.rb
```diff
@@ -1,4 +1,5 @@
 class Api::Auth::MediaResourceRepresenter < Api::MediaResourceRepresenter
   property :href
   property :original_url, writeable: false
+  property :variants, writeable: false
```
cavis, line 4: Okay, we made it to the API!

So in the auth api, we get some additional variant info:

Image

Notice the actual media `href` is that original giant MP4 you uploaded.  Which, is not directly usable by DTR.  We'll need to modify it to check for a `variants.audio` when it gets a request for the enclosure ending with `.mp3`.

And do any TBD alt-video-ing logic with `variants.hls` when it gets a request ending in `.m3u8`.

Open to other suggestions, but preserving `ep.media[0].href` being the file _you_ uploaded seemed most straightforward.

cavis:
Needs some work - we're only going to support uploading an `Uncut` for HLS video episodes.

Then we'll convert that to 1 MP3 (probably?)... mark ad breaks, slice the mp3 into `Content`s, and transcode the mp3 Uncut into a single HLS playlist.

So all the media...

```
# Uncut
https://f.prxu.org/1234/<ep_guid>/<uncut_guid>.mp4
https://f.prxu.org/1234/<ep_guid>/<uncut_guid>.mp4/preview.mp3
https://f.prxu.org/1234/<ep_guid>/<uncut_guid>.mp4/index.m3u8
https://f.prxu.org/1234/<ep_guid>/<uncut_guid>.mp4/audio.m3u8
https://f.prxu.org/1234/<ep_guid>/<uncut_guid>.mp4/audio.ts
https://f.prxu.org/1234/<ep_guid>/<uncut_guid>.mp4/240p.m3u8
https://f.prxu.org/1234/<ep_guid>/<uncut_guid>.mp4/240p.ts
...etc

# Content
https://f.prxu.org/1234/<ep_guid>/<content_1_guid>.mp3
https://f.prxu.org/1234/<ep_guid>/<content_2_guid>.mp3
https://f.prxu.org/1234/<ep_guid>/<content_3_guid>.mp3
```

config/locales/en.yml
```diff
@@ -398,10 +398,12 @@ en:
...
         medium: File Format
         mediums:
-          audio: Audio - upload multiple audio files
-          uncut: Audio - upload a single file and then segment
-          video: Video
+          audio: Audio - upload multiple files
           override: Override - specify external media url
+          passthru: Video - raw file
+          uncut: Audio - upload a single file and then segment
+          video: Video - upload multiple files
```
brandonhundt, line 405: I don't think we need this option. The multiple audio files is a station specific audio workflow. I say we cut it.

app/models/apple/episode.rb
```diff
@@ -507,8 +507,10 @@ def apple_episode_id
       apple_id
     end
 
-    def video_content_type?
-      feeder_episode.video_content_type?
+    # TODO: we do transcode an mp3 audio version of hls-video episodes, but i'm unsure
+    # if those should be included or excluded here
+    def audio?
+      feeder_episode&.audio?
```
cavis: Yeah, a bit unsure on this.  Maybe it should just be `!feeder_episode&.passthru`?  Would we want to send external media to Apple?

app/models/concerns/episode_media.rb
```diff
@@ -4,7 +4,7 @@ module EpisodeMedia
   extend ActiveSupport::Concern
 
   included do
-    enum :medium, [:audio, :uncut, :video, :override], prefix: true
+    enum :medium, [:audio, :uncut, :passthru, :override, :video], prefix: true
```
cavis: Realized this enum is the integer style... so I've renamed the old video to passthru (DTR just redirects you to the file). Naming is hard.

But reclaiming `video` for the new HLS stuff makes the most sense to me.

PR #1493, opened 2026-07-08
Apple show feed binding

### LINKED ISSUE #1488: HLS Video: Binding schema, model, backfill, sync writer
Prepare for the Apple HLS video config changes. Bindings are extracted from existing Apple configs and kept fresh on every save, while production reads stay untouched.
 
- Add the `apple_show_feed_bindings` table
- Add the `Apple::ShowFeedBinding` model with validations
- Give `Apple::Key` its `has_many :show_feed_bindings`
- Ship the backfill migration with the idempotent `backfill_show_feed_bindings!` / `verify_routing_equivalence!` service
- Hook the legacy `AppleSubscription` save path to upsert the binding (A5.3) so bindings never drift from legacy routing
- *No routing read behavior changes*

```
apple_show_feed_bindings
  feed_id        bigint    FK → feeds, null: false, unique
  apple_key_id   bigint    FK → apple_keys, null: false
  apple_show_id  string    null: false
  timestamps

apple_configs
  + show_feed_binding_id  bigint  FK → apple_show_feed_bindings, nullable
```

svevang:
closes https://github.com/PRX/feeder.prx.org/issues/1488


app/models/apple/show_feed_binding/backfill.rb
```diff
@@ -0,0 +1,234 @@
...
+
+        unless dry_run
+          binding.apple_key = key
+          binding.apple_show_id = legacy_show_id
+          binding.save!
+
+          if config.show_feed_binding_id != binding.id
+            config.update!(show_feed_binding: binding)
+          end
+        end
```
kookster, line 258: I see, so this block does the changes (if not dry run), while the above figures out what will change.
and below is how these changes get added to report. Looks good - love a good dry run option (and a report out of changes after)

app/models/apple/show_feed_binding/backfill.rb
```diff
@@ -0,0 +1,234 @@
...
+
+        increment_backfill_counts(report, binding_was_new: binding_was_new, needs_link: needs_link, changes: changes)
+        report[:actions] << {
+          config_id: config.id,
+          action: dry_run ? "would_#{action}" : action,
```
kookster, line 263: The report itself indicates this is a dry run or not, this `would_` seems unnecessary?
It's definitely just an aesthetic question, no need to change.

app/models/apple/show_feed_binding/backfill.rb
```diff
@@ -0,0 +1,234 @@
...
+          report[:errors] << {config_id: config.id, reason: "missing binding"}
+          return
+        end
+
+        show = config.build_show
+        remote_episode_ids = Apple::Show
+          .apple_episode_json(show.api, binding.apple_show_id)
+          .to_h { |episode_json| [episode_json.fetch("id").to_s, true] }
+
+        episode_ids = ::Episode.with_deleted
```
kookster, line 289: we need to check deleted as well? I guess that's fine, but if they're deleted, does it matter if the show id differs?

app/models/feeds/apple_subscription.rb
```diff
@@ -53,6 +53,25 @@ def set_tokens
...
+
+  def update_apple_show_feed_binding
+    config = apple_config
+    return unless config
+
+    public_feed = config.public_feed
+    return unless public_feed
+
+    if apple_show_id.blank?
+      public_feed.apple_show_feed_binding&.destroy!
```
kookster, line 68: that's thoughtful, agree - if there is no show id, then we have no reason to have a row at all

app/models/apple/show_feed_binding.rb
```diff
@@ -0,0 +1,26 @@
...
+module Apple
+  class ShowFeedBinding < ApplicationRecord
+    belongs_to :feed
+    belongs_to :apple_key, class_name: "Apple::Key"
+
+    has_many :delegated_delivery_configs,
+      class_name: "Apple::Config",
+      foreign_key: :show_feed_binding_id,
+      inverse_of: :show_feed_binding,
+      dependent: :nullify
```
kookster, line 11: So a specific show feed binding, it can have multiple configs?
But a podcast can only have one config? re: `podcast_has_one_apple_config`
I think I'm confused here? I would have guessed this would be a `has_one`?

kookster (commented):
I think I'm confused as to why the ShowFeedBinding has_many :delegated_delivery_configs, instead of one?
Everything else makes sense to me.

app/models/apple/show_feed_binding/backfill.rb
svevang: Just trying to nail down inconsistencies. The idea is not that we're fixing up deleted episode data, but that I'd like to know if there are problems generally.

app/models/apple/show_feed_binding.rb
svevang: Yes, this is not quite right. Patching now.

app/models/apple/show_feed_binding.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+module Apple
+  class ShowFeedBinding < ApplicationRecord
+    belongs_to :feed
+
+    has_one :config,
```
kookster, line 7: great!

kookster (approved):
has_many -> has_one, looks good

PR #1494, opened 2026-07-09
Show scoped apple episode

### LINKED ISSUE #1489: Show-scoped Apple episode identity
Apple episode identity today is implicitly single-show; this makes it show-scoped so a podcast can exist on more than one Apple show without colliding sync rows. Groundwork for the HLS publisher's linked-episode reads.
 
- Migrate `sync_logs.apple_show_id` and backfill resolvable Apple episode rows
- Replace the Apple episode uniqueness with the show-scoped key, read through `Apple::Episode`
- *Touches live delegated-delivery data — isolated on purpose*

```
sync_logs
  + apple_show_id  string  nullable

  unique (integration, feeder_type, feeder_id)
    → unique (integration, feeder_type, feeder_id, apple_show_id)
```

svevang:
closes https://github.com/PRX/feeder.prx.org/issues/1489

This branch makes all the per-episode Apple state show-scoped on top of it, and in the process untangles the Apple/Megaphone integration layer.

1. Apple Show scope for `sync_logs`, `apple_podcast_containers`, and `integrations_episode_delivery_statuses`
2. STI by integration to help with per integration/type validations: `Apple::SyncLog`, `Apple::EpisodeDeliveryStatus`, and `Megaphone::EpisodeDeliveryStatus`
3. Untangling the integration episodes so both `Integrations::Base::Episode` and `Apple::Episode` drop their method_missing-> feeder_episode delegation. This allows those integration episode models to be the only interface for episode state (e.g. `ep.mark_as_not_delivered!` instead of `ep.feeder_episode.apple_mark_as_not_delivered!`)
4. Data cleanup + backfill: `RemovePollutedAppleEpisodeDeliveryStatuses` (merged separately) and `Apple::EpisodeDeliveryIdentityBackfill`


app/models/apple/episode.rb
```diff
@@ -515,55 +512,86 @@ def video_content_type?
...
+      Apple::EpisodeDeliveryStatus
+        .where(episode_id: feeder_id, apple_show_id: [show_id, nil])
+        .order(created_at: :desc)
+    end
+
+    # The aggregate #needs_delivery? predicate answers whether any delivery work
+    # remains. This narrower predicate gates the Publisher#process_delivery!
+    # phase after any necessary upload has completed.
+    def needs_delivery_processing?
+      delivery_status(true).delivered == false
```
kookster: maybe `!delivery_status(true).delivered`
or maybe we're specifically not allowing nils?

app/models/concerns/apple_integration.rb
```diff
@@ -4,79 +4,24 @@ module AppleIntegration
...
-    has_many :apple_podcast_delivery_files, through: :apple_podcast_deliveries, source: :podcast_delivery_files,
-      class_name: "Apple::PodcastDeliveryFile"
+    has_many :apple_podcast_containers, class_name: "Apple::PodcastContainer"
 
-    alias_method :podcast_container, :apple_podcast_container
-    alias_method :apple_status, :apple_episode_delivery_status
+    private :apple_podcast_containers,
+      :apple_podcast_containers=,
+      :apple_podcast_container_ids,
+      :apple_podcast_container_ids=
```
kookster: There is so little left in this concern - can we refactor it out of existence?

app/models/apple/sync_log.rb
```diff
@@ -0,0 +1,19 @@
...
+        message: "already has an Apple episode sync log"
+      },
+      if: :episodes?
+    validates :apple_show_id, presence: true, if: :episodes?
+
+    def self.log!(attrs)
+      super(attrs.merge(integration: :apple))
+    end
+
+    def self.sti_name
```
kookster, line 12: minor, minor point, but I'd probably put sti_name as the first method, as it is in other models.

app/models/apple/sync_log.rb
```diff
@@ -0,0 +1,19 @@
+module Apple
+  class SyncLog < ::SyncLog
+    validates :feeder_id,
+      uniqueness: {
+        scope: [:integration, :feeder_type, :apple_show_id],
+        message: "already has an Apple episode sync log"
+      },
+      if: :episodes?
+    validates :apple_show_id, presence: true, if: :episodes?
```
kookster: `episodes?` here is unclear to me? is that a scope added by the enum type?

app/models/apple/episode.rb
```diff
@@ -515,55 +512,86 @@ def video_content_type?
...
 
     alias_method :container, :podcast_container
     alias_method :deliveries, :podcast_deliveries
     alias_method :delivery_files, :podcast_delivery_files
-    alias_method :delivery_status, :apple_episode_delivery_status
-    alias_method :delivery_statuses, :apple_episode_delivery_statuses
-    alias_method :apple_status, :apple_episode_delivery_status
-
-    # Delegate methods to feeder_episode
-    def method_missing(method_name, *arguments, &block)
```
kookster, line 557: +1 I am a fan of this refactor getting rid of these, and making calls more explicit and traceable.

app/models/integrations/episode_delivery_status.rb
```diff
@@ -1,9 +1,21 @@
...
 
+    INTEGRATION_CLASSES = {
+      "apple" => "Apple::EpisodeDeliveryStatus",
+      "megaphone" => "Megaphone::EpisodeDeliveryStatus"
+    }.freeze
+
+    def self.find_sti_class(type_name)
+      integration = base_class.type_for_attribute(inheritance_column).cast(type_name)
+      INTEGRATION_CLASSES.fetch(integration).constantize
+    end
```
kookster, line 17: that's cleaner, nice refactor

app/models/integrations/episode_integrations.rb
```diff
@@ -30,6 +14,10 @@ def integration_episode(integration)
     end
   end
 
+  def megaphone_episode
+    Megaphone::Episode.new.tap { |episode| episode.feeder_episode = self }
+  end
```
kookster: Should this megaphone specific method be in the generic EpisodeIntegrations module?

app/models/megaphone/episode_delivery_status.rb
```diff
@@ -0,0 +1,76 @@
...
+      new_status.save!
+      episode.episode_delivery_statuses.reset
+      new_status
+    end
+
+    def self.delete_status(episode)
+      episode.episode_delivery_statuses.megaphone.delete_all
+    end
+
+    def self.unfinished(episodes)
```
kookster: This seems like I would expect it on the episode, not in the status class?

app/models/megaphone/episode_delivery_status.rb
```diff
@@ -0,0 +1,76 @@
...
+      SQL
+
+      episodes
+        .joins(latest_status)
+        .where('(eds."episode_id" is null) or ((eds."delivered" = false or eds."uploaded" = false) and eds."integration" = ?)', integration)
+    end
+
+    def mark_as_uploaded!
+      update_status(uploaded: true)
+    end
```
kookster, line 33: I'm confused why these methods are both in episode and episode delivery status?
It seems like quite a bit of duplication in those megaphone classes now?

app/models/megaphone/episode.rb
```diff
@@ -49,9 +49,13 @@ def self.list(megaphone_podcast, published_only = true)
       episode.list(published_only)
     end
 
+    def self.unfinished(feeder_episodes)
+      Megaphone::EpisodeDeliveryStatus.unfinished(feeder_episodes)
+    end
```
kookster, line 68: so unfinished ends up in both, I would have had it in episode, and then referenced in status - it's a method to return episodes, that's a finder on a model?

db/migrate/20260708000003_add_apple_show_id_to_sync_logs.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddAppleShowIdToSyncLogs < ActiveRecord::Migration[7.2]
+  def change
+    add_column :sync_logs, :apple_show_id, :string
```
kookster: So, the `external_id` is used by both the megaphone and the apple integrations to store their respective episode id - could we have an external_show_id that could be used by both apple and megaphone?
I think the megaphone integration saves the megaphone podcast id to the sync_log external_id for the `feeder_type: :feeds`, so a more generic way to save that external podcast id might make sense in here?

app/models/apple/episode.rb
```diff
@@ -174,20 +176,17 @@ def self.upsert_sync_logs(episodes, results)
...
 
-      sl = SyncLog.log!(
-        integration: :apple,
+      sl = Apple::SyncLog.log!(
         feeder_id: ep.feeder_episode.id,
         feeder_type: :episodes,
         external_id: apple_id,
-        api_response: res
+        api_response: res,
+        apple_show_id: apple_show_id
```
kookster: I wonder if this needs to be an Apple specific show id?
We have a generic external_id, maybe a external_parent_id, or external_show_id that could be used by both Apple and Megaphone (and any other integration)?

Maybe we wouldn't also need an Apple specific Sync log subclass?

db/migrate/20260708000004_add_apple_show_id_to_apple_delivery_state.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class AddAppleShowIdToAppleDeliveryState < ActiveRecord::Migration[7.2]
```
kookster, line 3: yes, def should be apple specific names for an apple table - and makes sense to add show tracking to the containers +1

app/models/apple/episode.rb
kookster: I'll also say, having the same value consistently in `apple_show_id` is also valuable, so what you've got may actually be clearer in the end...

Maybe another middle road would be the db col with a generic name, and a accessor/mutator pair that is specific to the integration?

Or maybe this isn't that important and I'm overthinking it.

kookster (approved):
Lots of good refactoring in here.

The only thing I might actually change is to see if we could factor out the rest of `app/models/concerns/apple_integration.rb` and get rid of that concern? Almost all of its methods are moved out anyway.

And I had some questions about an apple specific sync log subclass and apple specific column in there - I think that is a matter of taste/opinion, so while I appreciate you considering my comments on that, please don't take those as a demand for change.

app/models/apple/episode.rb
svevang: external_show_id could work. Would be nice to keep this as concrete as possible. Will look at renaming

Yes, maybe we could get rid of the apple subclass, once the megaphone side lands. But having an apple specific contract (on top of the existing sync log contract) is not the worst thing?

app/models/imports/episode_megaphone_import.rb
```diff
@@ -215,7 +215,7 @@ def update_delivery_status
       delivered: true
     }
 
-    delivery_status = Megaphone::EpisodeDeliveryStatus.update_status(episode, delivery_status_attrs)
+    delivery_status = megaphone_episode.update_delivery_status(delivery_status_attrs)
```
kookster, line 218: +1 yeah, I think it already had that method

kookster (approved):
Looks like you did even more than I expected from my weekly suggested comments - looks good!

PR #1495, opened 2026-07-13
Include Feeds in auth'd Podcasts API

cavis:
To do https://github.com/PRX/augury.prx.org/issues/1901, we need some way to scrape feed data from the augury API.

Rather than standing up a new `/api/v1/authorization/feeds?since=` endpoint, seemed easier to just return with the podcast.  Since feeds always `touch: true` the podcast.

### DIFF
--- a/app/representers/api/auth/podcast_representer.rb
+++ b/app/representers/api/auth/podcast_representer.rb
@@ -1,6 +1,15 @@
 class Api::Auth::PodcastRepresenter < Api::PodcastRepresenter
   property :deleted_at, writeable: false
 
+  # TODO: this should probably be embedded, but zooms aren't working!
+  # update when we move off hal_api-rails (probably to halbuilder)
+  collection :feeds, writeable: false do
+    property :id
+    property :label
+    property :slug
+    property :private
+  end
+
   def self_url(podcast)
     api_authorization_podcast_path(podcast)
   end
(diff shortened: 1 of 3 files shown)


app/representers/api/auth/podcast_representer.rb
```diff
@@ -1,6 +1,15 @@
 class Api::Auth::PodcastRepresenter < Api::PodcastRepresenter
   property :deleted_at, writeable: false
 
+  # TODO: this should probably be embedded, but zooms aren't working!
+  # update when we move off hal_api-rails (probably to halbuilder)
+  collection :feeds, writeable: false do
+    property :id
+    property :label
+    property :slug
+    property :private
```
kookster, line 10: only in the auth'd response, so this is fine

app/representers/api/auth/podcast_representer.rb
```diff
@@ -1,6 +1,15 @@
 class Api::Auth::PodcastRepresenter < Api::PodcastRepresenter
   property :deleted_at, writeable: false
 
+  # TODO: this should probably be embedded, but zooms aren't working!
+  # update when we move off hal_api-rails (probably to halbuilder)
+  collection :feeds, writeable: false do
```
cavis, line 6: I tried many things to make this an `_embedded` (which CMS still has, but we've never used in Feeder).  So then you could `?zoom=prx:feeds` to get them returned.

For some reason, the `?zoom` params aren't getting passed down into the hal_api-rails methods that actually decide if we load that embed or not.  Spent probably too long trying to chase that down!

So, fast solution, just make it an attribute for now.  If/when we rip out hal_api-rails, can reconsider.

kookster (approved):
looks good, I might think about adding a test, but this is a pretty basic addition.

PR #1497, opened 2026-07-14
Apple config routing

### LINKED ISSUE #1496: Add environment-selected routing to `Apple::DelegatedDeliveryConfig`.
This prepares the HLS publisher to use `Apple::ShowFeedBinding` and lets intermediate releases roll forward or back safely.

  - Rename `Apple::Config` to `Apple::DelegatedDeliveryConfig` while preserving the `Apple::Config` compatibility alias
  - Add `APPLE_ROUTING_SOURCE` to select the Apple credential, show ID, and public feed Apple crawls:
    - `legacy` (default) preserves the current lookup
    - `show_feed_binding` follows `apple_configs.show_feed_binding_id` to `apple_show_feed_bindings` and reads `apple_key_id`, `feed_id`, and `apple_show_id`
  - Keep `apple_configs.feed_id` as the delivery feed under both routing sources

  **Done when**

  - [x] Omitting `APPLE_ROUTING_SOURCE` selects `legacy` by default
  - [x] An unsupported routing value fails explicitly
  - [x] Both readers pass the same delegated-delivery publishing tests

  **Deploy Checklist**
  - [ ] Both readers resolve the same Apple credential, show ID, public feed, and delivery feed for every active delegated-delivery config (on deploy)
  - [ ] `Apple::ShowFeedBinding::Backfill.verify_routing_equivalence!` reports zero mismatches (on deploy)
  - [ ] Existing delegated-delivery publishing succeeds in production with `APPLE_ROUTING_SOURCE=legacy`

svevang:
closes #1496


app/models/apple/show_feed_binding/backfill.rb
```diff
@@ -10,7 +10,7 @@ class Backfill
       def self.backfill!(dry_run: false)
         report = new_backfill_report(dry_run: dry_run)
 
-        Apple::Config.find_each do |config|
+        Apple::DelegatedDeliveryConfig.find_each do |config|
```
kookster: +1 you renamed the relationship in 1493 I think, this makes sense as a continuation of differentiating the configs by purpose

app/models/apple/delegated_delivery_config.rb
```diff
@@ -0,0 +1,103 @@
...
+      feed
+    end
+
+    def not_default_feed
+      if feed&.default?
+        errors.add(:feed, "cannot use default feed")
+      end
+    end
+
+    def podcast_has_one_apple_config
```
kookster, line 85: Wait, I thought we were making this so different feeds in the same podcast could publish to different shows on apple.

An example is Milk Street Kitchen, with a public feed / private feed config publishing an ad free version of the audio, which is one apple config with `sync_blocks_rss` true

Then in the same podcast, a best of desserts private feed, published as a different Apple show, with no associated public feed, just a subscriber feed only on Apple with `sync_blocks_rss` false

Wouldn't that need to be 2 configs for the podcast, not limited to 1?
(Or is this scenario not yet covered by this PR?)

kookster: re-reading #1496, I think this scenario is not in scope yet for this PR - so my comment is maybe more about where this will need to change when we do want to handle such a scenario?

app/models/apple/show.rb
```diff
@@ -119,7 +126,7 @@ def show_data(attributes, id: nil)
     end
 
     def apple_id
-      sync_log&.external_id
+      @configured_apple_show_id.presence || sync_log&.external_id
```
kookster, line 129: would it be bad to make that `||=`? maybe too much of a side effect to memoize this

kookster (approved):
Looks like this is faithfully handling how we do delegated delivery now - enforcing one DelegatedDeliveryConfig per podcast - my only real question is if this PR should get into that being potentially 1 DelegatedDeliveryConfig per feed instead?

svevang:
@kookster The datamodel-cutover in #1511 relaxes the per-podcast validation to a per-feed validation (But that could land in the UI!). At that point we're ready for the HLS publishing side of this and the new UI.

I've rebased this branch a bunch, and pulled the renaming into a parent branch in the stack. Other changes since you reviewed are cosmetic refactors.

PR #1498, opened 2026-07-15
Standardize enclosure urls

cavis:
Needed for #1491.

This PR removes the `feeds.enclosure_template` field from the DB.  We aren't currently using any meaningful customizations in prod.  And our Dovetail URL formats have gotten stricter over the years, so modifying them is generally a bad idea™.

Will require nulling out the few remaining altered templates, plus removing the no-op `/_/` underscore from a handful.


app/models/concerns/episode_enclosure.rb
```diff
@@ -0,0 +1,60 @@
...
+      orig_base = File.basename(orig_fn, orig_ext)
+      ext = feed&.file_ext || orig_ext[1..] || "mp3"
+      "#{orig_base}.#{ext}"
+    end
+  end
+
+  # TODO: not accurate for Feeds with audio_formats, or really any
+  # Contents that need transcoding to match the first original
+  def enclosure_file_size(feed: podcast&.default_feed)
+    media_file_size
```
cavis, line 58: Realizing the byte lengths we're putting in the RSS are _mostly_ inaccurate these days.  In addition to ad-injection, most newer Feeds have an explicit audio format.  So we're already transcoding originals.

Kind of makes me wonder if we should just remove this field from the RSS, since it's so misleading.  A problem for another day.

app/models/episode.rb
```diff
@@ -291,23 +284,8 @@ def podcast_image
...
-
-    if res.is_a?(Net::HTTPRedirection)
-      head_request(res[:location], redirects + 1)
-    elsif res.is_a?(Net::HTTPSuccess)
-      res
-    end
-  rescue URI::InvalidURIError, Socket::ResolutionError, Net::ReadTimeout => e
-    Rails.logger.info(e.message)
+  def head_request
+    http_head(enclosure_url(prefix: false), timeout: 0.1) unless override?
```
cavis, line 288: I already needed to update this method, and realized it had already been abstracted into a generic `HttpUtil`.

app/models/apple/episode.rb
```diff
@@ -230,13 +230,11 @@ def podcast
     end
 
     def enclosure_url
-      url = EnclosureUrlBuilder.new.base_enclosure_url(podcast, feeder_episode, private_feed)
-      EnclosureUrlBuilder.mark_authorized(url, private_feed)
+      feeder_episode.enclosure_dovetail_url(feed: private_feed, prefix: false, auth: true)
```
kookster, line 233: good method name - implies it is specifically a DTR url

app/models/concerns/episode_enclosure.rb
```diff
@@ -0,0 +1,60 @@
+require "active_support/concern"
+
+module EpisodeEnclosure
+  extend ActiveSupport::Concern
+
+  def enclosure_url(**opts)
+    override? ? enclosure_override_url : enclosure_dovetail_url(**opts)
```
kookster, line 7: nice, very readable

app/models/concerns/episode_enclosure.rb
```diff
@@ -0,0 +1,60 @@
...
+        nil
+      elsif auth == true
+        feed&.tokens&.first&.token
+      else
+        auth.presence
+      end
+
+    url_parts = []
+    url_parts << feed.enclosure_prefix if feed&.enclosure_prefix.present? && prefix
+    url_parts << (ENV["DOVETAIL_HOST"] || "dovetail")
```
kookster: should that default be an actual domain? `dovetail.prxu.org`?

app/models/concerns/episode_enclosure.rb
```diff
@@ -0,0 +1,60 @@
...
+    url_parts << (podcast_id || "podcast")
+    url_parts << feed.slug if feed&.slug.present?
```
kookster, line 24: should we check if it is not the default feed? I think we only do this on non-default feeds, not the public default one?

app/models/concerns/episode_enclosure.rb
```diff
@@ -0,0 +1,60 @@
...
+
+    query = {auth: auth, _t: prx_jwt}.compact.to_query
+    query.present? ? "#{url}?#{query}" : url
+  end
+
+  def enclosure_content_type(feed: podcast&.default_feed)
+    if override? || medium_video?
+      media_content_type
+    else
+      feed&.mime_type || media_content_type
```
kookster, line 39: I wonder how this will work when we have Apple HLS?
The enclosure will still be audio in that case, but the alt enclosure/Apple HLS will be video.
Seems like this will still work so long as the media_content_type comes back as audio for the derived mp3 version of the uploaded video file?

app/models/concerns/episode_enclosure.rb
kookster: I think putting in the ad free byte count is not terrible, and yes, it will be off.
I wonder what others are doing in their feeds for DAI audio...

app/models/concerns/episode_media.rb
```diff
@@ -184,24 +184,20 @@ def media=(files)
...
+  def media_content_type
     if override?
       external_media_resource&.mime_type || "audio/mpeg"
-    elsif (media_content_type || "").starts_with?("audio")
-      feed_content_type || media_content_type
-    elsif media_content_type
-      media_content_type
     else
-      "audio/mpeg"
+      media.first&.mime_type || "audio/mpeg"
```
kookster, line 195: so if I upload a video for Apple HLS, but then transcode and extract an mp3 from it for the enclosure, I'm not sure what type the first `media` will have...anyway, for another day.

app/models/megaphone/episode.rb
```diff
@@ -461,12 +461,7 @@ def url_filename(url)
     end
 
     def enclosure_url
-      url = EnclosureUrlBuilder.new.base_enclosure_url(
-        feeder_podcast,
-        feeder_episode,
-        private_feed
-      )
-      EnclosureUrlBuilder.mark_authorized(url, private_feed)
+      feeder_episode.enclosure_dovetail_url(feed: private_feed, prefix: false, auth: true)
```
kookster, line 464: both Apple and Megaphone extend `Integrations::Base::Episode`, could maybe move this there

kookster (approved):
lgtm! A few comments, thinking ahead to Apple HLS

app/models/concerns/episode_enclosure.rb
cavis: Yeah, will update.

app/models/concerns/episode_enclosure.rb
cavis: By definition, the default feed has `slug == nil`.  So same difference.

app/models/concerns/episode_enclosure.rb
cavis: Yeah, that was the motivation for this PR.  The `VideoContent` subclass for HLS stuff will need to override its `content` to be audio/mp3.  And then have some sort of `alt_content` method, for the alternate enclosure?  Something like that.

Issue #1426, opened 2026-02-20
Waveform editor improvements


cavis:
The waveform UI is now involved in stream-recording.  Some improvements may be in order to make it easier working with 1 hour clips.

Laundry list of ideas (not saying we need to do all of these):

- [ ] Start playback position at end of preroll slice
- [ ] Spacebar to play/pause, if any part of the waveform UI is focused.
- [ ] `I` and `O` for marking in and out points
- [ ] Other keyboard shortcuts for marking breaks?  (Vs pre/post slicing)
- [ ] Arrows to nav left-right.
- [ ] Maybe also mouse scroll left-right?  And the usual DAW zoom controls?  Overkill?
- [ ] Maybe a different UX to allow users to "select" the timeframe they want to crop to?
- [ ] Pre/Post are kind of misnomers for the recorded hours.  More like, "this is where we think the top of the hour is".
- [ ] The timestamp `00:00` numbers aren't editable.  Are they just a background or something?
- [ ] Allow zooming all the way out
- [ ] Update view while dragging the horizontal scrollbar
- [ ] An actual playhead instead of the change in color
- [ ] Some sort of bug I haven't tracked down yet ... clips copied to episodes always look like their segmentation "changed" to the media page.  Naving away warns you about unsaved changes.

PR #1502, opened 2026-07-21
Update podcast clips

### LINKED ISSUE #1492: Forgot to write podcast_clips#update
Turns out, I forgot to wire this up.  So you can't change/save the start and end trim timestamps from the clips page.

cavis:
Fixes #1492.  Adds the `podcast_clips#update` action.

image

NOTE: this reuses the episode Uncut media UI.  Which does let you trim both the start/end of the file.  However... we already know the exact duration of these, and that we were attempting to capture 1 hour.

So for now, I just ignore the end offset, if you set it.  And the clip end is always the start + file-duration.  You're just marking where the hour starts.  If/when we get to #1426, should revisit this.

### DIFF
--- a/app/controllers/podcast_clips_controller.rb
+++ b/app/controllers/podcast_clips_controller.rb
@@ -15,10 +15,19 @@ class PodcastClipsController < ApplicationController
   def show
   end
 
+  def update
+    if @clip.update(update_clip_params)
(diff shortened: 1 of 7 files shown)


brandonhundt (approved):
This makes sense. Certainly gets the start and end closer to the actual chop point. Thanks @cavis :shipit:

PR #1505, opened 2026-07-29
Upgrade to Rails 8.1.3.1

kookster:
I tried out the Rails upgrade skill for this:
https://github.com/ombulabs/claude-code_rails-upgrade-skill

It upgrades Rails 7.2.3 → 8.1.3.1, taken in sequence (7.2.3.2 patch → 8.0.5.1 → 8.1.3.1) with the suite green at each hop. Since this app gets exercised in staging only after merge, the verification checklist below matters more than usual — most of it covers code the test suite does not reach.

## Gem bumps the upgrade forced

- `clickhouse-activerecord` 1.1.2 → **1.6.7** (capped below 8.0). Five minor versions in one jump — the biggest single risk in this PR, and it sits under the analytics endpoints.
- `paranoia` 3.0.0 → 3.1.0 (capped below 8.1).

Everything else resolved unchanged.

## `benchmark` gem dropped

Active Support carried a transitive dependency on the `benchmark` gem through Rails 8.0. Rails 8.1 drops it, so `Benchmark.measure` in `FeederLogger#elapsed` / `#debug_elapsed` would have raised `NameError` in production. Added back with tests.

## Framework defaults

`load_defaults` stays at **7.1**. All 14 new defaults from 7.2/8.0/8.1 are staged in `config/initializers/new_framework_defaults_*.rb`; 8 are enabled, 6 stay commented with what they'd change here and what to check. Deferring keeps this PR to the version bump.

`strict_freshness` is set in `config/application.rb` rather than its initializer — the Action Dispatch railtie reads it before app initializers run, so assigning it in the initializer silently did nothing. 🤷 

## Decisions worth pushing back on

**`config.active_storage.variant_processor = :disabled`** — Rails 8.1 added a boot-time warning when the variant processor is configured without `image_processing`.  We don't use it; if we ever do, set `image_processing` instead.

**`action_on_path_relative_redirect = :raise`** — this *raises* where 8.0 only logged. We use url helpers, so this is unlikely to be an issue, but could cause > 500s.

**`raise_on_missing_required_finder_order_columns`** — the `Rollups::` ClickHouse models look like they'd break this: bare `ActiveRecord::Base`, no `id`, no `PRIMARY KEY`. They don't. ClickHouse reports a MergeTree `ORDER BY` as `PRI` in `SHOW COLUMNS`, and the adapter turns that into a composite primary key. Verified against the real castlehouse schema with the config on; all 54 models with a table resolve order columns. Taken now because the opt-out is removed in 8.2.

**`database.yml` `pool:` → `max_connections:`** — 8.1 keeps `pool:` as a deprecated alias and still honors the value; the alias goes away in 8.2. `DATABASE_POOL_SIZE` still drives it.

## Verification checklist for staging

Weighted toward paths with no automated coverage. Run single-process (parallel workers are not merged, which understates the figure), the suite covers **39.6%** — but very unevenly:

- Jobs / workers — 94%
- Representers — 90%
- Controllers — 61%
- Models — 42%
- Lib — 5%
- **Helpers — 0%** (1117 LOC)

Helpers and the UI get hit by integration tests, but still, good to know.


lib/feeder_logger.rb
```diff
@@ -1,5 +1,9 @@
 # frozen_string_literal: true
 
+# Active Support dropped its transitive dependency on benchmark in Rails 8.1,
+# so Benchmark is no longer loaded for us.
```
svevang: Comment seems to be chronologically rooted in the development, not pointing at a future concern. So maybe not worth it?

test/lib/feeder_logger_test.rb
```diff
@@ -0,0 +1,70 @@
+require "test_helper"
+
+describe FeederLogger do
+  # Guards a real regression: Active Support carried a transitive dependency on
+  # the benchmark gem through Rails 8.0, and Benchmark.measure here relied on
+  # it. Rails 8.1 dropped it, so these methods raised NameError until
+  # feeder_logger.rb started requiring benchmark itself.
```
svevang: Love the tests, I'm not sure if this comment is worth it.

config/puma.rb
```diff
@@ -1,43 +1,44 @@
...
 #
-worker_timeout 3600 if ENV.fetch("RAILS_ENV", "development") == "development"
-
-# Specifies the `port` that Puma will listen on to receive requests; default is 3000.
+# You can control the number of workers using ENV["WEB_CONCURRENCY"]. You
+# should only set this value when you want to run 2 or more workers. The
+# default is already 1. You can set it to `auto` to automatically start a worker
+# for each available processor.
 #
-port ENV.fetch("PORT", 3002)
```
svevang, line 18: Removing this means the fallback port is 3000. Port 3002 is also mentioned in the README and may be tied to some id.prx.org state.

svevang (approved):
Looks good!

config/puma.rb
kookster: yeah, it defaults below to 3000, I'll put it back to 3002, ty!

lib/feeder_logger.rb
kookster: good call.

test/lib/feeder_logger_test.rb
kookster: sure, happy to delete

PR #1506, opened 2026-08-03
Add Arizona timezone

### LINKED ISSUE #1504: Live Stream Recording Arizona Timezone Issue
From Paloma:

We received this feedback from Kaja Brown at KRDP (in Arizona) about the live stream recording. 

> Arizona technically has its own time zone since we don’t observe Daylight Saving Time.
> Is it possible to add a time zone option for Arizona that is UTC-7 all year long? That way the live feed will always record the correct hour of our schedule for the podcast.

cavis:
Fixes #1504.


brandonhundt (approved):
LGTM. Thanks @cavis

PR #1508, opened 2026-08-04
fix(css): mute bg color of prx-blank-lines in dark mode

rpeterman-gp:
Change the `.prx-blank-lines` background color to a muted neutral color. Embed preview should now standout more and be the primary focused element in that "in content" preview layout.

image


brandonhundt (approved):
Thanks @rpeterman-gp

